PR #1, opened 2018-06-04
MAID-2757: add ID traits

Fraser999:


src/parsec.rs
```diff
@@ -32,6 +32,8 @@ pub struct Parsec<T: Serialize + DeserializeOwned, S: SecretId> {
     extra_votes: BTreeMap<S::PublicId, Vec<Vote<T, S::PublicId>>>,
 }
 
+// TODO - remove
+#[cfg_attr(feature = "cargo-clippy", allow(needless_pass_by_value))]
```
maqi: I put it as pass by reference in my commit.
which I think is better than having a `TODO` and `omit of clippy complain?

src/id.rs
```diff
@@ -48,6 +48,23 @@ pub trait SecretId {
 #[serde(bound = "")]
 #[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
 pub struct Proof<P: PublicId> {
```
maqi: do we need a constructor for Proof?

src/parsec.rs
Fraser999: I was favouring leaving the function's signature as is, since that's how we want the signature to be - we want to pass these by value really I think.

src/id.rs
Fraser999: Apart from deserialising, we have `PublicId::create_proof()` - I think that should be enough?

maqi: create_proof() is of SecretId ?
do you think it will be better to move it to PublicId instead?

Fraser999: No - sorry - I meant `SecretId::create_proof()` :)

maqi: when we try to compose a block from payload and votes, we only have pub_id, and have to construct a proof.
In my commit, I have to change the members of Proof to be `pub`. However I think a constructor will make more sense?

Fraser999: Yeah - for that we should probably implement something like [this](https://github.com/maidsafe/routing/blob/dev/src/data_chain/vote.rs#L44-L52)

Fraser999: Done in https://github.com/maidsafe/parsec/pull/2/commits/b89eab7a7e161e7d50481b85362c0ef2fe0a4f89.

PR #3, opened 2018-06-04
MAID-2759 implement Block

maqi:


src/block.rs
```diff
@@ -6,14 +6,57 @@
...
+        &self.payload
+    }
+
+    /// Returns the proofs of this block.
+    pub fn proofs(&self) -> &BTreeSet<Proof<P>> {
+        &self.proofs
+    }
+
+    /// Convert the vote to a proof and insert it into proofs of the block.
+    pub fn add_vote(&mut self, peer_id: &P, vote: &Vote<T, P>) -> Result<bool, Error> {
```
Fraser999: Should this just return `Result<(), Error>` since we can't get `Ok(false)`, or else we could remove the `ProofAlreadyExisted` variant and just return `Ok(self.proofs.insert(proof))`?

maqi: I prefer return with `Ok(self.proofs.insert(proof))` and remove `ProofAlreadyExisted` variant.
however the task description states `log error` for that case.
which one you prefer?

Fraser999: Returning a `bool` is fine by me.

src/block.rs
```diff
@@ -6,14 +6,56 @@
...
+    pub fn payload(&self) -> &T {
+        &self.payload
+    }
+
+    /// Returns the proofs of this block.
+    pub fn proofs(&self) -> &BTreeSet<Proof<P>> {
+        &self.proofs
+    }
+
+    /// Convert the vote to a proof and insert it into proofs of the block.
```
ustulation: There's a discussion and it seems sorted but the comments explain nothing about the return type. Can we pls have comments explaining the function purpose and the return type a bit better (and it's a part of the API which is why we have the doc comment anyway i suppose).

can this be done for all the functions too pls, makes the review easier. Ok to do in the subsequent PRs but pls note it.
(cc @maqi @Fraser999 )

PR #4, opened 2018-06-05
Maid-2762 implement PeerManager

maqi:


src/peer_manager.rs
```diff
@@ -6,17 +6,98 @@
...
+
+    /// Returns peer info: created events index and hash.
+    pub fn peer_info(&self, peer_id: &S::PublicId) -> Option<&BTreeMap<u64, Hash>> {
+        self.peers.get(peer_id)
+    }
+
+    /// Returns all sorted peer_ids.
+    pub fn all_ids(&self) -> Vec<&S::PublicId> {
+        let mut ids: Vec<_> = self.peers.keys().collect();
+        ids.sort();
```
Fraser999: These are sorted already since `peers` is a `BTreeMap`.

src/peer_manager.rs
```diff
@@ -6,17 +6,98 @@
...
+    pub fn add_peer(&mut self, peer_id: S::PublicId) {
+        let _ = self.peers.entry(peer_id).or_insert_with(BTreeMap::new);
+    }
+
+    /// Check whether the input count becomes the super majority of the network.
+    pub fn is_super_majority(&self, count: usize) -> bool {
+        3 * count > 2 * self.peers.len()
+    }
+
+    /// Returns the index of the last event created by this peer. Returns 0 if cannot find.
```
Fraser999: This would probably be better to return `None` if the peer isn't found or if we don't have any events from the peer yet (similar to `get_last_created_event_hash()`) since `0` is a valid index for an event.

src/peer_manager.rs
```diff
@@ -6,17 +6,98 @@
...
+    pub fn get_last_created_event_index(&self, peer_id: &S::PublicId) -> u64 {
```
Fraser999: I think we should remove all `get_`s from the getters and I also feel that `created` could be removed from the 4 functions too without sacrificing clarity.

src/peer_manager.rs
```diff
@@ -6,17 +6,98 @@
...
+    /// Add event created by the peer.
+    pub fn add_created_event<T: Serialize + DeserializeOwned + Debug + Eq, P: PublicId>(
+        &mut self,
+        peer_id: S::PublicId,
+        event: &Event<T, P>,
+    ) {
+        let entry = self
+            .peers
+            .entry(peer_id)
+            .or_insert_with(BTreeMap::new)
```
Fraser999: I wouldn't think this function should silently add a new peer?  Seems like we should return an error in that case?

src/peer_manager.rs
```diff
@@ -6,17 +6,97 @@
...
+
+    /// Returns the hash of the indexed event.
+    pub fn event_by_index(&self, peer_id: &S::PublicId, index: u64) -> Option<&Hash> {
+        self.peers
+            .get(peer_id)
+            .and_then(|events| events.get(&index))
+    }
+
+    /// Add event created by the peer. Returns an error if `peer` is not known, `Ok(None)` if the
+    /// `index` wasn't previously held for that peer, or `Ok(Some(hash))` if it was previously held
```
ustulation: Why not return an error if we get something from the peer with exactly the same index but a different event for that index (i.e. hash doesn't match) ? That seems more like a malice/error case than normal and we shouldn't be doing work for that (even if it's replacement of previous value).

ustulation: cool, merging this then

PR #5, opened 2018-06-06
chore/readme: add crate and doc badges

Fraser999:
Many thanks to https://github.com/kvark and https://github.com/csherratt for passing ownership of
"parsec" on crates.io over to MaidSafe!


ustulation:
it has passed all but nightly which is still running - but merging anyway because it affects only readme and we need a PR after this to test Cargo Publishing anyway to crates.

PR #6, opened 2018-06-06
MAID-2761: add all structs related to gossiping messages

Fraser999:


src/block.rs
```diff
@@ -50,7 +49,10 @@ impl<T: Serialize + DeserializeOwned + Debug + Eq, P: PublicId> Block<T, P> {
         &self.proofs
     }
 
-    /// Convert the vote to a proof and insert it into proofs of the block.
+    /// Converts `vote` to a `Proof` and attempts to add it to the block.  Returns an error if
+    /// `vote` is invalid (i.e. signature check fails or the `vote` is for a different network
+    /// event, `Ok(true)` if the `Proof` wasn't previously held in this `Block`, or `Ok(false)` if
```
maqi: is there a missing `)` after the word of `event` ?

src/error.rs
```diff
@@ -20,12 +20,6 @@ quick_error! {
             description("Signature cannot be verified")
             display("The message or signature might be corrupted, or the signer is wrong.")
         }
-        /// IO error.
-        Io(error: ::std::io::Error) {
-            description(error.description())
-            display("I/O error: {}", error)
-            from()
-        }
```
maqi: the crate may need to dump the stuff to a file eventually. so maybe this IO error is still required?

src/gossip/event.rs
```diff
@@ -6,23 +6,133 @@
...
+            Hash::from(serialised_content.as_slice())
+        } else {
+            return Err(Error::SignatureFailure);
+        };
+
+        Ok(Self {
+            content: packed_event.content,
+            signature: packed_event.signature,
+            hash,
+            index: Some(0),
```
maqi: I assume this index needs to be updated outside?
shall we make a comment here to explain it?
and same for the following `new` function.

src/error.rs
Fraser999: I'd prefer to leave it out until we need it.

src/gossip/event.rs
Fraser999: Fair point.  I'll actually change the `index` to be initialised to `None` to make this clearer too.

src/peer_manager.rs
```diff
@@ -86,17 +83,35 @@ impl<S: SecretId> PeerManager<S> {
...
     }
 
-    /// Add event created by the peer. Returns an error if `peer` is not known, `Ok(None)` if the
-    /// `index` wasn't previously held for that peer, or `Ok(Some(hash))` if it was previously held
-    /// and the previous hash is returned.
-    pub fn add_event<T: Serialize + DeserializeOwned + Debug + Eq, P: PublicId>(
+    /// Adds event created by the peer. Returns an error if `peer_id` is not known, or if we already
+    /// held an event from this peer with this index, but that event's hash is different to the one
+    /// being added (in which case `peers` is left unmodified).
+    pub fn add_event<T: NetworkEvent, P: PublicId>(
```
maqi: We don't have `remove_event` function.
Which means the first coming event will occupy the position forever and cannot be corrected.
Will there be chance such correction will be required to carry out?

Fraser999: I think the plan is that we should only be adding events here once we know they're valid.  For example, to add event 4, we'd need to have event 3 already so we can validate 4's `self_parent` is 3.  Or if we're about add an `Initial` event for a peer, we should check we don't already have one.  So really the `InvalidEventIndex` error shouldn't ever get returned.

PR #7, opened 2018-06-07
MAID-2819: add some private helper functions

Fraser999:


src/parsec.rs
```diff
@@ -75,4 +75,101 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+        }
+
+        let creator_id = event.creator().clone();
+        let _ = event.last_ancestors.insert(creator_id, event_index);
+        Ok(())
+    }
+
+    // This should be called only once `event` has had its `index` and `last_ancestors` set
+    // correctly (i.e. after calling `Parsec::set_last_ancestors()` on it)
+    fn set_first_descendants(&mut self, event: &mut Event<T, S::PublicId>) -> Result<(), Error> {
```
maqi: just wondering at the last line of `set_last_ancestors`, instead of returning `OK(())`, can we returns `self.set_first_descendants(event)` ?

src/error.rs
```diff
@@ -36,5 +36,10 @@ quick_error! {
             description("Invalid event index")
             display("The given event has an invalid index.")
         }
+        /// The given event is invalid or malformed.
+        InvalidEvent {
```
maqi: we have `InvalidEventIndex` , which is also kind of `invalid` or `malformed`.
shall we make these two errors more mutual to each other?

src/parsec.rs
```diff
@@ -75,4 +75,101 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+                            opt_hash = other_event.self_parent();
+                            continue;
+                        }
+                    }
+                }
+                break;
+            }
+        }
+
+        // ...then actually modify the events.
```
ustulation: just a speculation: the borrow seems to be that of two separate/unrelated variables `self.peer_manager` and `self.events`. Can not `self.events` be borrowed mutably if `self.peer_manager` is borrowed (mutably or otherwise) ? Or was the conflict due to something else ?

src/parsec.rs
Fraser999: I'm not so sure about this one.  I'd think that wherever we call this from, we'd need to set the index correctly (probably via another private method), then call `set_last_ancestors()` then `set_first_descendants()` which makes that code fairly self-documenting.

If we do as you suggest, we should rename `set_last_ancestors()` since it'd be doing more than just that.

src/error.rs
Fraser999: Yeah - maybe.  Are you thinking it'd be better to just use `InvalidEvent` for everything?

src/parsec.rs
Fraser999: The problem is with multiple mutable borrows of `self.events` if we change line 153 to `self.events.get_mut(hash)`.

src/parsec.rs
maqi: as we will always call `set_last_ancestors` then `set_first_descendants`, then I think it will be better to have `set_first_descendants` be called inside of `set_last_ancestors` and rename it, to avoid accident misuse.

src/error.rs
maqi: I am fine with using `InvalidEvent` for all. Not sure about others.
or, we have `InvalidEventIndex`, `InvalidEventParent`, `InvalidEventLastAncestor` to be explicit.

btw. in your current change, there are two places shall use `InvalidEventIndex` instead of `InvalidEvent` ? as the error is about `None` event.index ?

src/parsec.rs
Fraser999: Well, we'd really have to just remove `set_first_descendants()` to achieve that I think - if it remained as a private method it'd still be available to be called from outside `set_last_ancestors()`.  So I still prefer the two smaller, more focused methods than one larger one (which itself is prone to the same misuse if `event.index` isn't set correctly).

maqi: If there will only be one place (during add_event) calling these two functions, then I am fine with keep these two smaller functions.
If they could be called in multiple places, I'd prefer to couple them into one.

PR #8, opened 2018-06-07
MAID-2764 implement see/strongly_see; implement collect meta votes

maqi:
implement see/strongly_see;
implement collect meta votes when crawling along the gossip_graph

### DIFF
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -10,11 +10,59 @@ use block::Block;
 use error::Error;
 use gossip::{Event, Request, Response};
 use hash::Hash;
-use id::SecretId;
+use id::{PublicId, SecretId};
 use network_event::NetworkEvent;
 use peer_manager::PeerManager;
 use std::collections::{BTreeMap, BTreeSet, VecDeque};
 use vote::Vote;
+
+// The struct holds an event's meta votes towards a candidate voter, according to the PARSEC.
+#[derive(Default)]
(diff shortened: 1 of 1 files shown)


src/parsec.rs
```diff
@@ -75,4 +125,114 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+    fn collect_meta_votes(
+        &self,
+        cur_round: u32,
+        cur_step: u32,
+        voter: &S::PublicId,
+        event: &Event<T, S::PublicId>,
+        collections: &mut MetaVoteCollection<S::PublicId>,
+        last_consensed_events: &BTreeMap<S::PublicId, Event<T, S::PublicId>>,
+    ) {
+        if let Some(self_parent) = self.self_parent(event) {
```
Fraser999: Rather than duplicating these 25 lines, I'd create a new method, e.g.:
```rust
fn collect_parent_meta_votes(
    &self,
    cur_round: u32,
    cur_step: u32,
    voter: &S::PublicId,
    parent_event: Option<&Event<T, S::PublicId>>,
    collections: &mut MetaVoteCollection<S::PublicId>,
    last_consensed_events: &BTreeMap<S::PublicId, Event<T, S::PublicId>>,
) {
    if let Some(parent) = parent_event {
    ...
}
```
and call that twice from `collect_meta_votes()`.

PR #10, opened 2018-06-08
MAID-2766

maqi:
feat/parsec: implement functions of new, vote_for, create_gossip and have_voted_for

### DIFF
--- a/src/gossip/event.rs
+++ b/src/gossip/event.rs
@@ -97,6 +97,15 @@ impl<T: NetworkEvent, P: PublicId> Event<T, P> {
         }
     }
 
+    /// Returns `Some(vote)` if the event is for a vote of network event, otherwise returns `None`.
+    pub fn vote(&self) -> Option<&Vote<T, P>> {
+        if let Cause::Observation(ref vote) = self.content.cause {
+            Some(vote)
+        } else {
+            None
+        }
+    }
+
     pub fn creator(&self) -> &P {
         &self.content.creator
     }
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -120,7 +120,7 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
             *self_parent_hash,
             network_event,
         )?;
-        let _ = self.events.insert(event.hash().clone(), event);
+        let _ = self.events.insert(*event.hash(), event);
         Ok(())
     }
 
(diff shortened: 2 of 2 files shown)


src/parsec.rs
```diff
@@ -84,17 +85,48 @@ pub struct Parsec<T: NetworkEvent, S: SecretId> {
...
+                last_hash
+            } else {
+                return Err(Error::InvalidEvent);
+            };
+        let event = Event::new_from_observation(
+            self.peer_manager.our_id(),
+            *self_parent_hash,
+            network_event,
+        )?;
+        let _ = self.events.insert(event.hash().clone(), event);
```
ustulation: why clone() something that's copyable ? So `*event.hash()` should work. Does not clippy/linter complain about this ? If you agree can you pls check it elsewhere and correct it

src/parsec.rs
```diff
@@ -122,8 +154,19 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
     /// Check if the given `network_event` has already been voted for by us.
-    pub fn have_voted_for(&self, network_event: &T) -> bool {
-        unimplemented!();
+    pub fn have_voted_for(&self, network_event: &T) -> Result<bool, Error> {
+        let serialised_content = serialise(network_event)?;
+        let hash = Hash::from(serialised_content.as_slice());
+        Ok(
+            self.polled_blocks.contains(&hash) || self.consensused_blocks.iter().any(|block| {
+                if let Ok(serialised_payload) = serialise(block.payload()) {
+                    let payload_hash = Hash::from(serialised_payload.as_slice());
```
ustulation: instead of going through a hashing step can we not directly compare `serialised_content == serialised_payload` ?

src/parsec.rs
maqi: `*event.hash()` works, however clippy doesn't complain with this.
maybe because `event.hash()` is a ref, and we need to clone or de-ref anyway?

src/parsec.rs
```diff
@@ -122,8 +154,18 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
     }
 
     /// Check if the given `network_event` has already been voted for by us.
-    pub fn have_voted_for(&self, network_event: &T) -> bool {
-        unimplemented!();
+    pub fn have_voted_for(&self, network_event: &T) -> Result<bool, Error> {
```
Fraser999: For this method, shouldn't we only check in `self.events` for ones created by us rather than looking in `polled_blocks` or `consensused_blocks`?

maqi: ah, yes, good catch

src/gossip/event.rs
```diff
@@ -18,7 +18,7 @@ use std::collections::BTreeMap;
 use vote::Vote;
 
 pub(crate) struct Event<T: NetworkEvent, P: PublicId> {
-    content: Content<T, P>,
+    pub content: Content<T, P>,
```
Fraser999: Rather than exposing private gossip internals to the crate, can we add the following method to `Event`:
```rust
pub fn vote(&self) -> Option<&Vote<T, P>> {
    if let Cause::Observation(ref vote) = self.content.cause {
        Some(vote)
    } else {
        None
    }
}
```
and use that inside `Parsec::have_voted_for()`?

PR #11, opened 2018-06-13
Maid-2769 test framework

maqi:


src/tests.rs
```diff
@@ -0,0 +1,50 @@
...
+        assert!(!peer.have_voted_for(&transaction));
+        assert!(peer.vote_for(transaction.clone()).is_ok());
+    }
+
+    let mut consensued_peer = BTreeSet::new();
+    let mut iterations = 0;
+    loop {
+        network.send_random_syncs(&mut rng);
+        for (peer_id, peer) in network.peers_mut() {
+            if peer.poll().is_ok() {
```
Fraser999: `poll()` returns `Result<Option<Block<T, S::PublicId>>, Error>` - shouldn't we be unwrapping the result, and checking for the option `is_some()` to add the peer to `consensused_peers`?

src/test_utils.rs
```diff
@@ -0,0 +1,267 @@
...
+}
+
+impl PublicId for PeerId {
+    type Signature = sign::Signature;
+    fn verify_signature(&self, signature: &Self::Signature, data: &[u8]) -> bool {
+        sign::verify_detached(signature, data, &self.pk)
+    }
+}
+
+#[derive(Clone)]
```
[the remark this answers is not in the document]
maqi: will remove it.

tests/tests/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+            if let Ok(result) = peer.poll() {
+                if result.is_some() {
+                    let _ = consensued_peer.insert(*peer_id);
+                }
+            }
+        }
+        if consensued_peer.len() == num_peers {
+            break;
+        }
+        assert!(iterations < max_iterations);
```
ustulation: just a nitpick - iteration starts at the beginning of the loop, so this assertion should be there. Once you are here you have already completed that iteration. Also if you see it this way: if max_iteration = 1 this loop will have run twice the way it's done, if 0 it would have run once etc.

tests/tests/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+        assert!(!peer.have_voted_for(&transaction));
+        assert!(peer.vote_for(transaction.clone()).is_ok());
+    }
+
+    let mut consensued_peer = BTreeSet::new();
+    let mut iterations = 0;
+    loop {
+        network.send_random_syncs(&mut rng);
+        for (peer_id, peer) in &mut network.peers {
+            if let Ok(result) = peer.poll() {
```
Fraser999: I think we should unwrap the result here rather than silently swallowing the error.

tests/tests/utils.rs
```diff
@@ -0,0 +1,208 @@
...
+        Network { peers }
+    }
+
+    /// Node of `sender_id` sends a parsec request to the node of `receiver_id`, which causes
+    /// `receiver_id` node to reply with a parsec response.
+    pub fn send_sync(&mut self, sender_id: &PeerId, receiver_id: &PeerId) {
+        self.exchange_messages(sender_id, receiver_id);
+    }
+
+    /// Randomly choose two nodes, make one as node of `sender_id`, which sends a parsec request
```
Fraser999: It's not just a pair of nodes though.  We iterate through _all_ nodes, with each node sending a request to a random peer and handling the peer's response in each step.

maqi: the code itself only choose two nodes.
But I agree it shall be the way you described.
Will change the code and update the comment

tests/tests/utils.rs
```diff
@@ -48,17 +48,15 @@ impl Network {
...
+    /// For each node of `sender_id`, which sends a parsec request to a randomly chosen peer of
+    /// `receiver_id`, which causes `receiver_id` node to reply with a parsec response.
     pub fn send_random_syncs<R: Rng>(&mut self, rng: &mut R) {
         let peers: Vec<PeerId> = self.peers.keys().cloned().collect();
 
         for sender_id in &peers {
             while let Some(receiver_id) = rng.choose(&peers) {
                 if receiver_id != sender_id {
                     self.exchange_messages(sender_id, receiver_id);
-                    break;
```
Fraser999: Isn't that `break` required to exit the `while` loop?

Cargo.toml
```diff
@@ -16,3 +16,8 @@ serde = "~1.0.66"
 serde_derive = "~1.0.66"
 tiny-keccak = "~1.4.2"
 unwrap = "~1.1.0"
+
+[dev-dependencies]
+lazy_static = "~1.0.0"
+rand = "~0.4.2"
```
Fraser999: Can we use the latest version here since these are dev-dependencies?

tests/tests/utils.rs
```diff
@@ -0,0 +1,206 @@
...
+
+    /// For each node of `sender_id`, which sends a parsec request to a randomly chosen peer of
+    /// `receiver_id`, which causes `receiver_id` node to reply with a parsec response.
+    pub fn send_random_syncs<R: Rng>(&mut self, rng: &mut R) {
+        let peers: Vec<PeerId> = self.peers.keys().cloned().collect();
+
+        for sender_id in &peers {
+            while let Some(receiver_id) = rng.choose(&peers) {
+                if receiver_id != sender_id {
+                    self.exchange_messages(sender_id, receiver_id);
```
Fraser999: Don't we need a `break` here to exit the `while` loop?

Cargo.toml
maqi: we're tied to rand 0.4.2 by rust_sodium

pierrechevalier83:
Another git etiquette tip:
A commit should represent a self contained unit of work. It must be able to pass tests and code review on its own. These follow up commits that fixup the original commit should all actually be "fixed up".
Day to day, a convenient way to do this is to perform these two steps
```
git commit --fixup the-sha1-of-the-commit-im-fixing
git rebase -i --autosquash the-sha1-of-the-commit-im-fixing^
```
Could I ask you to squash all these commits into one?

pierrechevalier83:
I reviewed the code (as a diff with master) and have no objection so far. I'll have another look once things are squashed, but will probably be good to go.

pierrechevalier83:
As discussed, we're defering the git tips discussion to a better time. Looks good to me.

PR #12, opened 2018-06-14
Maid-2846 dump gossip graph

maqi:


src/id.rs
```diff
@@ -18,6 +18,8 @@ pub trait PublicId: Clone + Eq + Ord + Hash + Serialize + DeserializeOwned + Deb
     type Signature: Clone + Eq + Ord + Hash + Serialize + DeserializeOwned + Debug;
     /// Verifies `signature` against `data` using this `PublicId`.  Returns `true` if valid.
     fn verify_signature(&self, signature: &Self::Signature, data: &[u8]) -> bool;
+    /// Returns the first charater of the public id.
+    fn first_char(&self) -> char;
```
Fraser999: I definitely don't think we should have this method.  The `PublicId` trait already requires `Debug` - can't we just use the first char of the debug output instead?

src/tests.rs
```diff
@@ -32,7 +36,23 @@ fn minimal_network() {
         if consensued_peer.len() == num_peers {
             break;
         }
-        assert!(iterations < max_iterations);
+        if iterations > max_iterations {
```
Fraser999: I think this functionality would be best at the very least as a method of `Network` since it will get reused a lot.  I'd call it from `Network::drop()` if panicking, e.g:
```rust
impl Drop for Network {
    fn drop(&mut self) {
        if thread::panicking() {
            let folder_name = rand::thread_rng().gen_ascii_chars().take(6).collect::<String>();
            let dir = env::temp_dir().join("parsec_graphs").join(folder_name);
            if let Err(error) = fs::create_dir_all(&dir) {
                println!("Failed to create folder for dot files: {:?}", error);
            } else {
                println!("Writing dot files in {:?}", dir);
            }
            for (peer_id, peer) in &self.peers {
                let file_path = dir.join(format!("{:?}.dot", peer_id));
                if let Ok(mut file) = File::create(&file_path) {
                    let _ = write!(file, "{:?}", peer);
                } else {
                    println!("Failed to create {:?}", file_path);
                }
            }
        }
    }
}
```

src/dump_network.rs
```diff
@@ -0,0 +1,259 @@
...
+            if !positions.contains_key(hash) {
+                let self_parent_pos = if let Some(self_parent_hash) = event.self_parent() {
+                    if let Some(self_parent_pos) = positions.get(self_parent_hash) {
+                        *self_parent_pos
+                    } else {
+                        continue;
+                    }
+                } else {
+                    event.index.unwrap_or(0)
+                };
```
pierrechevalier83: Lines 168-176 are duplicated with 177-185. Could we make a function and deduplicate it? (takes a gossip event which is one of the parents. The hash can be deduced from the event, if I recall)

src/dump_network.rs
```diff
@@ -0,0 +1,259 @@
...
+
+    writeln!(f)
+}
+
+fn update_pos<T: NetworkEvent, S: SecretId>(
+    positions: &mut BTreeMap<Hash, u64>,
+    gossip_graph: &BTreeMap<Hash, Event<T, S::PublicId>>,
+) {
+    loop {
+        let mut updated = false;
```
pierrechevalier83: Style: I'm not a fan of these stateful "flags" kind of variables.
For instance here, I had rather see the for loop as a function that returns a bool.
Your outer loop would end up more idiomatic:
```
while set_next_pos(&mut self, ...) {
    // Keep going until there is nothing to set.
}
```

src/meta_vote.rs
```diff
@@ -39,6 +51,66 @@ pub(crate) struct MetaVote {
...
+
+        if self.estimates.is_empty() {
+            write!(f, "est:{{}}")?;
+        } else {
+            write!(f, "est:{{")?;
+            for estimate in &self.estimates {
+                if *estimate {
+                    write!(f, "t")?;
+                } else if self.estimates.len() > 1 {
+                    write!(f, "f,")?;
```
pierrechevalier83: This loop feels buggy. It feels like if true happend to be first, it could produce: "tf," instead of "t, f".
I think it would be much clearer if you had a function:
```
fn write_bool(f: Fmt, a_bool: bool) -> fmt::Result {
    match a_bool {
        true => write!(f, "t") ,
        true => write!(f, "f") ,
    }
}
```
Then, writing the estimates could look like this which is clearer:
```
write!(f, "est:{{")?;
if self.estimates.len() == 1 {
    write_bool(f, self.estimates[0])?;
} else if estimates.len() == 2 {
    write_bool(f, self.estimates[0])?;
    write!(f, ", ")?;
    write_bool(f, self.estimates[1])?;
}
write!(f, "}}")?;
}
```

Writing this, I'm realising: doesn't a BTreeSet<bool> loose the order? Maybe we should have used a Vec! to maintain the order as it's needed for the aux values computation. (I'll go check how we compute the auxiliary values outside of this code review)

src/meta_vote.rs
```diff
@@ -39,6 +51,66 @@ pub(crate) struct MetaVote {
...
+                } else {
+                    write!(f, "f")?;
+                }
+            }
+            write!(f, "}}, ")?;
+        }
+
+        if self.bin_values.is_empty() {
```
pierrechevalier83: Please, try and dedup what's common with est (after having refactored that one)

src/meta_vote.rs
```diff
@@ -39,6 +51,66 @@ pub(crate) struct MetaVote {
...
+            if aux_vote {
+                write!(f, "aux:{{t}}, ")?;
+            } else {
+                write!(f, "aux:{{f}}, ")?;
+            }
+        } else {
+            write!(f, "aux:{{}}, ")?;
+        }
+
+        if let Some(dec) = self.decision {
```
pierrechevalier83: lines 90 to 98 and 100 to 108 could also be a single function.

Like lines 59-111 could be:
```
write_multiple_bool_values(f, "est", self.estimates)?;
write_multiple_bool_values(f, "bin", self.bin_values)?;
write_optional_single_bool_value(f, "aux", self.aux_values)?;
write_optional_single_bool_value(f, "dec", self.decision)?;
```
This is the best way to avoid inconsistent formatting that always creeps up in these duplicated formatting type things...

src/dump_network.rs
```diff
@@ -0,0 +1,259 @@
...
+use hash::Hash;
+use id::SecretId;
+use meta_vote::MetaVote;
+use network_event::NetworkEvent;
+use std::cmp;
+use std::collections::BTreeMap;
+use std::fmt::{self, Debug, Formatter};
+
+fn first_char<D: Debug>(id: &D) -> char {
+    format!("{:?}", id).chars().next().unwrap_or('E')
```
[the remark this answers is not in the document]
maqi: It's only used during printout (and shall only for debug), so using `unwrap_or` instead of Option/Result just to avoid hassle of another round of checking outside.
The situation of `no char for id` shall not happen? and even it happens, a panic during debug print out (normally due to a panic during test) is not a good option ?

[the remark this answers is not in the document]
maqi: just read your other comment regarding `first_char`.
will change this to return Option, and using `unwrap_or` at the places calling it.

src/dump_network.rs
```diff
@@ -0,0 +1,259 @@
...
+        if let Some(self_parent) = event.self_parent() {
+            let parent = if let Some(parent) = gossip_graph.get(self_parent) {
+                parent
+            } else {
+                continue;
+            };
+            let event_pos = *positions.get(event.hash()).unwrap_or(&0);
+            let self_parent_pos = *positions.get(parent.hash()).unwrap_or(&0);
+            if event_pos <= (self_parent_pos + 1) {
+                writeln!(f, "    \"{:?}\" -> \"{:?}\"", self_parent, event.hash())?
```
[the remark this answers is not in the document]
maqi: will update it to have consistent output: printout minlen everywhere.

src/meta_vote.rs
maqi: `BTreeSet ` ensure the order of `f, t`

src/dump_network.rs
```diff
@@ -0,0 +1,267 @@
...
+}
+
+fn write_evaluates<T: NetworkEvent, S: SecretId>(
+    f: &mut Formatter,
+    gossip_graph: &BTreeMap<Hash, Event<T, S::PublicId>>,
+    meta_votes: &BTreeMap<Hash, BTreeMap<S::PublicId, Vec<MetaVote>>>,
+    initial_events: &[Hash],
+) -> fmt::Result {
+    for (event_hash, event) in gossip_graph.iter() {
+        if let Some(event_meta_votes) = meta_votes.get(event_hash) {
```
pierrechevalier83: else, could we still print things with a label: B_0, B_1,...?
It is currently inconsistent: hashes when no meta-votes, nice labels when contains meta votes

src/dump_network.rs
```diff
@@ -0,0 +1,267 @@
...
+
+                write!(f, " \"{:?}\" ", event.hash())?;
+                write!(
+                    f,
+                    " [label=\"{}_{}",
+                    first_char(event.creator()).unwrap_or('E'),
+                    event.index.unwrap_or(0)
+                )?;
+
+                for (peer, votes) in event_meta_votes.iter() {
```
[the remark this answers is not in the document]
maqi: they are already sorted by the order of peer_id.
the votes are in `vector` which I think is the `history order`, sorting them could lost the track of history?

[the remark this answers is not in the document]
pierrechevalier83: Uhm. Maybe I commented on the wrong line. I meant: 
```
D: [ { 0/0, est:{f} bin:{f} aux:{} dec:{} }]
C: [ { 0/0, est:{t} bin:{t} aux:{} dec:{} }]
B: [ { 0/0, est:{t} bin:{t} aux:{} dec:{} }]
A: [ { 0/0, est:{t} bin:{t} aux:{} dec:{} }]
```
would look better as
```
A: [ { 0/0, est:{t} bin:{t} aux:{} dec:{} }]
B: [ { 0/0, est:{t} bin:{t} aux:{} dec:{} }]
C: [ { 0/0, est:{t} bin:{t} aux:{} dec:{} }]
D: [ { 0/0, est:{f} bin:{f} aux:{} dec:{} }]
```

src/dump_network.rs
```diff
@@ -0,0 +1,271 @@
...
+}
+
+fn write_nodes<S: SecretId>(f: &mut Formatter, nodes: &[S::PublicId]) -> fmt::Result {
+    writeln!(f, "  {{")?;
+    writeln!(f, "    rank=same")?;
+    for node in nodes {
+        writeln!(f, "    {:?} [style=filled, color=white]", node)?;
+    }
+    writeln!(f, "  }}")?;
+    // Order the nodes alphabetically
```
pierrechevalier83: This is said, but not done ;)

PR #14, opened 2018-06-14
MAID-2774: implement "handle_request()" and "handle_response()"

Fraser999:


src/parsec.rs
```diff
@@ -739,4 +753,20 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
 
         self.peer_manager.is_super_majority(count)
     }
+
+    // Constructs a sync event to prove receipt of a `Request` or `Response` (depending on the value
```
maqi: shall we avoid using `sync` ?  maybe `gossip event` or something else?

src/parsec.rs
```diff
@@ -86,26 +84,42 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+        let other_parent = self.events.get(other_parent_hash).ok_or(Error::Logic)?;
+        let last_ancestors_hashes = other_parent
+            .last_ancestors
+            .iter()
+            .filter_map(|(peer_id, &index)| self.peer_manager.event_by_index(peer_id, index))
+            .collect::<BTreeSet<_>>();
+        let response_events_iter = self
+            .events_order
+            .iter()
+            .skip_while(|hash| !last_ancestors_hashes.contains(hash))
```
maqi: `last_ancestors_hashes` contains last hashes of all peer_ids, but skip_while will stop at the first hash we hit.
is it acceptable?

src/parsec.rs
Fraser999: Well, I wanted to convey in the comment and function name that it's _not_ for creating any kind of gossip event, but specifically for a request/response event only; not an initial or observation event.

src/parsec.rs
Fraser999: I think so.  This could be optimised in the future to avoid any potential duplicates, but I didn't feel that was important for milestone 1, especially given how inefficient the `Request` is.

Or are you thinking we're possibly missing some events to send?

src/parsec.rs
```diff
@@ -86,26 +84,42 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
     ) -> Result<Response<T, S::PublicId>, Error> {
-        unimplemented!();
+        for event in req.unpack() {
+            self.add_event(event)?;
+        }
+        self.create_sync_event(src, true)?;
+
+        // Return all the events we think `src` doesn't yet know about, packed into a `Response`.
+        let other_parent_hash = self.peer_manager.last_event_hash(src).ok_or(Error::Logic)?;
+        let other_parent = self.events.get(other_parent_hash).ok_or(Error::Logic)?;
```
pierrechevalier83: Just a general style comment because I've seen analogous situations in a few places.
When a variable is only needed for a small scope, I like to restrict the scope in which it leaves. It makes reading and later refactoring (for instance extracting functions) easier.
For instance here, this could be:
```
let other_parent = {
    let other_parent_hash = self.peer_manager.last_event_hash(src).ok_or(Error::Logic)?;
    self.events.get(other_parent_hash).ok_or(Error::Logic)?
};
```

Fraser999: I agree with all your points except probably the readability one, so I'll try and follow this principle from now on.

pierrechevalier83: I guess readibility is kinda subjective depending on which style you're most used too. I guess I find it more readable because the indentation is enough to let me know I don't have to worry about this variable anymore when reading the rest of the code.

src/parsec.rs
maqi: This is actually turned out to be a bug and be resolved in the later on commit.

PR #17, opened 2018-06-20
MAID-2778: add basic tests incorporating faulty nodes

Fraser999:


tests/integration_tests.rs
```diff
@@ -41,45 +42,168 @@ extern crate unwrap;
...
+    // Gossip to let all remaining peers reach consensus on all the blocks.
+    utils::loop_with_max_iterations(100, || {
+        network.send_random_syncs(&mut rng);
+        for peer in &mut network.peers {
+            peer.poll();
+        }
+        network
+            .peers
+            .iter()
+            .all(|peer| peer.blocks.len() >= num_transactions)
```
maqi: in what scenario will a peer having more blocks than the transactions voted?

tests/integration_tests.rs
```diff
@@ -41,45 +42,168 @@ extern crate unwrap;
...
+        }
+
+        if rng.gen_weighted_bool(10)
+            || network
+                .peers
+                .iter()
+                .any(|peer| peer.blocks.len() >= num_transactions / 2)
+        {
+            rng.shuffle(&mut network.peers);
+            network.peers.truncate(num_peers - num_faulty);
```
maqi: will this removal of faulty nodes be called multiple times?

tests/utils/id.rs
```diff
@@ -0,0 +1,106 @@
...
+        sign::sign_detached(data, &self.sk)
+    }
+}
+
+pub mod friendly_names {
+    use super::PeerId;
+    use std::collections::BTreeMap;
+    use std::sync::RwLock;
+
+    const NAMES: &[&str] = &["Alice", "Bob", "Carol", "Dave"];
```
maqi: do we need to expand this list to cover the case of network having 10 peers, and the graph is also reader friendly?

tests/utils/network.rs
```diff
@@ -0,0 +1,96 @@
...
+                    self.exchange_messages(sender_id, receiver_id);
+                    break;
+                }
+            }
+        }
+    }
+
+    /// Returns true if all peers hold the same sequence of stable blocks.
+    pub fn blocks_all_in_sequence(&self) -> bool {
+        let payloads = self.peers[0].blocks_payloads();
```
pierrechevalier83: Should we return a Result of bool and err if self.peers is empty?

pierrechevalier83: Judging by further code, it seems we don't mind panicking because it is test code. Feel free to ignore :)

tests/utils/id.rs
pierrechevalier83: Won't we need more than 4 for some basic tests?

pierrechevalier83: I'm thinking stuff like: a supermajority is one away from > 1/3 with 4, which is a bit edge-casey...

pierrechevalier83: Ninja-ed :smiley:

tests/utils/network.rs
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /// For each node of `sender_id`, which sends a parsec request to a randomly chosen peer of
+    /// `receiver_id`, which causes `receiver_id` node to reply with a parsec response.
+    pub fn send_random_syncs<R: Rng>(&mut self, rng: &mut R) {
+        let peer_ids: Vec<PeerId> = self.peers.iter().map(|peer| peer.id).collect();
+
+        for sender_id in &peer_ids {
+            while let Some(receiver_id) = rng.choose(&peer_ids) {
+                if receiver_id != sender_id {
```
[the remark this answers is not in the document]
Fraser999: Again I just moved this code rather than creating it... but even so, I'm not quite clear on your point.

Should I change it to something more like:
```rust
for sender_id in &peer_ids {
    let receiver_id = unwrap!(
        peer_ids
            .iter()
            .filter(|&id| id != sender_id)
            .nth(rng.gen_range(0, peer_ids.len() - 1))
    );
    self.exchange_messages(sender_id, receiver_id);
}
```

tests/utils/network.rs
```diff
@@ -0,0 +1,96 @@
...
+use std::collections::BTreeSet;
+
+pub struct Network {
+    pub peers: Vec<Peer>,
+}
+
+impl Network {
+    /// Create test network with the given initial number of peers (the genesis group).
+    pub fn new(count: usize) -> Self {
+        let all_ids: Vec<_> = (0..count).map(|_| FullId::new()).collect();
```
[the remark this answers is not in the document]
pierrechevalier83: My bad! My suggestion was buggy. Please ignore it ;)

PR #20, opened 2018-06-21
MAID-2777: add tests for multiple votes applied before gossiping

Fraser999:


tests/integration_tests.rs
```diff
@@ -78,6 +78,70 @@ fn minimal_network() {
     }
 }
 
+#[ignore]
+#[test]
+fn multiple_votes_before_gossip() {
+    let mut rng = utils::init_rng(None);
+    let num_peers = 4;
```
maqi: just a minor comment, we have many tests having the same code to setup network with same number of peers, then populate transactions,
and the functor passed to loop_with_max_iteration all mostly they same.
does it worth to de-duplicate them?

Fraser999: Yes, I'm all for deduplicating where possible.  I'm not sure there's much we can do about the closures passed to the loop - I think they're different enough to not need changed.  But I was toying with adding a new struct like `Environment` or something that would construct (and hold) the rng, the `Network` and the transactions.

If others are in favour of this change, I'm happy to add it to this PR.

pierrechevalier83: I'm in favour :)

PR #23, opened 2018-06-22
MAID-2767: add basic example

Fraser999:


examples/basic.rs
```diff
@@ -0,0 +1,297 @@
...
+        Ok(count) if count >= MIN_PEER_COUNT && count <= NAMES.len() => count,
+        _ => {
+            println!("'{}' {}", PEERS, peers_info);
+            process::exit(-2);
+        }
+    };
+    println!(
+        "Running example with {} random network event{} and {} peers...",
+        event_count,
+        if event_count > 1 { "s" } else { "" },
```
pierrechevalier83: Nice detail :)

examples/basic.rs
```diff
@@ -0,0 +1,297 @@
...
+    let (event_count, peer_count) = get_params();
+
+    // Set up the requested number of peers and random network events.
+    let all_ids: Vec<_> = NAMES
+        .iter()
+        .take(peer_count)
+        .cloned()
+        .map(PeerId::new)
+        .collect();
+    let genesis_group: BTreeSet<_> = all_ids.iter().cloned().collect();
```
pierrechevalier83: This one is really nitpicky, but we're inconsistent between specifying the type in collect (`collect::<BTreeSet<_>>()`) and specifying it where we declare the variable.

I would be in favour to consistently specify the type in collect and let normal type deduction apply for the variable (seems slightly more idiomatic: applies in more contexts).

This is so nitpicky, though I'm not asking for an update of this commit. Just something to keep in mind. (Except if there is a reason for the inconsistency that I'm missing ?)

examples/basic.rs
```diff
@@ -0,0 +1,297 @@
...
+        rng.shuffle(&mut peers);
+        // Each peer will send a request and handle the corresponding response.  For each peer,
+        // there is also a chance that they will vote for one of the transactions if they haven't
+        // already done so.
+        for sender_index in 0..peers.len() {
+            let receiver_index = if sender_index == peers.len() - 1 {
+                0
+            } else {
+                sender_index + 1
+            };
```
pierrechevalier83:
```
let receiver_index = (sender_index + 1) % peers.len();
```

examples/basic.rs
```diff
@@ -0,0 +1,297 @@
...
+            let receiver_id = peers[receiver_index].id.clone();
```
pierrechevalier83: nitpick: I would start with a couple more variables:
```
let sender = peers[sender_index];
let receiver = peers[receiver_index];
```
I think it would read a bit better if I didn't have to re-parse `peers[receiver_index]` for a number of lines in a row...

examples/basic.rs
```diff
@@ -0,0 +1,297 @@
...
+            println!("{:?}", peer);
+        }
+
+        if peers.iter().all(|peer| peer.blocks.len() == event_count) {
+            break;
+        } else if iteration == max_iterations - 1 {
+            println!(
+                "\n!!! Failed to reach consensus within {} iterations... giving up !!!",
+                max_iterations
+            );
```
pierrechevalier83: Did you forget to break here?

examples/basic.rs
```diff
@@ -0,0 +1,297 @@
...
+
+    let mut rng = SeededRng::new();
+    let mut transactions = vec![];
+    while transactions.len() < event_count {
+        transactions.push(Transaction(
+            rng.gen_ascii_chars().take(5).collect::<String>(),
+        ));
+    }
+
+    let max_iterations = 1000;
```
pierrechevalier83: Should we also take max_iterations from the CLI, so that if it's specified we can run for less iterations?

Fraser999: Yes - I think that would be helpful.

examples/basic.rs
Fraser999: No, this is a for loop, so it'll end at just after this prints anyway.

pierrechevalier83: duh :man_facepalming:

PR #24, opened 2018-06-25
MAID-2864: Fix set valid blocks carried

pierrechevalier83:
This is one of the bugs found when testing the integration tests (MAID-2864)

### DIFF
--- a/src/dump_network.rs
+++ b/src/dump_network.rs
@@ -162,6 +162,13 @@ fn write_evaluates<T: NetworkEvent, S: SecretId>(
                 event.hash()
             )?;
         }
+        for (event_hash, _payload) in event.valid_blocks_carried.values() {
+            writeln!(
+                f,
+                " \"{:?}\" [shape=rectangle, style=filled, fillcolor=crimson]",
+                event_hash
+            )?;
+        }
         if meta_votes.get(event_hash).is_some() {
             writeln!(f, " \"{:?}\" [shape=rectangle]", event.hash())?;
         }
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -307,118 +307,111 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
     }
 
     fn set_valid_blocks_carried(&mut self, event: &mut Event<T, S::PublicId>) -> Result<(), Error> {
-        // If this node already has meta votes running, no need to calculate `valid_blocks_carried`.
-        if self
-            .self_parent(event)
-            .and_then(|parent| self.meta_votes.get(parent.hash()))
-            .is_some()
-        {
-            return Ok(());
+        // If my self_parent already carries a valid block for this peer, use it
+        if let Some(self_parent) = self.self_parent(event) {
+            event
+                .valid_blocks_carried
+                .append(&mut self_parent.valid_blocks_carried.clone())
         }
-        // Union of
-        // * my self_parent's valid_blocks_carried,
-        // * my other_parent's valid_blocks_carried,
-        // * any observation I made that makes a block valid
-        // Note that if any block becomes stable due to this gossip event, we will learn
-        // that fact later and prune valid_blocks_carried accordingly
-        let self_parents_valid_blocks = {
-            if let Some(self_parent) = self.self_parent(event) {
-                self_parent.valid_blocks_carried.clone()
-            } else {
-                BTreeSet::new()
-            }
-        };
-        let other_parents_valid_blocks = {
-            if let Some(other_parent) = self.other_parent(event) {
-                other_parent.valid_blocks_carried.clone()
-            } else {
-                BTreeSet::new()
-            }
-        };
-        let blocks_made_valid_now = {
-            if let Some(vote) = event.vote() {
-                let mut valid_votes: BTreeSet<Hash> = BTreeSet::new();
-                let n_same_votes = self
-                    .events
+        // If my other_parent already carries a valid block for this peer, use it
+        if let Some(other_parent) = self.other_parent(event) {
+            event
+                .valid_blocks_carried
+                .append(&mut other_parent.valid_blocks_carried.clone())
+        }
+        // If none of my parents already carries a valid block for this peer, see if this event
+        // makes a block valid
+        if event.valid_blocks_carried.get(event.creator()).is_none() {
+            let valid_block_carried = {
+                let payloads_made_valid = self
+                    .peer_manager
                     .iter()
-                    .filter(|(_, value)| {
-                        (**value)
-                            .vote()
-                            .map_or(false, |iter_vote| iter_vote.payload() == vote.payload())
+                    .flat_map(|(_peer, events)| {
+                        events.iter().filter_map(|(_index, event_hash)| {
+                            self.events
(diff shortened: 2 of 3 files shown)


src/dump_network.rs
```diff
@@ -162,6 +162,18 @@ fn write_evaluates<T: NetworkEvent, S: SecretId>(
                 event.hash()
             )?;
         }
+        let self_parent_carries_no_blocks = event.self_parent().map_or(true, |self_parent_hash| {
+            gossip_graph[self_parent_hash]
```
maqi: I think we shall avoid this kind of direct indexing (which causes panic when in problem) in the production code as much as possible?

src/parsec.rs
```diff
@@ -332,23 +324,23 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-                let n_same_votes = self
-                    .events
-                    .iter()
-                    .filter(|(_, value)| {
-                        (**value)
-                            .vote()
-                            .map_or(false, |iter_vote| iter_vote.payload() == vote.payload())
+            let n_valid_blocks = self
+                .peer_manager
+                .iter()
```
maqi: can we just use `values()` ? and same comment to the below

src/parsec.rs
```diff
@@ -332,23 +324,23 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+                .flat_map(|(_peer, events)| {
+                    events.iter().filter_map(|(_index, event_hash)| {
+                        self.events
+                            .get(event_hash)
+                            .map(|event| event.vote().map(|vote| vote.payload().clone()).clone())
```
Fraser999: If `n_ancestors_carrying_payload()` is changed to take a `payload: &Option<&T>`, we can avoid cloning the payload here.

src/parsec.rs
```diff
@@ -363,6 +355,32 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+                    match self.events.get(event_hash) {
+                        Some(that_event) => {
+                            (*payload == that_event.vote().map(|vote| vote.payload().clone()))
+                                && (last_ancestor_index.map_or(false, |last_index| {
+                                    that_event
+                                        .index
+                                        .map_or(false, |that_index| *last_index >= that_index)
+                                }))
+                        }
+                        None => false,
```
maqi: this sounds more like an error instead of a logic false?

src/parsec.rs
```diff
@@ -332,23 +324,23 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-                }
-                valid_votes
+                })
+                .filter(|this_payload| {
+                    self.peer_manager
+                        .is_super_majority(self.n_ancestors_carrying_payload(event, this_payload))
+                })
+                .count();
+            if n_valid_blocks > 0 {
+                vec![*event.hash()].into_iter().collect()
```
Fraser999: A more idiomatic approach might be
```rust
iter::once(*event.hash()).collect()
```

src/parsec.rs
Fraser999: We'd need to add that as a function to `PeerManager`, and it wouldn't be useful in the other instance below since we use the key and value in that case.  Not sure it's worth it for just one instance?

src/parsec.rs
```diff
@@ -332,23 +324,23 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
             }
         };
         let blocks_made_valid_now = {
-            if let Some(vote) = event.vote() {
```
maqi: previously, blocks_made_valid_now at https://github.com/maidsafe/parsec/blob/master/src/parsec.rs#L334  contains the hash of events that casting the vote.
but with your change, it now contains the hash of event when such block is seen.  https://github.com/pierrechevalier83/parsec/blob/fix_set_valid_blocks_carried/src/parsec.rs#L343
which is problematic as the later on checks relies on the valid_blocks_carried pointing to events of voting.

pierrechevalier83: As discussed, we want to bundle the payload with the hash of the gossip event for each of these.

src/parsec.rs
```diff
@@ -303,66 +304,88 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+        payload: &Option<&T>,
+    ) -> usize {
+        self.peer_manager
+            .iter()
+            .filter(|(peer, events)| {
+                events.iter().any(|(_index, event_hash)| {
+                    let last_ancestor_index = &event.last_ancestors.get(peer);
+                    match self.events.get(event_hash) {
+                        Some(that_event) => {
+                            (*payload == that_event.vote().map(|vote| vote.payload()))
```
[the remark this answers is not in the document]
pierrechevalier83: Does last_ancestors exclude itself?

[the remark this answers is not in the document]
maqi: doesn't matter, as the event has got inserted into self.events yet, so definitely it won't get counted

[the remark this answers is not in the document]
pierrechevalier83: Uhm. It does matter: it's the difference between having to ignore my last_ancestor created by the same event as my creator or not...

src/parsec.rs
```diff
@@ -800,20 +769,27 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         Ok(())
     }
 
-    // Returns whether event X can strongly see the event Y.
-    fn does_strongly_see(&self, x: &Event<T, S::PublicId>, y: &Event<T, S::PublicId>) -> bool {
-        let count = y
-            .first_descendants
+    // Returns the number of peers through which their is a directed path in the gossip graph between event X and event Y.
```
Fraser999: "their" → "there"

src/parsec.rs
```diff
@@ -665,74 +657,72 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
                             }
                         })
-                        .collect::<Vec<_>>();
-                    // We sort the events by their vote to avoid ties when picking the event
+                        .cloned()
+                        .collect::<Vec<(Hash, T)>>();
+                    // We sort the events by their payloads to avoid ties when picking the event
                     // with the most represented payload.
-                    // Because `max_by` guarentees that "If several elements are equally
+                    // Because `max` guarentees that "If several elements are equally
```
Fraser999: "guarentees" → "guarantees"

src/parsec.rs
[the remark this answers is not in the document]
maqi: the last_ancestor doesn't exclude itself.

However, the problem here is that the counter is calculated from the events exists in the self.events.
last_ancestor is only used to ensure the index and because we are using `>=`, so it won't affect the counter issue here.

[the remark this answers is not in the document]
pierrechevalier83: OK. I think I get what's going on now and I think my fixup commit is correct. Could you have a look and let me know if you agree?

src/parsec.rs
pierrechevalier83: As discussed in slack with Fraser, it seems that this could happen as `event` is present in `peer_manager` but hasn't yet been added to `self.events`.

PR #25, opened 2018-06-26
MAID-2864: initialise 'Parsec::round_hashes' properly in its constructor

Fraser999:


src/parsec.rs
```diff
@@ -45,8 +45,12 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         let responsiveness_threshold = (genesis_group.len() as f64).log2().ceil() as usize;
 
         let mut peer_manager = PeerManager::new(our_id);
-        for peer_id in genesis_group.iter() {
+        let mut round_hashes = BTreeMap::new();
+        let initial_hash = Hash::from([].as_ref());
+        for peer_id in genesis_group.iter().cloned() {
             peer_manager.add_peer(peer_id.clone());
+            let round_hash = RoundHash::new(&peer_id, initial_hash)?;
```
maqi: isn't this `initial_hash` shall be the hash of the initial_event or the hash of the `latest_valid_block`?
the RoundHash::new function expect that parameter to be : `latest_block_hash` for round 0

Fraser999: Yes, it is normally.  But at the start when we don't have any `latest_valid_block`, we just all need to use the same value.  It _could_ be the hash of the payload of the initial event, but it doesn't really matter what we use.  I think we agreed to use the hash of an empty slice (should be pretty much optimal).

PR #26, opened 2018-06-27
MAID-2864: update meta-votes properly

Fraser999:
When meta-votes are updated, their own changes are now considered at each stage of the calculation.  This also ensures that once a decision has been made at a given round and step, it persists from then on.

### DIFF
--- a/src/meta_vote.rs
+++ b/src/meta_vote.rs
@@ -118,7 +118,7 @@ impl MetaVote {
             .iter()
             .map(|meta_vote| {
                 let counts = MetaVoteCounts::new(meta_vote, others, total_peers);
-                Self::update_meta_vote(meta_vote, &counts, &coin_tosses)
+                Self::update_meta_vote(meta_vote, counts, &coin_tosses)
             })
             .collect::<Vec<_>>();
         while let Some(next_meta_vote) =
@@ -131,18 +131,24 @@ impl MetaVote {
 
     fn update_meta_vote(
         meta_vote: &MetaVote,
-        counts: &MetaVoteCounts,
+        mut counts: MetaVoteCounts,
         coin_tosses: &BTreeMap<usize, bool>,
     ) -> MetaVote {
-        let coin_toss = coin_tosses.get(&meta_vote.round);
-        MetaVote {
-            round: meta_vote.round,
-            step: meta_vote.step.clone(),
-            estimates: Self::calculate_new_estimates(&meta_vote, &counts, coin_toss),
-            bin_values: Self::calculate_new_bin_values(&meta_vote, &counts),
-            aux_value: Self::calculate_new_auxiliary_value(&meta_vote),
-            decision: Self::calculate_new_decision(&meta_vote, &counts),
+        if meta_vote.decision.is_some() {
+            return meta_vote.clone();
         }
+        let coin_toss = coin_tosses.get(&meta_vote.round);
+        let mut updated_meta_vote = meta_vote.clone();
+        Self::calculate_new_estimates(&mut updated_meta_vote, &mut counts, coin_toss);
+        let bin_values_was_empty = updated_meta_vote.bin_values.is_empty();
+        Self::calculate_new_bin_values(&mut updated_meta_vote, &mut counts);
+        Self::calculate_new_auxiliary_value(
+            &mut updated_meta_vote,
+            &mut counts,
+            bin_values_was_empty,
+        );
+        Self::calculate_new_decision(&mut updated_meta_vote, &counts);
+        updated_meta_vote
     }
 
     fn next_meta_vote(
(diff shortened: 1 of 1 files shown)


src/meta_vote.rs
```diff
@@ -131,18 +131,24 @@ impl MetaVote {
...
-            bin_values: Self::calculate_new_bin_values(&meta_vote, &counts),
-            aux_value: Self::calculate_new_auxiliary_value(&meta_vote),
-            decision: Self::calculate_new_decision(&meta_vote, &counts),
+        if meta_vote.decision.is_some() {
+            return meta_vote.clone();
         }
+        let coin_toss = coin_tosses.get(&meta_vote.round);
+        let mut updated_meta_vote = meta_vote.clone();
+        Self::calculate_new_estimates(&mut updated_meta_vote, &mut counts, coin_toss);
+        let bin_values_was_empty = updated_meta_vote.bin_values.is_empty();
```
maqi: we better move this line of code before `calculate_new_estimates`, and use `meta_vote.bin_values.is_empty()` to match the name of `was_empty` more explicitly?

src/meta_vote.rs
```diff
@@ -131,18 +131,24 @@ impl MetaVote {
...
+        Self::calculate_new_bin_values(&mut updated_meta_vote, &mut counts);
+        Self::calculate_new_auxiliary_value(
+            &mut updated_meta_vote,
+            &mut counts,
+            bin_values_was_empty,
+        );
+        Self::calculate_new_decision(&mut updated_meta_vote, &counts);
```
maqi: just to confirm that the `counts` here could have been modified during previous calculate new est/bin/aux .
it is expected, right?

src/meta_vote.rs
```diff
@@ -165,82 +174,88 @@ impl MetaVote {
...
-        } else if meta_vote.aux_value.is_none() && meta_vote.bin_values.is_empty() {
+    fn calculate_new_auxiliary_value(
+        meta_vote: &mut MetaVote,
+        counts: &mut MetaVoteCounts,
+        bin_values_was_empty: bool,
+    ) {
+        if meta_vote.aux_value.is_some() {
+            return;
+        }
+        if bin_values_was_empty {
```
maqi: hmm...   so when bin_values was not empty (but the previous aux_value is empty), then we do nothing?

src/meta_vote.rs
```diff
@@ -165,82 +174,88 @@ impl MetaVote {
...
-                } else {
-                    None
-                },
-                Step::GenuineFlip => None,
-            }
-        } else {
-            None
+    fn calculate_new_decision(meta_vote: &mut MetaVote, counts: &MetaVoteCounts) {
+        let opt_decision = match meta_vote.step {
+            Step::ForcedTrue => if meta_vote.bin_values.contains(&true)
```
maqi: so the `step` is from the previous meta_vote, meanwhile the `bin_values` & count is from the new calculated.
it is expected, right?

src/meta_vote.rs
Fraser999: Well, to keep the scope of the variable down to the minimum, I put it at the latest viable point just before it's needed.  I'm not quite sure about your other point.  Are you thinking I should rename it to `bin_values_is_empty`?

src/meta_vote.rs
Fraser999: Yes - `counts` could have been modified by any of the three previous functions.

src/meta_vote.rs
Fraser999: Yes, I think so; that step also applies to this meta-vote.  We'd only update the step later in the call to `next_meta_vote()`.

src/meta_vote.rs
maqi: what I mean is `bin_values_was_empty` indicates the status is deduced from the `previous` meta_vote, but here we use the `updated_meta_vote` to calculate it (although the content is the same.)

pierrechevalier83: My understanding is that we record if it *was* empty before we called `calculate_new_bin_values` as this information is useful for calculating auxiliary values.

src/meta_vote.rs
```diff
@@ -131,18 +131,24 @@ impl MetaVote {
...
-            decision: Self::calculate_new_decision(&meta_vote, &counts),
+        if meta_vote.decision.is_some() {
+            return meta_vote.clone();
         }
+        let coin_toss = coin_tosses.get(&meta_vote.round);
+        let mut updated_meta_vote = meta_vote.clone();
+        Self::calculate_new_estimates(&mut updated_meta_vote, &mut counts, coin_toss);
+        let bin_values_was_empty = updated_meta_vote.bin_values.is_empty();
+        Self::calculate_new_bin_values(&mut updated_meta_vote, &mut counts);
+        Self::calculate_new_auxiliary_value(
```
maqi: the `calculate_new_auxiliary_value` will only be called once at https://github.com/Fraser999/parsec/blob/update_meta_vote/src/meta_vote.rs#L145
which means the `counts.aux_values_true` will only be 0 or 1 at https://github.com/Fraser999/parsec/blob/update_meta_vote/src/meta_vote.rs#L224 .
However, when we try to calculate decision at https://github.com/Fraser999/parsec/blob/update_meta_vote/src/meta_vote.rs#L239, we expect the value of `counts.aux_values_true` becomes super_majority (7 out of 10), which always fails.

PR #30, opened 2018-06-28
Maid 2864 print more graphs

pierrechevalier83:


pierrechevalier83:
@Fraser999 , I've implemented your suggestion 1 because it's small enough that we may as well do it now.
I'll keep suggestion 2 in mind for a future PR (going to focus on strictly Milestone 1 work for now :) )

PR #31, opened 2018-06-28
fix/parsec: fix infinite loop bug in 'stop_waiting'

Fraser999:


src/parsec.rs
```diff
@@ -564,19 +564,26 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         event: &Event<T, S::PublicId>,
     ) -> bool {
         let mut response_count = 0;
-        let mut event_hash = *event.hash();
-        while response_count < self.responsiveness_threshold {
-            if let Some(evnt) = self.self_parent(event) {
```
pierrechevalier83: OK. I see how this was causing an infinite loop since it was always taking the self_parent of the same event which was immutable...

src/parsec.rs
```diff
@@ -564,19 +564,26 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+        let mut event_hash = event.self_parent();
```
pierrechevalier83: However, I don't see why we changed the initial hash from *event.hash() to self_parent(). Could you explain the reasoning?

Fraser999: It wasn't actually being used at all until we bailed out of the while loop after hitting the required number of response events.  After leaving the loop, it represented the `responsiveness_threshold`-th response event, which is the event we want to consider when settling the coin toss under non-responsive leader scenario.

I've changed this so we actually _do_ use the event hash as a means of iterating backwards through the events, but retained the behaviour that it represents the `responsiveness_threshold`-th response event on leaving the loop.

PR #32, opened 2018-06-29
MAID-2864: fix/parsec: consensused payloads are uninteresting

pierrechevalier83:
Once a block with a given payload already reached consensus, never
consider this payload again as a candidate for consensus.


src/parsec.rs
```diff
@@ -363,6 +363,13 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         event: &Event<T, S::PublicId>,
         payload: &Option<&T>,
     ) -> usize {
+        if self
+            .consensused_blocks
```
maqi: consensused_blocks will be pop_front when poll called, so the check here is no-help.
there shall be some other cache to remember `history consensued blocks` ?

pierrechevalier83: Great catch, @maqi ! Addressing this now :)

PR #34, opened 2018-07-02
fix/parsec.git: update all meta_votes

maqi:
When update meta_votes of a peer due to received new events, the
new updated meta_votes shall be taken into account as well.

### DIFF
--- a/src/meta_vote.rs
+++ b/src/meta_vote.rs
@@ -114,13 +114,17 @@ impl MetaVote {
         coin_tosses: &BTreeMap<usize, bool>,
         total_peers: usize,
     ) -> Vec<Self> {
-        let mut next = parent
-            .iter()
-            .map(|meta_vote| {
-                let counts = MetaVoteCounts::new(meta_vote, others, total_peers);
-                Self::update_meta_vote(meta_vote, counts, &coin_tosses)
-            })
-            .collect::<Vec<_>>();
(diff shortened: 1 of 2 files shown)


src/meta_vote.rs
```diff
@@ -114,13 +114,17 @@ impl MetaVote {
...
-            })
-            .collect::<Vec<_>>();
+        let mut next = Vec::new();
+        for vote in parent {
+            let counts = MetaVoteCounts::new(vote, others, total_peers);
+            let updated = Self::update_meta_vote(vote, counts, &coin_tosses);
+            let changed = &updated != vote;
+            next.push(updated);
+            if changed {
+                break;
```
pierrechevalier83: I don't get that: isn't it possible that many of the meta votes get updated? What happens if we break out of the loop then?

If you're trying to fix the issue with a decision being taken, maybe break if a decision was reached instead?

maqi: I think the later on calls to the `next` shall generate new meta_votes ?

pierrechevalier83: OK. In that case, why do we need that loop at all? What if we only called next for all of them?

maqi: 1. you can't remove an entry by calling next
2. next of vote_2 doesn't equal to next of vote_2' (which is a next from updated vote_1)

PR #37, opened 2018-07-03
MAID-2864: fix/parsec: fix receeding consensus

pierrechevalier83:
The valid blocks carried don't get recalculated on old gossip events
when clearing the consensus data, which causes consensus to occur later
and later as old events aren't ever again considered as interesting
events.

### DIFF
--- a/src/dump_graph.rs
+++ b/src/dump_graph.rs
@@ -230,17 +230,27 @@ mod detail {
         initial_events: &[Hash],
     ) -> io::Result<()> {
         for (event_hash, event) in gossip_graph.iter() {
-            // Write the label
-            write!(writer, " \"{:?}\" ", event.hash())?;
+            write!(writer, " \"{:?}\" [", event.hash())?;
+            if meta_votes.contains_key(event_hash) {
+                write!(writer, " shape=rectangle, ")?;
+            }
             write!(
                 writer,
-                " [label=\"{}_{}",
+                "fillcolor=white, label=\"{}_{}",
                 first_char(event.creator()).unwrap_or('E'),
                 event.index.unwrap_or(0)
             )?;
+
             if let Some(event_payload) = event.vote().map(|vote| vote.payload()) {
                 write!(writer, "\n{:?}", event_payload)?;
             }
+
+            // Write the `valid_blocks_carried` if have
+            if !event.valid_blocks_carried.is_empty() {
+                write!(writer, "\n{:?}", event.valid_blocks_carried)?;
+            }
+
+            // Write the `meta_votes` if have
             if let Some(event_meta_votes) = meta_votes.get(event_hash) {
                 if event_meta_votes.len() >= initial_events.len() {
                     let mut peer_ids: Vec<&P> = event_meta_votes.keys().collect();
@@ -266,26 +276,18 @@ mod detail {
             }
             writeln!(writer, "\"]")?;
             // Add any styling
-            if event.vote().is_some() {
+            if !event.valid_blocks_carried.is_empty() {
+                writeln!(
+                    writer,
+                    " \"{:?}\" [shape=rectangle, style=filled, fillcolor=crimson]",
+                    event.hash()
+                )?;
+            } else if event.vote().is_some() {
                 writeln!(
                     writer,
                     " \"{:?}\" [shape=rectangle, style=filled, fillcolor=cyan]",
                     event.hash()
                 )?;
-            }
-            for (event_hash, payload) in event.valid_blocks_carried.values() {
-                let that_event = &gossip_graph[event_hash];
-                writeln!(
-                writer,
-                " \"{:?}\" [shape=rectangle, style=filled, fillcolor=crimson, label=\"{}_{}\n{:?}\"]",
-                event_hash,
-                first_char(that_event.creator()).unwrap_or('Z'),
-                that_event.index.unwrap_or(0),
-                payload
-            )?;
-            }
-            if meta_votes.get(event_hash).is_some() {
-                writeln!(writer, " \"{:?}\" [shape=rectangle]", event.hash())?;
             }
         }
 
--- a/src/gossip/event.rs
+++ b/src/gossip/event.rs
@@ -29,9 +29,8 @@ pub(crate) struct Event<T: NetworkEvent, P: PublicId> {
     pub last_ancestors: BTreeMap<P, u64>,
     // Index of each peer's earliest event that is a descendant of this event.
     pub first_descendants: BTreeMap<P, u64>,
-    // The hashes of the oldest events by each peer which comprise not-yet-stable blocks this event can see,
-    // along with the largest payload of all blocks made valid by that event.
-    pub valid_blocks_carried: BTreeMap<P, (Hash, T)>,
+    // Payloads of all the blocks made valid by this event
+    pub valid_blocks_carried: BTreeSet<T>,
     // The set of peers for which this event can strongly-see an event by that peer which carries a
     // valid block.  If there are a supermajority of peers here, this event is an "observer".
     pub observations: BTreeSet<P>,
@@ -93,7 +92,7 @@ impl<T: NetworkEvent, P: PublicId> Event<T, P> {
             index: None,
             last_ancestors: BTreeMap::default(),
             first_descendants: BTreeMap::default(),
-            valid_blocks_carried: BTreeMap::default(),
+            valid_blocks_carried: BTreeSet::default(),
             observations: BTreeSet::default(),
         })
     }
@@ -167,7 +166,7 @@ impl<T: NetworkEvent, P: PublicId> Event<T, P> {
             index: None,
             last_ancestors: BTreeMap::default(),
             first_descendants: BTreeMap::default(),
-            valid_blocks_carried: BTreeMap::default(),
+            valid_blocks_carried: BTreeSet::default(),
             observations: BTreeSet::default(),
         })
     }
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -18,6 +18,7 @@ use network_event::NetworkEvent;
 use peer_manager::PeerManager;
 use round_hash::RoundHash;
 use std::collections::{BTreeMap, BTreeSet, VecDeque};
+use std::iter;
 
 /// The main object which manages creating and receiving gossip about network events from peers, and
 /// which provides a sequence of consensused `Block`s by applying the PARSEC algorithm.
@@ -28,6 +29,8 @@ pub struct Parsec<T: NetworkEvent, S: SecretId> {
     events: BTreeMap<Hash, Event<T, S::PublicId>>,
     // The sequence in which all gossip events were added to this `Parsec`.
     events_order: Vec<Hash>,
+    // The hashes of events for each peer that have a non-empty set of `valid_blocks_carried`
+    events_with_valid_blocks: BTreeMap<S::PublicId, VecDeque<Hash>>,
     // Consensused network events that have not been returned via `poll()` yet.
     consensused_blocks: VecDeque<Block<T, S::PublicId>>,
     // Hash of all payloads that were consensused ever
@@ -58,6 +61,7 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
             peer_manager,
             events: BTreeMap::new(),
             events_order: vec![],
+            events_with_valid_blocks: BTreeMap::new(),
             consensused_blocks: VecDeque::new(),
             consensus_history: vec![],
             meta_votes: BTreeMap::new(),
@@ -202,11 +206,11 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         self.peer_manager.add_event(&event)?;
         self.set_last_ancestors(&mut event)?;
         self.set_first_descendants(&mut event)?;
-        self.set_valid_blocks_carried(&mut event)?;
 
         let event_hash = *event.hash();
         self.events_order.push(event_hash);
         let _ = self.events.insert(event_hash, event);
+        self.set_valid_blocks_carried(&event_hash)?;
         self.process_event(&event_hash)
     }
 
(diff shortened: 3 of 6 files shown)


pierrechevalier83:
Note: it's a Work In Progress for now as it causes 3 tests to hang. I still raised a PR to avoid forgetting about the change.

src/parsec.rs
```diff
@@ -703,49 +709,53 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
                             }
                         })
                         .cloned()
-                        .collect::<Vec<(Hash, T)>>();
+                        .collect::<Vec<(Hash, Vec<T>)>>();
                     // We sort the events by their payloads to avoid ties when picking the event
                     // with the most represented payload.
                     // Because `max` guarantees that "If several elements are equally
```
Fraser999: This comment can probably be removed now?

src/parsec.rs
```diff
@@ -761,7 +771,13 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
                 .valid_blocks_carried
                 .clone()
                 .into_iter()
-                .filter(|(_peer, (_hash, this_payload))| payload != this_payload)
+                .map(|(peer, (hash, mut these_payloads))| {
+                    if Some(payload) == these_payloads.last() {
```
Fraser999: Shouldn't we find and remove the specific payload passed into this method rather than just `these_payloads.last()`?

src/parsec.rs
```diff
@@ -347,13 +347,19 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
                         }
-                    });
-                payloads_made_valid.max().map(|max_payload_made_valid| {
-                    (
+                    })
+                    .cloned()
+                    .collect::<Vec<_>>();
+                payloads_made_valid.sort();
+                payloads_made_valid.dedup();
+                if payloads_made_valid.is_empty() {
```
maqi: can we do:
```
if payloads_made_valid.is_empty() {
    None
} else {
    payloads_made_valid.sort();
    payloads_made_valid.dedup();
    Some((
        event.creator().clone(),
        (*event.hash(), payloads_made_valid),
    ))
}
```

src/parsec.rs
pierrechevalier83: It's sorted, so I thought this would be fine. Will double check.

src/gossip/event.rs
```diff
@@ -30,8 +30,8 @@ pub(crate) struct Event<T: NetworkEvent, P: PublicId> {
     // Index of each peer's earliest event that is a descendant of this event.
     pub first_descendants: BTreeMap<P, u64>,
     // The hashes of the oldest events by each peer which comprise not-yet-stable blocks this event can see,
-    // along with the largest payload of all blocks made valid by that event.
-    pub valid_blocks_carried: BTreeMap<P, (Hash, T)>,
+    // along with the sorted payloads of all blocks made valid by that event.
+    pub valid_blocks_carried: BTreeMap<P, (Hash, Vec<T>)>,
```
maqi: if payloads are expected to be sorted, why not use BTreeSet directly?
so there is no need to call `sort` & `dedup` explicitly.

src/parsec.rs
Fraser999: But they won't all necessarily contain the same payloads I think, even if they are sorted.

src/gossip/event.rs
pierrechevalier83: fair point :D

src/parsec.rs
pierrechevalier83: OK. I'm leaning towards your intuition that there is a bug in there. Will have a closer look.

src/parsec.rs
```diff
@@ -312,86 +314,76 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-        if event.valid_blocks_carried.get(event.creator()).is_none() {
-            let valid_block_carried = {
-                let payloads_made_valid = self
+    fn set_valid_blocks_carried(&mut self, event_hash: &Hash) -> Result<(), Error> {
+        let valid_blocks_carried = {
+            let event = self.events.get(event_hash).ok_or(Error::Logic)?;
+            let mut blocks = event.valid_blocks_carried.clone();
+            // If my self_parent already carries a valid block for this peer, use it
+            if let Some(self_parent) = self.self_parent(event) {
+                blocks.append(&mut self_parent.valid_blocks_carried.clone())
```
pierrechevalier83: I don't think that's correct anymore

src/parsec.rs
```diff
@@ -312,86 +314,76 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+            }
+            // If my other_parent already carries a valid block for this peer, use it
+            if let Some(other_parent) = self.other_parent(event) {
+                blocks.append(&mut other_parent.valid_blocks_carried.clone())
```
pierrechevalier83: neither is this

src/dump_graph.rs
```diff
@@ -265,6 +275,20 @@ mod detail {
                 }
             }
             writeln!(writer, "\"]")?;
+            // Write the `valid_blocks_carried` if have
```
maqi: Line 278 - Line 291 can be removed.
As `valid_blocks_carried` already got printed at Line 249 - Line 251

src/gossip/event.rs
```diff
@@ -29,9 +29,8 @@ pub(crate) struct Event<T: NetworkEvent, P: PublicId> {
     pub last_ancestors: BTreeMap<P, u64>,
     // Index of each peer's earliest event that is a descendant of this event.
     pub first_descendants: BTreeMap<P, u64>,
-    // The hashes of the oldest events by each peer which comprise not-yet-stable blocks this event can see,
-    // along with the largest payload of all blocks made valid by that event.
-    pub valid_blocks_carried: BTreeMap<P, (Hash, T)>,
+    // All the payloads made valid by this event
```
maqi: This is more shall be: All the payloads viewed as valid by this event ? 
`made valid` sounds shall be only one event?

pierrechevalier83: maybe:
// Payloads of all the blocks made valid by that event
?
(I want to indicate that it's only the first gossip event that views that supermajority of same payloads by different authors...)

src/dump_graph.rs
pierrechevalier83: That must have been due to an incorrect conflict resolution. I will try your suggestion and see how it goes.

pierrechevalier83: OK. I addressed this (and a few more esthetical fixups in a later fixup commit)

src/gossip/event.rs
pierrechevalier83: I also pushed this in a later commit. If you disagree with the wording, we can discuss it in comments to that fixup commit :)

src/parsec.rs
```diff
@@ -312,107 +315,95 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-            }
+                .filter(|&this_payload| !self.payload_is_already_carried(event, this_payload))
+                .filter(|&this_payload| {
+                    self.peer_manager
+                        .is_super_majority(self.n_ancestors_carrying_payload(event, this_payload))
+                })
+                .cloned()
+                .collect::<BTreeSet<T>>()
+        };
+        // If this event doesn't yet have valid blocks, record this fact
```
Fraser999: Don't we record the fact regardless of the event's current state?

src/parsec.rs
```diff
@@ -312,107 +315,95 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+        if !valid_blocks_carried.is_empty() {
+            let event = self.events.get(event_hash).ok_or(Error::Logic)?;
+            let creator_id = event.creator().clone();
+            let _ = self
+                .events_with_valid_blocks
+                .entry(creator_id)
+                .and_modify(|hashes| {
+                    hashes.push_back(*event_hash);
+                })
+                .or_insert_with(|| vec![*event_hash].into_iter().collect::<VecDeque<_>>());
```
Fraser999: I think automatic type-deduction should work here, and `iter::once` is maybe more canonical?  So:
```rust
.or_insert_with(|| iter::once(*event_hash).collect());
```

src/gossip/event.rs
```diff
@@ -29,9 +29,8 @@ pub(crate) struct Event<T: NetworkEvent, P: PublicId> {
     pub last_ancestors: BTreeMap<P, u64>,
     // Index of each peer's earliest event that is a descendant of this event.
     pub first_descendants: BTreeMap<P, u64>,
-    // The hashes of the oldest events by each peer which comprise not-yet-stable blocks this event can see,
-    // along with the largest payload of all blocks made valid by that event.
-    pub valid_blocks_carried: BTreeMap<P, (Hash, T)>,
+    // Payloads of all the blocks made valid by that event
```
Fraser999: "that" → "this" for consistency?

src/parsec.rs
```diff
@@ -312,107 +315,95 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-        &self,
-        event: &Event<T, S::PublicId>,
-        payload: &Option<&T>,
-    ) -> usize {
+    fn payload_is_already_carried(&self, event: &Event<T, S::PublicId>, payload: &T) -> bool {
+        let hashes = self.events_with_valid_blocks.get(event.creator());
+        hashes.map_or(false, |hashes| {
+            hashes
+                .iter()
+                .map(|hash| {
```
Fraser999: I think `.map(|hash| ...).any()` call could just be replaced with `.any(|hash| ...)`.

src/parsec.rs
```diff
@@ -753,17 +739,23 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         // Clear all leftover data from previous consensus
         self.round_hashes = BTreeMap::new();
         self.meta_votes = BTreeMap::new();
-        self.events.iter_mut().for_each(|(_hash, event)| {
+
+        let mut events_made_empty: Vec<Hash> = vec![];
```
Fraser999: We can remove the `: Vec<Hash>` here I think.

src/parsec.rs
```diff
@@ -753,17 +739,23 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-                .clone()
-                .into_iter()
-                .filter(|(_peer, (_hash, this_payload))| payload != this_payload)
-                .collect();
-
-            event.valid_blocks_carried = new_valid_blocks;
-        });
+            let removed = event.valid_blocks_carried.remove(payload);
+            if removed && event.valid_blocks_carried.is_empty() {
+                events_made_empty.push(event.hash().clone())
```
Fraser999: Should be able to copy here rather than cloning.

src/parsec.rs
```diff
@@ -753,17 +739,23 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+            }
+        }
+        for event_hash in &events_made_empty {
+            if let Some(event) = self.events.get(event_hash) {
+                let id = event.creator();
+                if let Some(hashes) = self.events_with_valid_blocks.get_mut(id) {
+                    let _ = hashes.pop_front();
```
Fraser999: Can't this be the wrong hash we're popping here?

src/parsec.rs
```diff
@@ -312,107 +315,95 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-                        (*event.hash(), max_payload_made_valid.clone()),
-                    )
                 })
-            };
-            if let Some((peer, event_with_valid_block)) = valid_block_carried {
-                let _ = event
-                    .valid_blocks_carried
-                    .insert(peer, event_with_valid_block);
-            }
+                .filter(|&this_payload| !self.payload_is_already_carried(event, this_payload))
```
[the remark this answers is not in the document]
pierrechevalier83: They could, but should they? Sometimes, I like splitting things up to avoid too big closures... I could be convinced to change my mind though...

src/parsec.rs
pierrechevalier83: Uhm. Leftover comment :) Good catch :+1:

src/parsec.rs
pierrechevalier83: I underestimated rustc :)

src/parsec.rs
pierrechevalier83: I don't think so as they're pushed in the same order in which events are added.

pierrechevalier83:
I pushed a fixup commit for all the improvements suggested by @Fraser999 minus the last one. (Will deserve its own commit if I decide to be a bit more defensive there)

src/parsec.rs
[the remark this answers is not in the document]
Fraser999: I would personally, but don't feel strongly.  I'd guess it'd be slightly more efficient to only call `filter` once, but that'd be my main argument for making the change (and it's a weak or possibly invalid one without profiling data).

tests/utils/network.rs
```diff
@@ -42,7 +42,38 @@ impl Network {
...
+            .collect::<Vec<_>>();
+        for sender_id in &peer_ids {
+            let receiver_id = unwrap!(
+                peer_ids
+                    .iter()
+                    .filter(|&id| id != sender_id)
+                    .nth(rng.gen_range(0, peer_ids.len() - 1))
+            );
+            rng.shuffle(transactions);
+            if rng.gen_weighted_bool(10) {
```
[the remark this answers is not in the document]
pierrechevalier83: Yeah, from experience, it never runs out of iteration so I would be for keeping it as is.

src/parsec.rs
pierrechevalier83: I changed my mind (it doesn't have to be the first one), hence the latest fixup commit.

maqi: but `hashes.swap(0, ii);` will jump the 0th entry directly into ii, which means the order of events addition has been broken?
isn't that shall be something like:
```
for index in ii..hashes.len()-1 {
    hashes[index] = hashes[index+1]
}
```

src/parsec.rs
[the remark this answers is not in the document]
pierrechevalier83: OK. For tastes reason, I'll keep it as is since you seem open to the idea. :)

src/parsec.rs
pierrechevalier83: My bad. I've addressed it. It feels a bit dirty to do these repetitive swaps manually, but I haven't found rotate for VecDeque in rust...

PR #40, opened 2018-07-04
MAID-2901: stop example if consensus broken

Fraser999:


examples/basic.rs
```diff
@@ -307,12 +311,25 @@ fn main() {
...
+                peer.blocks_payloads(),
                 max_width
             );
         }
         println!();
 
+        peers.sort_by_key(|peer| usize::max_value() - peer.blocks.len());
+        let mut payloads = peers[0].blocks_payloads();
+        for peer in peers.iter().skip(1) {
+            payloads.truncate(peer.blocks.len());
```
maqi: the peer which has a short list but the same sequence won't be detected?
and will causing the later on checks only check against the shorten list?

pierrechevalier83: Good catch. I hadn't noticed this one. Probably replacing line 321 with `max_by(most paylaods)` would be a good and easy fix.

Fraser999: I'm not quite sure, but I think you're maybe both making separate points?

@maqi If a peer has fewer payloads but has them in the right order while the example is still running, that's OK.  If that's still the case after the max number of gossip rounds though, we print an error message about failing to reach consensus.

@pierre I think this is handled by sorting from most payloads to least at line 320, then taking the payloads of the first peer, and iterating through the rest, just truncating as we go.  Do you feel it's worth adding a comment describing this?

pierrechevalier83: No I'm happy with the code as is. I think I was making the same point as @maqi, but now I'm not sure anymore :)

maqi: so there is an extra check to ensure the length is `>=` the number of votes.   right?
In that case, then it will be ok.

a side point is that if the blocks is checked against `>=`, then a peer having extra block will be considered as OK (as long as the previous blocks' sequence is the same) ?

PR #43, opened 2018-07-05
MAID-2903: add test for voting at random intervals

Fraser999:


tests/utils/network.rs
```diff
@@ -42,7 +42,38 @@ impl Network {
...
+            let receiver_id = unwrap!(
+                peer_ids
+                    .iter()
+                    .filter(|&id| id != sender_id)
+                    .nth(rng.gen_range(0, peer_ids.len() - 1))
+            );
+            rng.shuffle(transactions);
+            if rng.gen_weighted_bool(10) {
+                self.peer_mut(sender_id)
+                    .vote_for_first_not_already_voted_for(&transactions);
```
maqi: what's the point to have this separate `self.peer_mut(sender_id).vote_for_first_not_already_voted_for(&transactions)` ?
the one inside `exchange_messages` calls the same

tests/utils/network.rs
```diff
@@ -75,6 +111,11 @@ impl Network {
                 .handle_request(sender_id, request)
         );
 
+        if let Some(transactns) = transactions {
+            self.peer_mut(sender_id)
+                .vote_for_first_not_already_voted_for(transactns);
```
maqi: typo of `transactns`

tests/utils/network.rs
Fraser999: It's so we have a 1:10 chance of adding a vote before we send a request, but also a 1:10 chance of adding a vote between sending a request and receiving a response.

tests/utils/network.rs
Fraser999: That was deliberate :) (to avoid hiding the outer `transactions`)

tests/utils/network.rs
```diff
@@ -42,7 +42,38 @@ impl Network {
...
+            .collect::<Vec<_>>();
+        for sender_id in &peer_ids {
+            let receiver_id = unwrap!(
+                peer_ids
+                    .iter()
+                    .filter(|&id| id != sender_id)
+                    .nth(rng.gen_range(0, peer_ids.len() - 1))
+            );
+            rng.shuffle(transactions);
+            if rng.gen_weighted_bool(10) {
```
maqi: will the chance be too risky?
in each iteration, there is 1/5 chance to send out a transaction, which means averagely requires 50 iterations to vote for all 10.
Meanwhile the max_iterations is set at just 100.

Fraser999:
Closed as code contained in this PR was added to #37.

PR #54, opened 2018-07-25
MAID-2944: don't add our own ID twice to peer_manager

Fraser999:


src/peer_manager.rs
```diff
@@ -24,15 +24,11 @@ pub(crate) struct PeerManager<S: SecretId> {
...
     /// Constructor of `PeerManager`.
     pub fn new(our_id: S) -> Self {
-        let our_public_id = our_id.public_id().clone();
-        let mut peer_manager = PeerManager {
+        PeerManager {
             our_id,
             peers: BTreeMap::new(),
             peer_id_hashes: vec![],
-        };
-        // TODO - we shouldn't just automatically add ourself once we begin work on milestone 2.
```
maqi: is this comment needs to be kept?

Fraser999: No - the code deletion is satisfying what the todo was reminding us about.

PR #59, opened 2018-08-09
replace BTreeSet<bool> with a much lighter enum

tafia:
This PR replaces `MetaVote`'s `BTreeSet<bool>` with a much lighter enum `BoolSet`.

I have tried to replicate the `BTreeSet` api so the change is as simple as possible.

### DIFF
--- a/src/meta_vote.rs
+++ b/src/meta_vote.rs
@@ -6,15 +6,9 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use std::collections::{BTreeMap, BTreeSet};
+use std::collections::BTreeMap;
 use std::fmt::{self, Debug, Formatter};
 use std::iter;
-
-fn new_set_with_value(value: bool) -> BTreeSet<bool> {
-    let mut set = BTreeSet::new();
-    let _ = set.insert(value);
-    set
-}
 
 #[derive(Clone, PartialEq, PartialOrd)]
 pub(crate) enum Step {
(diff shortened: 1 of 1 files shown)


pierrechevalier83:
Thanks for your PR, @tafia. It failed CI because of the cargo fmt check, probably because you're using a different version.

We use 0.8.2 (see our style guide: https://github.com/maidsafe/QA/blob/master/Documentation/Rust%20Style.md) which you can install with
`cargo install rustfmt-nightly -f --vers=0.8.2`

src/meta_vote.rs
```diff
@@ -40,13 +34,70 @@ impl Debug for Step {
...
+
+impl Default for BoolSet {
+    fn default() -> BoolSet {
+        BoolSet::Empty
+    }
+}
+
+impl BoolSet {
+    pub fn is_empty(&self) -> bool {
+        if let BoolSet::Empty = *self {
```
pierrechevalier83: If you make `BoolSet` `PartialEq`, this can simplify to `*self == BoolSeet::Empty`, which feels more natural

src/meta_vote.rs
```diff
@@ -40,13 +34,70 @@ impl Debug for Step {
     }
 }
 
+/// A simple enum to hold a set of bools
+#[derive(Clone)]
+pub(crate) enum BoolSet {
+    Empty,
+    True,
```
pierrechevalier83: Rather than manually handling the `true` `false`, why not have something like this:
```
pub(crate) enum BoolSet {
    Empty,
    Single(bool),
    Both,
}
```
That would make these functions read better: `insert`, `contains`, `len`, `from_bool`, `write_multiple_bool_values`

src/meta_vote.rs
tafia: Sure. I didn't want to have too many impl because it is very local, but I can do it.

pierrechevalier83: It's a sufficiently minor point that I'm happy with you following your own preference :)

src/meta_vote.rs
tafia: I thought it would make the enum larger (I've just tested: it didn't). I can do it.

tafia:
> BTW, for our defense, we have not given any thought to performance so far ;) That being said, I'm happy to incorporate your change as it's a low hanging fruit and actually improves readability overall 

It's ok :smile:! The amount of work on the whole MaidSAFE organization is really impressive. I was just randomly reading the code and `BTreeSet<bool>` looked wrong. It was a small PR with very limited impact (`pub(crate)`).

I didn't expect such results!

tafia:
Let me know if anything else needs to be done.

src/meta_vote.rs
```diff
@@ -51,24 +50,20 @@ impl Default for BoolSet {
...
         }
         true
     }
     fn contains(&self, val: &bool) -> bool {
-        match (self.clone(), *val) {
-            (BoolSet::TrueFalse, _) | (BoolSet::True, true) | (BoolSet::False, false) => true,
+        match self.clone() {
+            BoolSet::Both => true,
+            BoolSet::Single(ref s) if *s == *val => true,
             _ => false,
```
pierrechevalier83: May as well spell out that the remaining variant is BoolSet::Empty :)

tafia: Nope, there is also the case of `BoolSet::Single(s) if *s != val`

src/meta_vote.rs
```diff
@@ -51,24 +50,20 @@ impl Default for BoolSet {
...
             _ => return false,
         }
         true
     }
     fn contains(&self, val: &bool) -> bool {
-        match (self.clone(), *val) {
-            (BoolSet::TrueFalse, _) | (BoolSet::True, true) | (BoolSet::False, false) => true,
+        match self.clone() {
+            BoolSet::Both => true,
+            BoolSet::Single(ref s) if *s == *val => true,
```
pierrechevalier83: OK, now I'm properly nitpicking but isn't this marginally more readable?
`BoolSet::Single(b) => b == *val`

src/meta_vote.rs
tafia: ... I can rewrite it this way, if it make it more readable:
```rust
        match self.clone() {
            BoolSet::Empty => false,
            BoolSet::Both => true,
            BoolSet::Single(ref s) => *s == *val,
        }
```

src/meta_vote.rs
```diff
@@ -51,24 +50,20 @@ impl Default for BoolSet {
...
+        *self == BoolSet::Empty
     }
     fn insert(&mut self, val: bool) -> bool {
-        match (self.clone(), val) {
-            (BoolSet::Empty, true) => *self = BoolSet::True,
-            (BoolSet::Empty, false) => *self = BoolSet::False,
-            (BoolSet::True, false) | (BoolSet::False, true) => *self = BoolSet::TrueFalse,
+        match self.clone() {
+            BoolSet::Empty => *self = BoolSet::Single(val),
+            BoolSet::Single(ref s) if *s != val => *self = BoolSet::Both,
```
pierrechevalier83: Not sure what we gain by taking a ref to the bool and then dereferencing it?

tafia: In this case with a `clone` before, nothing indeed. Changed

tafia:
Awesome!

PR #63, opened 2018-08-20
[MAID-3107] chore/ci: update scripts to Rust stable 1.28.0

pierrechevalier83:
nightly to 2018-07-07
clippy to 0.0.212
rustfmt to 0.99.2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ env:
   global:
     - RUST_BACKTRACE=1
     - PATH=$PATH:$HOME/.cargo/bin
+    - RUST_RUSTFMT=0.99.2
+    - RUST_CLIPPY=0.0.212
 os:
   - linux
   - osx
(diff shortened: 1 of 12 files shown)


cargo_install.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
m-cat: Why are we duplicating this file when we have it in the QA repo?

.travis.yml
```diff
@@ -19,8 +19,8 @@ before_script:
   - curl -sSL https://github.com/maidsafe/QA/raw/master/travis/cargo_install.sh > cargo_install.sh
   - bash cargo_install.sh cargo-prune;
   - if [[ $TRAVIS_RUST_VERSION =~ ^nightly.* && $TRAVIS_OS_NAME == linux ]]; then
-      bash cargo_install.sh rustfmt-nightly 0.8.2;
```
m-cat: Can we move these version strings into variables at the top of the script, like in our other CI scripts?

m-cat: Can we move these version strings into variables at the top of the script, like in our other CI scripts?

cargo_install.sh
pierrechevalier83: Oh oh! pushed by mistake! I curled it to try and reproduce an issue on CI... Good catch!

tests/utils/schedule.rs
```diff
@@ -408,6 +408,7 @@ impl Schedule {
     }
 
     /// Creates a new pseudo-random schedule based on the given options
+    #[allow(unknown_lints, let_and_return)]
```
Fraser999: Rather than allowing unknown lints, shouldn't we just enable the clippy lint when building with clippy?
```rust
#[cfg_attr(feature = "cargo-clippy", allow(let_and_return))]
```

.travis.yml
pierrechevalier83: Done.

PR #65, opened 2018-08-22
MAID-3103 & MAID-3105: coin toss correction

maqi:
When checking whether we shall stop waiting to toss a coin, check
along the events created by the event's creator.

### DIFF
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -517,7 +517,7 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         }
 
         // If we've already waited long enough, get the aux value of the highest ranking leader.
-        if self.stop_waiting(round) {
+        if self.stop_waiting(round, event) {
             for (_, creator) in &peer_id_hashes[1..] {
                 if let Some(creator_event_index) = event.last_ancestors.get(creator) {
                     if let Some(aux_value) =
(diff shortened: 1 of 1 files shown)


src/parsec.rs
```diff
@@ -551,10 +551,11 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
 
-    // Skips back through our events until we've passed `responsiveness_threshold` response events
-    // and sees if we had our own `aux_value` set at this round and step.  If so, returns `true`.
-    fn stop_waiting(&self, round: usize) -> bool {
-        let mut event_hash = self.peer_manager.last_event_hash(self.our_pub_id());
+    // Skips back through events created by the peer until passed `responsiveness_threshold`
+    // response events and sees if the peer had its `aux_value` set at this round and step.  If so,
+    // returns `true`.
+    fn stop_waiting(&self, round: usize, event: &Event<T, S::PublicId>) -> bool {
+        let mut event_hash = event.self_parent();
```
Fraser999: This gives us an off-by-one error.  We need to also include `event` itself in the count if it is a response too:
```rust
let mut event_hash = Some(event.hash());
```
or if we're sure `responsiveness_threshold` will always be > 1
```rust
let mut event_hash = event.self_parent();
let mut response_count = if event.is_response() { 1 } else { 0 };
```
(I'd prefer the first even though it might be slightly less efficient - or maybe there's a way to get the best of both?)

src/parsec.rs
```diff
@@ -551,10 +551,11 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
 
-    // Skips back through our events until we've passed `responsiveness_threshold` response events
-    // and sees if we had our own `aux_value` set at this round and step.  If so, returns `true`.
-    fn stop_waiting(&self, round: usize) -> bool {
-        let mut event_hash = self.peer_manager.last_event_hash(self.our_pub_id());
+    // Skips back through events created by the peer until passed `responsiveness_threshold`
+    // response events and sees if the peer had its `aux_value` set at this round.  If so, returns
+    // `true`.
+    fn stop_waiting(&self, round: usize, event: &Event<T, S::PublicId>) -> bool {
+        let mut event_hash = Some(event.hash());
```
pierrechevalier83: If we're going to take the event, we know it always exists, so let's avoid wrapping it in an Option and then have to `and_then` ;)

maqi: that `event_hash` will be recursively fetched from `event.self_parent` which is an `Option`
so here the wrap is just to ensure the variable is in the same format.

PR #67, opened 2018-08-23
MAID-3042 chore/parsec: make initial event more type safer

maqi:


src/gossip/cause.rs
```diff
@@ -14,12 +14,12 @@ use vote::Vote;
...
-    Request(Hash),
-    // Hash is of the latest `Event` of the peer which sent the response.
-    Response(Hash),
-    // Vote for a single network event of type `T`.
-    Observation(Vote<T, P>),
+    // Hashes are the latest `Event` of own and the peer which sent the request (self, other).
+    Request(Hash, Hash),
+    // Hashes are the latest `Event` of own and the peer which sent the request (self, other).
+    Response(Hash, Hash),
+    // Hash is the latest `Event` of own. Vote for a single network event of type `T`.
```
pierrechevalier83: // Hash of our latest `Event`.
would sound better (?)

src/gossip/cause.rs
```diff
@@ -14,12 +14,12 @@ use vote::Vote;
...
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub(super) enum Cause<T: NetworkEvent, P: PublicId> {
-    // Hash is of the latest `Event` of the peer which sent the request.
-    Request(Hash),
-    // Hash is of the latest `Event` of the peer which sent the response.
-    Response(Hash),
-    // Vote for a single network event of type `T`.
-    Observation(Vote<T, P>),
+    // Hashes are the latest `Event` of own and the peer which sent the request (self, other).
+    Request(Hash, Hash),
```
pierrechevalier83: I think it would be worth creating a little struct:
```
struct Parents {
    self_parent: Hash,
    other_parent: Hash,
}
```

Just so we can have the type system help us figure out which one is which rather than relying on comments.

pierrechevalier83: Potentially, the other_parent could be in an Option so that the Observation can also use this type. I'm a bit undecided on that, so your call :)

pierrechevalier83: Good idea! Same principle, but more elegant that my proposal :smile:

src/gossip/cause.rs
```diff
@@ -11,15 +11,18 @@ use id::PublicId;
 use network_event::NetworkEvent;
 use vote::Vote;
 
+type OtherParentHash = Hash;
```
Fraser999: I disagree that this is a good way to handle the issue.  Those two aliases for `Hash` are only visible inside that file, and it makes it just as easy for the caller to swap the two hashes accidentally.

I'd prefer something more like:
```rust
Request {
    self_parent: Hash,
    other_parent: Hash,
},
Response {
    self_parent: Hash,
    other_parent: Hash,
},
Observation {
    self_parent: Hash,
    vote: Vote<T, P>,
},
Initial,
```
This would mean that calling code would be a bit safer, e.g.
```rust
Cause::Request(_, other_hash) | Cause::Response(_, other_hash) => Some(other_hash),
```
would become
```rust
Cause::Request { other_parent, .. } | Cause::Response { other_parent, .. } => Some(other_parent)
```
In the former version, it would be easy to mistakenly write `Cause::Request(other_hash, _)` which would compile fine, but would likely be a bug since the variable `other_hash` would actually have the value of `self_parent`.

src/gossip/cause.rs
```diff
@@ -14,12 +14,21 @@ use vote::Vote;
...
-    // Hash is of the latest `Event` of the peer which sent the response.
-    Response(Hash),
-    // Vote for a single network event of type `T`.
-    Observation(Vote<T, P>),
+    // Hashes are the latest `Event` of own and the peer which sent the request.
+    Request {
+        self_parent: Hash,
+        other_parent: Hash,
+    },
+    // Hashes are the latest `Event` of own and the peer which sent the request.
```
Fraser999: This should be `response` I guess.  I think with this new format though, we can safely remove comments from all variants.

PR #69, opened 2018-08-27
MAID-2965: update dependencies

Fraser999:
Specifically this is to make the new log_or_panic macro in maidsafe_utilities available to Parsec.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,13 +28,13 @@ script:
   - if [[ ! $TRAVIS_RUST_VERSION =~ ^nightly.* ]]; then
       (
         set -x;
-        cargo test --verbose --release
+        cargo test --verbose --release --features=testing
       );
     elif [[ $TRAVIS_OS_NAME == linux ]]; then
       (
         set -x;
         cargo fmt -- --check &&
-        cargo clippy --verbose --all-targets --features dump-graphs
+        cargo clippy --verbose --all-targets --features=dump-graphs
       );
     fi
 before_cache:
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,10 @@ script:
       (
         set -x;
         cargo fmt -- --check &&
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -10,18 +10,20 @@ repository = "https://github.com/maidsafe/parsec"
 version = "0.5.0"
 
 [dependencies]
-lazy_static = { version = "~1.0.1", optional = true }
```
pierrechevalier83: Do all of these dependencies need to be updated to keep in sync with maidsafe_utilities. It's OK if they are; but if they aren't, let's maybe keep updates to the minimum in this commit.

.travis.yml
```diff
@@ -28,13 +28,13 @@ script:
...
         set -x;
-        cargo test --verbose --release
+        cargo test --verbose --release --features=testing
       );
     elif [[ $TRAVIS_OS_NAME == linux ]]; then
       (
         set -x;
         cargo fmt -- --check &&
-        cargo clippy --verbose --all-targets --features dump-graphs
+        cargo clippy --verbose --all-targets --features=dump-graphs
```
maqi: do we now need to execute something like:
`cargo clippy --verbose --all-targets && cargo clippy --verbose --all-targets --features=dump-graphs,testing` ?

Cargo.toml
Fraser999: Couldn't help myself since I was updating _some_ :)  I'll revert the unneeded changes.

.travis.yml
Fraser999: I toyed with adding that, but since we only use `testing` to pass it down to utils (i.e. there's no extra or different code getting compiled in Parsec when using this feature) I don't think it's worthwhile.

Possibly it should be added now for completeness and future-proofing?

maqi: Yeah, I agree adding it in this PR is mainly for future-proofing.
We could add it in the work of task 2966

PR #70, opened 2018-08-27
MAID-3009 construct Event with gossip graph

maqi:


src/peer_list.rs
```diff
@@ -94,14 +94,10 @@ impl<S: SecretId> PeerList<S> {
         &mut self,
         event: &Event<T, S::PublicId>,
     ) -> Result<(), Error> {
-        let (peer, index) = if let Some(index) = event.index {
```
Fraser999: This could be simplified a bit to e.g.
```rust
if let Some(peer) = self.peers.get_mut(event.creator()) {
    match peer.entry(event.index) {
        Entry::Occupied(entry) => {
            if entry.get() != event.hash() {
                return Err(Error::InvalidEvent);
            }
        }
        Entry::Vacant(entry) => {
            let _ = entry.insert(*event.hash());
        }
    }
    Ok(())
} else {
    Err(Error::UnknownPeer)
}
```

src/parsec.rs
```diff
@@ -769,12 +689,12 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         x: &Event<T, S::PublicId>,
         y: &Event<T, S::PublicId>,
     ) -> usize {
-        y.first_descendants
+        y.last_ancestors
```
Fraser999: I don't think this is correct, and I'm wondering if removing `first_descendants` was actually such a good idea!

I think a fix would be something like:
```
for each of x.last_ancestors {
    if y.can_see(that_ancestor) then ++total
}
```

It also seems like the way this works, event `x` would have to come later than `y` to return a non-zero value.  Maybe we should update the comment to be explicit about the direction e.g. "from event X to event Y" and make `x` need to be the earlier event?  And if so, then in my pseudo-code above, we'd need to swap `x` and `y`.

src/gossip/event.rs
```diff
@@ -95,15 +117,16 @@ impl<T: NetworkEvent, P: PublicId> Event<T, P> {
...
-        // All fields except `content`, `signature` and `hash` still need to be set correctly by the
-        // caller.
+        let (index, last_ancestors) =
+            Self::index_and_last_ancestors(&packed_event.content, events, peer_list)?;
+
+        // `valid_blocks_carried` and `observations` still need to be set correctly by the caller.
         Ok(Self {
-            content: packed_event.content,
-            signature: packed_event.signature,
+            content: packed_event.content.clone(),
```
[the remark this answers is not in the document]
maqi: I thught cloning content/signature will be less work than pass request/response by value?

src/parsec.rs
maqi: I agree the comment shall be more explicit by about the direction.

src/gossip/event.rs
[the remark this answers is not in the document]
Fraser999: We'd be consuming the request/response, so I expect it'd get moved and that should be cheaper than cloning the contents here.

src/parsec.rs
```diff
@@ -763,19 +681,20 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-            .filter(|(peer_id, descendant)| {
-                x.last_ancestors
-                    .get(&peer_id)
-                    .map(|last_ancestor| last_ancestor >= *descendant)
-                    .unwrap_or(false)
+            .filter(|(peer_id, ancestor)| {
+                self.peer_list
+                    .event_by_index(peer_id, **ancestor)
+                    .and_then(|hash| self.events.get(hash))
+                    .and_then(|event| event.last_ancestors.get(x.creator()))
```
Fraser999: This has reversed `x` and `y` from before the PR.  We either need to swap the variables when calling this function, or preferably change them in here and update the comment.

src/gossip/event.rs
```diff
@@ -24,11 +25,9 @@ pub(crate) struct Event<T: NetworkEvent, P: PublicId> {
     signature: P::Signature,
     hash: Hash,
     // Sequential index of this event: this event is the `index`-th one made by its creator.
-    pub index: Option<u64>,
+    pub index: u64,
```
Fraser999: We can make `index` and `last_ancestors` private now and accessible via public getters.

src/parsec.rs
```diff
@@ -764,19 +724,20 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
     }
 
     // Returns the number of peers through which there is a directed path in the gossip graph
-    // between event X and event Y.
+    // from event X (ancestor) to event Y (descendant).
```
pierrechevalier83: Could we actually rename `x` to `ancestor` and `y` to `descendant` to reduce the cognitive overhead?

src/gossip/event.rs
```diff
@@ -155,24 +175,65 @@ impl<T: NetworkEvent, P: PublicId> Event<T, P> {
...
+        if let Some(self_parent) = content.self_parent().and_then(|hash| events.get(hash)) {
+            last_ancestors = self_parent.last_ancestors().clone();
+
+            if let Some(other_parent) = content.other_parent().and_then(|hash| events.get(hash)) {
+                for (peer_id, _) in peer_list.iter() {
+                    if let Some(other_index) = other_parent.last_ancestors().get(peer_id) {
+                        let existing_index = last_ancestors
+                            .entry(peer_id.clone())
+                            .or_insert(*other_index);
+                        if *existing_index < *other_index {
```
pierrechevalier83: Style: these 2 lines could be better expressed with `cmp::max`

src/parsec.rs
```diff
@@ -802,19 +721,20 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
             .iter()
-            .filter(|(peer_id, descendant)| {
-                x.last_ancestors
-                    .get(&peer_id)
-                    .map(|last_ancestor| last_ancestor >= *descendant)
-                    .unwrap_or(false)
+            .filter(|(peer_id, ancestor)| {
+                self.peer_list
+                    .event_by_index(peer_id, **ancestor)
+                    .and_then(|hash| self.events.get(hash))
```
Fraser999: I think we should use `self.get_known_event(hash).ok()` here to get the benefit of `log_or_panic!`

maqi: As discussed, this will be covered in a separate PR, which I will raise soon.

PR #71, opened 2018-08-28
MAID-2966: use log_or_panic macro

Fraser999:


src/parsec.rs
```diff
@@ -792,30 +830,36 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-        let other_parent = {
-            let other_parent_hash = self
-                .peer_list
-                .last_event_hash(peer_id)
-                .ok_or(Error::Logic)?;
-            self.events.get(other_parent_hash).ok_or(Error::Logic)?
+        let peer_last_event = if let Some(event_hash) = self.peer_list.last_event_hash(peer_id) {
+            self.get_known_event(event_hash)?
+        } else {
+            log_or_panic!("{:?} doesn't have peer {:?}", self.our_pub_id(), peer_id);
```
maqi: it could also be that there is no event for that peer recorded in the peer_list ?

src/parsec.rs
```diff
@@ -150,12 +156,41 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         })
     }
 
+    /// Must only be used for events which have already been added to our graph.
+    fn get_known_event(&self, event_hash: &Hash) -> Result<&Event<T, S::PublicId>, Error> {
```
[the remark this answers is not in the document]
Fraser999: No - I think the behaviour in calling code is the same.  For example, in some places we were using `?` on the `Option` and now we're using it on the `Result`.

src/parsec.rs
```diff
@@ -103,11 +109,11 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
                     .map(Request::new);
             }
         }
-        Ok(Request::new(
-            self.events_order
-                .iter()
-                .filter_map(|hash| self.events.get(hash)),
-        ))
+        let mut events = vec![];
```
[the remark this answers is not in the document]
Fraser999: I don't think we should, since we'd be returning a (probably) invalid `Request` as `Ok` - requests and responses shouldn't have missing events.

src/parsec.rs
Fraser999: This is maintaining the current behaviour - we actually ensure `self.peer_list.last_event_hash()` can't fail before calling this function.  I'll update the comment to try and make that bit clearer.

src/parsec.rs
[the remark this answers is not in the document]
maqi: >the behaviour in calling code is the same

src/parsec.rs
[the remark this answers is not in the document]
maqi: Yeah, I agree.
It's just it is the previoius behaviour, and I'd just to confirm the change is expected.

src/parsec.rs
```diff
@@ -103,11 +109,11 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+        for event_hash in &self.events_order {
+            events.push(self.get_known_event(event_hash)?);
```
pierrechevalier83: Nitpick/style:
Was there a reason for not directly mapping self.events_order to events?
Borrow checker?

src/parsec.rs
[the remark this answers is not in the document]
pierrechevalier83: I think it's a great idea to encapsulate that logic. It was duplicated everywhere prior to this change :+1:

src/parsec.rs
pierrechevalier83: Maybe it should be done in a different PR, but I agree with Qi: the existing behaviour isn't really nice as the caller is responsible of doing the special handling for the first time we gossip to a peer. I think it's this function's responsability and should be handled here.
If the peer is not in our peer-list, I agree that it's ok to just err.

src/parsec.rs
Fraser999: No, it's just that we can't bail out of a loop of `iter().filter_map(fn)` by using `?` in the fn.

src/parsec.rs
Fraser999: I do see the point your both making, and I agree to an extent.  My reasons for preferring keeping the existing behaviour are:
* it's a private method, currently only called from two places
* one of those call sites is `handle_request` which I expect to be called fairly frequently, and which already naturally ensures the preconditions for calling `events_to_gossip_to_peer()` are met.  It'll be wasteful to check the preconditions inside `events_to_gossip_to_peer()` when it's called from here.
* because we're using `log_or_panic` now and implementing unit tests, if we break the preconditions in a future change, I'm pretty confident it'll get caught in tests

Having said all that, I'm not strongly against changing to have the extra checks inside here - it's mainly a performance and code style argument ultimately I suppose.

pierrechevalier83: Yeah, it's a fair agreement. I agree it comes down to style and since it's private, I'm not too worried that the callers will proliferate and make the overhead unbearable. Happy to keep as is.

src/parsec.rs
pierrechevalier83: Ah! Of course! Yes, that's a slight ergonomic annoyance with error handling...

PR #72, opened 2018-08-30
MAID-3010: Test Event construction

octol:
The second half of MAID-3010. The first half is to be done later.

### DIFF
--- a/src/gossip/event.rs
+++ b/src/gossip/event.rs
@@ -20,6 +20,7 @@ use std::collections::{BTreeMap, BTreeSet};
 use std::fmt::{self, Debug, Formatter};
 use vote::Vote;
 
+#[derive(PartialEq)]
 pub(crate) struct Event<T: NetworkEvent, P: PublicId> {
     content: Content<T, P>,
     // Creator's signature of `content`.
@@ -161,7 +162,7 @@ impl<T: NetworkEvent, P: PublicId> Event<T, P> {
         self.content.self_parent()
     }
 
-    #[cfg(feature = "dump-graphs")]
+    #[cfg(any(test, feature = "dump-graphs"))]
     pub fn other_parent(&self) -> Option<&Hash> {
         self.content.other_parent()
     }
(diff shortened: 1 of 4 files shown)


src/gossip/event.rs
```diff
@@ -203,3 +204,100 @@ impl<T: NetworkEvent, P: PublicId> Debug for Event<T, P> {
...
+        assert!(event_from_observation.other_parent().is_none());
+    }
+
+    #[test]
+    fn event_construction_unpack() {
+        let peer_id = PeerId::new("initial_peer_id");
+        let initial = Event::<Transaction, PeerId>::new_initial(&peer_id);
+
+        let other_peer_id = PeerId::new("other_peer_id");
+        let other_initial = Event::<Transaction, PeerId>::new_initial(&other_peer_id);
```
octol: Quite a lot of duplicated data setup, should we use some form of test fixture?

pierrechevalier83: I don't think the concept of "test fixture" from C++ unit test framework has a direct mapping in rust, but I suspect that a couple of functions and structs would do the job of fixing your duplication issue if you felt so inclined. I'll let the other reviewer suggest some more fancy approach that I'm not familiar with :smile:

src/gossip/event.rs
```diff
@@ -203,3 +204,100 @@ impl<T: NetworkEvent, P: PublicId> Debug for Event<T, P> {
...
+        assert!(event_from_request.index == None);
+
+        assert!(event_from_request.self_parent().unwrap() == self_parent);
+        assert!(event_from_request.other_parent().unwrap() == other_parent);
+    }
+
+    #[test]
+    fn event_construction_from_response() {
+        let peer_id = PeerId::new("initial_peer_id");
+        let initial = Event::<Transaction, PeerId>::new_initial(&peer_id);
```
pierrechevalier83: unneeded

src/gossip/event.rs
```diff
@@ -203,3 +204,100 @@ impl<T: NetworkEvent, P: PublicId> Debug for Event<T, P> {
...
+
+        let self_parent = initial.hash();
+        let net_event = Transaction::new("transaction");
+
+        let event_from_observation =
+            Event::<Transaction, PeerId>::new_from_observation(&peer_id, *self_parent, net_event);
+
+        assert!(!event_from_observation.is_initial());
+        assert!(!event_from_observation.is_response());
+        assert!(initial.index == None);
```
pierrechevalier83: From memory, this will conflict with PR 70. Sorry you had to do duplicated work. Could you rebase and test the new behaviour?

src/gossip/event.rs
```diff
@@ -203,3 +204,100 @@ impl<T: NetworkEvent, P: PublicId> Debug for Event<T, P> {
...
+        let initial = Event::<Transaction, PeerId>::new_initial(&peer_id);
+
+        let other_peer_id = PeerId::new("other_peer_id");
+        let other_initial = Event::<Transaction, PeerId>::new_initial(&other_peer_id);
+
+        let self_parent = initial.hash();
+        let other_parent = other_initial.hash();
+
+        let event_from_request =
+            Event::<Transaction, PeerId>::new_from_request(&peer_id, *self_parent, *other_parent);
```
maqi: These tests are testing the previous constructors, not the new ones as specified in task 3008 ?

octol: Yep I'm updating them :)

src/gossip/event.rs
```diff
@@ -260,3 +261,116 @@ impl<T: NetworkEvent, P: PublicId> Debug for Event<T, P> {
         )
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use gossip::Event;
+    use mock::PeerId;
```
Fraser999: These two can be clubbed together like `use mock::{PeerId, Transaction};`

src/gossip/event.rs
```diff
@@ -260,3 +261,116 @@ impl<T: NetworkEvent, P: PublicId> Debug for Event<T, P> {
...
+        let event_from_request = Event::<Transaction, PeerId>::new_from_request(
+            *initial.hash(),
+            *other_initial.hash(),
+            &events,
+            &another_peer_list,
+        );
+
+        assert!(!event_from_request.is_initial());
+        assert!(!event_from_request.is_response());
+        assert!(event_from_request.self_parent().unwrap() == initial.hash());
```
Fraser999: We prefer the `unwrap!` macro because it gives a useful backtrace.  You'll need to add something like the following to lib.rs:
```rust
#[cfg(test)]
#[macro_use]
extern crate unwrap;
```
since it's currently only used by the example and not imported into the main lib.

Also, there's the `assert_eq!` macro which is usually more suitable than `assert!(... == ...)` since it prints the two objects being compared in the case of failure.  The main cases I can think of where we _would_ prefer `assert!` is if one of the types doesn't implement `Debug` (hence can't be printed - so you'd get a compiler error anyway) or if you're comparing something which would create a huge amount of output if it failed (e.g. two 1MB chunks of binary data).

src/gossip/event.rs
```diff
@@ -260,3 +261,174 @@ impl<T: NetworkEvent, P: PublicId> Debug for Event<T, P> {
...
+            &alice_peer_list,
+        );
+
+        assert_eq!(event_from_observation.content.creator, PeerId::new("Alice"));
+        match &event_from_observation.content.cause {
+            Cause::Observation { self_parent, vote } => {
+                assert_eq!(self_parent, &initial_event_hash);
+                assert_eq!(vote.payload(), &Transaction::new("event_observed_by_alice"));
+            }
+            _ => panic!("Shouldn't happen"),
```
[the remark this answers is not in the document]
octol: Yes good point!

src/gossip/event.rs
```diff
@@ -260,3 +261,174 @@ impl<T: NetworkEvent, P: PublicId> Debug for Event<T, P> {
...
+        assert!(initial.is_initial());
+        assert!(!initial.is_response());
+        assert!(initial.self_parent().is_none());
+        assert!(initial.other_parent().is_none());
+        assert_eq!(initial.index, 0);
+    }
+
+    #[test]
+    fn event_construction_from_observation() {
+        let alice = PeerId::new("Alice");
```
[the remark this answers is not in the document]
octol: Yep sounds very sensible. As soon it's all confirmed to be logically sound I'll try to refactor out the repeated lines

src/gossip/event.rs
```diff
@@ -260,3 +261,174 @@ impl<T: NetworkEvent, P: PublicId> Debug for Event<T, P> {
...
+                assert_eq!(vote.payload(), &Transaction::new("event_observed_by_alice"));
+            }
+            _ => panic!("Shouldn't happen"),
+        }
+        assert_eq!(event_from_observation.index, 1);
+        assert!(!event_from_observation.is_initial());
+        assert!(!event_from_observation.is_response());
+        assert_eq!(
+            unwrap!(event_from_observation.self_parent()),
+            &initial_event_hash
```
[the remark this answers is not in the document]
octol: Ah I didn't know that, that indeed sounds much more elegant!

src/gossip/event.rs
```diff
@@ -375,25 +384,60 @@ mod tests {
...
-        let bob_initial = Event::<Transaction, PeerId>::new_initial(&bob_list);
-
-        // Create Gossip graph
+    struct PeerListAndEvent {
+        peer_list: PeerList<PeerId>,
+        event: Event<Transaction, PeerId>,
+    }
+
+    fn create_two_events(id0: &str, id1: &str) -> (PeerListAndEvent, PeerListAndEvent) {
+        let peer_id0 = PeerId::new(id0);
```
pierrechevalier83: You're still duplicating `create_event_with_single_peer` here. Couldn't this simply be called twice from here? (I may be missing something)

src/gossip/event.rs
```diff
@@ -412,23 +456,18 @@ mod tests {
...
-        let alice_initial = Event::<Transaction, PeerId>::new_initial(&alice_list);
-        let bob_initial = Event::<Transaction, PeerId>::new_initial(&bob_list);
-
-        // Create Gossip graph
-        let mut events = BTreeMap::new();
-        let alice_initial_hash = *alice_initial.hash();
-        let bob_initial_hash = *bob_initial.hash();
-        let _ = events.insert(alice_initial_hash, alice_initial);
-        let _ = events.insert(bob_initial_hash, bob_initial);
+        let (
```
pierrechevalier83: see previous comment

src/gossip/event.rs
```diff
@@ -375,25 +384,60 @@ mod tests {
...
+        let mut peer_id0_list = PeerList::<PeerId>::new(peer_id0.clone());
+        let mut peer_id1_list = PeerList::<PeerId>::new(peer_id1.clone());
+        peer_id0_list.add_peer(peer_id1);
+        peer_id1_list.add_peer(peer_id0);
+
+        let initial0 = Event::<Transaction, PeerId>::new_initial(&peer_id0_list);
+        let initial1 = Event::<Transaction, PeerId>::new_initial(&peer_id1_list);
+
+        (
+            PeerListAndEvent {
```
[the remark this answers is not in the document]
octol: Good idea, I've tried to do something like the above suggestion

src/gossip/event.rs
octol: Well the difference is that PeerLists are mutated before the new Events are created, so it wouldn't be a straight call to `create_event_with_single_peer`.' I agree it would be nice to merge them

src/gossip/event.rs
```diff
@@ -375,25 +384,60 @@ mod tests {
...
         let alice_initial_hash = *alice_initial.hash();
         let bob_initial_hash = *bob_initial.hash();
         let _ = events.insert(alice_initial_hash, alice_initial);
         let _ = events.insert(bob_initial_hash, bob_initial);
+        (alice_initial_hash, bob_initial_hash, events)
+    }
+
+    #[test]
+    fn event_construction_from_request() {
+        let (
```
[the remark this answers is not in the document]
octol: I agree!

src/gossip/event.rs
octol: Well I had a go at deduplicating

src/gossip/event.rs
octol: I'll see what I can do!

src/gossip/event.rs
```diff
@@ -344,29 +357,49 @@ mod tests {
...
+    }
+
     #[test]
     fn event_construction_initial() {
-        let (_, initial) = create_event_with_single_peer("Alice");
-        assert!(initial.is_initial());
-        assert!(!initial.is_response());
-        assert!(initial.self_parent().is_none());
-        assert_eq!(initial.index, 0);
+        let alice = create_event_with_single_peer("Alice");
```
pierrechevalier83: this could be:
```
let initial = create_event_with_single_peer("Alice").event;
```
So you don't have to repeat `alice.event` 4 times.

src/gossip/event.rs
```diff
@@ -312,3 +313,178 @@ impl<T: NetworkEvent, P: PublicId> Debug for Event<T, P> {
...
+        let (_, peer_list) = create_peer_list(id);
+        PeerListAndEvent::new(peer_list)
+    }
+
+    fn insert_into_gossip_graph(
+        initial_event: Event<Transaction, PeerId>,
+        mut events: BTreeMap<Hash, Event<Transaction, PeerId>>,
+    ) -> (Hash, BTreeMap<Hash, Event<Transaction, PeerId>>) {
+        let initial_event_hash = *initial_event.hash();
+        let _ = events.insert(initial_event_hash, initial_event);
```
Fraser999: I think we should `assert!(insert().is_none())` here since it'd probably be an error to insert a duplicate event.

src/gossip/event.rs
```diff
@@ -312,3 +313,178 @@ impl<T: NetworkEvent, P: PublicId> Debug for Event<T, P> {
...
+    }
+
+    fn create_event_with_single_peer(id: &str) -> PeerListAndEvent {
+        let (_, peer_list) = create_peer_list(id);
+        PeerListAndEvent::new(peer_list)
+    }
+
+    fn insert_into_gossip_graph(
+        initial_event: Event<Transaction, PeerId>,
+        mut events: BTreeMap<Hash, Event<Transaction, PeerId>>,
```
Fraser999: I don't have a problem with this approach of moving in `events`, mutating it inside and then returning it.  However, I'd have gone for just passing `events` by mut ref.

The downsides with my approach are that you'd have to always construct the map before calling this function, and also `events` would be mutable.  The downside with your approach is that it lends itself to shadowing (see lines 375-377 where we have three different actual `events` variables) and this is probably more confusing than just passing `&mut events` to the function.

I'm not suggesting you change this unless you want to - just an observation I made!

src/gossip/event.rs
```diff
@@ -312,3 +313,178 @@ impl<T: NetworkEvent, P: PublicId> Debug for Event<T, P> {
...
+        let (bob_initial_hash, events) = insert_into_gossip_graph(bob_initial, events);
+        (alice_initial_hash, bob_initial_hash, events)
+    }
+
+    #[test]
+    fn event_construction_initial() {
+        let initial = create_event_with_single_peer("Alice").event;
+        assert!(initial.is_initial());
+        assert!(!initial.is_response());
+        assert!(initial.self_parent().is_none());
```
Fraser999: In all these tests where we check `self_parent()`, I think we should also check `other_parent()`.

You can either call `event.content.other_parent()`, or you can expose the existing `Event::other_parent()` by changing the attached configurations to include `test` - i.e.
```rust
#[cfg(any(test, feature = "dump-graphs"))]
pub fn other_parent(&self) -> Option<&Hash> {
...
```

src/gossip/event.rs
```diff
@@ -312,3 +313,178 @@ impl<T: NetworkEvent, P: PublicId> Debug for Event<T, P> {
...
+        let net_event = Transaction::new("event_observed_by_alice");
+
+        let event_from_observation = Event::<Transaction, PeerId>::new_from_observation(
+            initial_event_hash,
+            net_event,
+            &events,
+            &alice.peer_list,
+        );
+
+        assert_eq!(event_from_observation.content.creator, PeerId::new("Alice"));
```
Fraser999: Rather than writing "Alice" twice, it'd be more robust to use `alice.peer_list.our_id().public_id()`.

Same goes for line 409 (we should compare `vote.payload()` to `net_event`, which means cloning `net_event` at line 400) and similarly in the other tests.

src/gossip/event.rs
octol: I think it's a good argument, I'll go ahead and make the change!

src/gossip/event.rs
octol: Yes agree. That check was in the first version, but I removed it when `feature = "dump-graphs"` was added to `other_parent`. I'll just ahead and add the `test` cfg to that function then

src/gossip/event.rs
octol: Agree!

src/gossip/event.rs
octol: Yep I agree

src/gossip/event.rs
```diff
@@ -312,3 +313,242 @@ impl<T: NetworkEvent, P: PublicId> Debug for Event<T, P> {
...
+        );
+        assert_eq!(event_from_request.index, 1);
+        assert!(!event_from_request.is_initial());
+        assert!(!event_from_request.is_response());
+        assert_eq!(event_from_request.self_parent(), Some(&alice_initial_hash));
+        assert_eq!(event_from_request.other_parent(), Some(&bob_initial_hash));
+    }
+
+    #[test]
+    #[should_panic]
```
maqi: better use `[should_panic(expected = "Alice constructed an invalid event.")` here to give a clue of expected result?
same advice to the other panic tests.

octol: Ah sounds sensible, willdo!

src/gossip/event.rs
```diff
@@ -312,3 +313,178 @@ impl<T: NetworkEvent, P: PublicId> Debug for Event<T, P> {
...
+
+        let event_from_observation = Event::<Transaction, PeerId>::new_from_observation(
+            initial_event_hash,
+            net_event,
+            &events,
+            &alice.peer_list,
+        );
+
+        let packed_event = event_from_observation.pack();
+        let unpacked_event = unwrap!(unwrap!(Event::<Transaction, PeerId>::unpack(
```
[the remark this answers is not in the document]
pierrechevalier83: Just for the record, we prefer `unwrap!` over `.unwrap()` since it adds useful output.

PR #73, opened 2018-08-30
chore/event: move some checks from parsec to event

Fraser999:


src/gossip/event.rs
```diff
@@ -187,36 +197,55 @@ impl<T: NetworkEvent, P: PublicId> Event<T, P> {
...
             observations: BTreeSet::default(),
+        };
+
+        if let Ok((index, last_ancestors)) =
+            Self::index_and_last_ancestors(&event.content, events, peer_list)
+        {
+            event.index = index;
+            event.last_ancestors = last_ancestors;
+        } else {
+            log_or_panic!(
```
maqi: shall we print the error returned from `index_and_last_ancestor` ?

src/gossip/event.rs
```diff
@@ -187,36 +197,55 @@ impl<T: NetworkEvent, P: PublicId> Event<T, P> {
         };
         let serialised_content = serialise(&content);
 
-        let (index, last_ancestors) = Self::index_and_last_ancestors(&content, events, peer_list);
-
-        // `valid_blocks_carried` and `observations` still need to be set correctly by the caller.
-        Self {
+        let mut event = Self {
```
maqi: any particular reason changed to have a `default mut event` first then update the index & last_ancestor fields?

src/gossip/event.rs
```diff
@@ -187,36 +197,55 @@ impl<T: NetworkEvent, P: PublicId> Event<T, P> {
...
         peer_list: &PeerList<S>,
-    ) -> (u64, BTreeMap<P, u64>) {
-        // An initial event, which doesn't have self_parent, will have an index of 0.
-        let index = content
-            .self_parent()
-            .and_then(|hash| events.get(hash))
-            .map_or(0, |parent| parent.index + 1);
+    ) -> Result<(u64, BTreeMap<P, u64>), Error> {
+        let self_parent = if let Some(self_parent_hash) = content.self_parent() {
+            if let Some(event) = events.get(&self_parent_hash) {
```
pierrechevalier83: Could this be more concise with
`let event = events.get(&self_parent_hash).ok_or(Error::UnknownParent)?`
?

src/gossip/event.rs
```diff
@@ -187,36 +197,55 @@ impl<T: NetworkEvent, P: PublicId> Event<T, P> {
...
+                event
+            } else {
+                return Err(Error::UnknownParent);
+            }
+        } else {
+            // This must be an initial event, i.e. having index 0
```
pierrechevalier83: Maybe you could replace this comment with a `log_or_panic` if the event isn't initial?

src/gossip/event.rs
Fraser999: Hmm - I don't think so actually, since it can only be `Error::UnknownParent`.  I guess we'd be better to maybe log that inside `index_and_last_ancestors()` so we know which ancestor is missing.

src/gossip/event.rs
Fraser999: Yes, it could 👍   However, if I'm going to action https://github.com/maidsafe/parsec/pull/73/files#r214324572 I'd have to retain this more verbose form.

src/gossip/event.rs
Fraser999: I was going to, but because we've tightened up our type safety, the compiler is already [guaranteeing this](https://github.com/Fraser999/parsec/blob/master/src/gossip/content.rs#L37-L43).  I can still add the `log_or_panic!` if you want though?

src/gossip/event.rs
pierrechevalier83: Fair enough. If you're doing more things and need the `if`, go for it :wink:

src/gossip/event.rs
pierrechevalier83: This comment makes me feel like the `index` should be populated at construction and immutable.
It will be `0` when `new_from_initial` is called and `self_parent.index` + 1 in the other cases.

Is there an issue I'm missing with that approach?

PR #74, opened 2018-08-31
MAID-3011: add logging

Fraser999:


src/parsec.rs
```diff
@@ -103,6 +104,11 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         &self,
         peer_id: Option<S::PublicId>,
     ) -> Result<Request<T, S::PublicId>, Error> {
+        debug!(
+            "{:?} preparing gossip request for {:?}",
```
pierrechevalier83: Minor: I would say "creating" rather than "preparing" to have a closer mapping to the function name

PR #75, opened 2018-09-03
MAID-2993: extract 'sees' to new method

Fraser999:


src/parsec.rs
```diff
@@ -105,7 +105,7 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         peer_id: Option<S::PublicId>,
     ) -> Result<Request<T, S::PublicId>, Error> {
         debug!(
-            "{:?} preparing gossip request for {:?}",
+            "{:?} creating gossip request for {:?}",
```
pierrechevalier83: I don't think you meant to make this change in this commit, but I'll close an eye because it's so minor.

src/parsec.rs
```diff
@@ -740,12 +740,11 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-            .filter(|(peer_id, ancestor)| {
+            .filter(|(peer_id, &event_index)| {
                 self.peer_list
-                    .event_by_index(peer_id, **ancestor)
-                    .and_then(|hash| self.get_known_event(hash).ok())
-                    .and_then(|event| event.last_ancestors().get(y.creator()))
-                    .map_or(false, |last_index| *last_index >= y.index())
+                    .event_by_index(peer_id, event_index)
+                    .and_then(|event_hash| self.get_known_event(event_hash).ok())
+                    .map_or(false, |ancestor| ancestor.sees(y))
```
pierrechevalier83: I think the name `ancestor` here is a bit misleading because it makes me think: the descendant should see its ancestor and not vice-versa...
If I look a few lines up and see: `x.last_ancestors()`, things get cleared up; but this had a cognitive overhead.

I'm wondering if a name like `last_ancestor` or `last_ancestor_from_peer` wouldn't avoid the distraction ?

Fraser999: Agreed.  I'd prefer `last_ancestor` or `last_ancestor_of_x`.

pierrechevalier83: last_ancestor_of_x :+1:

PR #77, opened 2018-09-03
Decouple testing framework from implementation

nbaksalyar:
This PR makes the further integration of the FFI testing much easier and allows for code reuse.

I.e. instead of introducing a new test suite and duplicating the code, we can just make a new implementation struct calling into FFI functions.

### DIFF
--- a/tests/utils/peer.rs
+++ b/tests/utils/peer.rs
@@ -6,23 +6,23 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
+use super::{BlockImpl, ParsecImpl};
 use parsec::mock::{PeerId, Transaction};
-use parsec::{Block, Parsec};
 use std::collections::BTreeSet;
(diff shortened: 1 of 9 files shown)


tests/utils/peer.rs
```diff
@@ -43,12 +43,12 @@ impl Peer {
     }
 
     // Returns the payloads of `self.blocks` in the order in which they were returned by `poll()`.
-    pub fn blocks_payloads(&self) -> Vec<&Transaction> {
-        self.blocks.iter().map(Block::payload).collect()
+    pub fn blocks_payloads(&self) -> Vec<Transaction> {
```
pierrechevalier83: Returning `Transaction`s by value rather than by ref,
- Is it intentional?
- Could you explain the motivation? (I'm guessing it may be necessary with some kinds of bindings ?)

pierrechevalier83:
This work was de-prioritized, so it doesn't look like it will be picked up anytime soon. Closing until the timing is better.

PR #79, opened 2018-09-04
MAID-2999: allow custom quorum function

Fraser999:
This allows a custom function to be given for Parsec to use to determine if enough votes for a given payload have been received to class the corresponding block as valid.

A basic supermajority function is exported which can be used to retain the current behaviour.


src/parsec.rs
```diff
@@ -20,6 +20,15 @@ use serialise;
 use std::collections::{BTreeMap, BTreeSet, VecDeque};
 use std::iter;
 
+pub type QuorumFn<P> = fn(voters: &BTreeSet<&P>, current_peers: &BTreeSet<&P>) -> bool;
```
pierrechevalier83: I understand that in our use case, we generally think of a quorum as the input to the consensus process, but I think a motivation behind this task is to allow a more general way to distinguish "interesting events".

I would extract a bit more functionality (basically all the logic that checks that the event makes a block valid) and call it: `InterestingEventFn`.

The idea is to allow PARSEC to operate on the same premise as most ABFT consensus protocols out there: you can reach consensus on the order of individual transactions. The logic with making a block valid is specific to our use case.

src/gossip/event.rs
```diff
@@ -29,8 +29,8 @@ pub(crate) struct Event<T: NetworkEvent, P: PublicId> {
     index: u64,
     // Index of each peer's latest event that is an ancestor of this event.
     last_ancestors: BTreeMap<P, u64>,
-    // Payloads of all the blocks made valid by this event
-    pub valid_blocks_carried: BTreeSet<T>,
+    // Payloads of all the votes deemed interesting by this event
```
pierrechevalier83: `s/votes/gossip events/` ?

src/parsec.rs
```diff
@@ -283,23 +284,23 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
     }
 
     fn set_valid_blocks_carried(&mut self, event_hash: &Hash) -> Result<(), Error> {
-        let valid_blocks_carried = self.newly_valid_blocks_carried(event_hash)?;
-        if !valid_blocks_carried.is_empty() {
+        let interesting_content = self.interesting_content(event_hash)?;
```
pierrechevalier83: maybe I would name this one:
`my_interesting_content` to distinguish it from `event.interesting_content` further down

src/parsec.rs
```diff
@@ -283,23 +284,23 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
     }
 
     fn set_valid_blocks_carried(&mut self, event_hash: &Hash) -> Result<(), Error> {
```
pierrechevalier83: `s/set_valid_blocks_carried/set_interesting_content/` ?
Could be done in another fixup commit after squashing this one.

src/gossip/event.rs
Fraser999: Well, conceptually I think they're the same thing.  I went for "votes" because they do actually have a "payload" getter whereas events don't.

Not a big deal though - I can change this if you still prefer.

src/parsec.rs
Fraser999: Hmm - I'd actually go the other way.  Since we're assigning `interesting_content` to `event.interesting_content` it makes it clearer in my opinion to use the same name.

Again, happy to change if you disagree :)

src/parsec.rs
Fraser999: Can't defend this one! :D  Was an oversight.

src/gossip/event.rs
pierrechevalier83: OK. As it stands, it's fine then.

This `BTreeSet<T>` could end up becoming generic though. When it does, we can think of a better (more generic) wording.

src/parsec.rs
pierrechevalier83: No, if you prefer as it, that's fine with me.

src/parsec.rs
```diff
@@ -268,80 +284,90 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
                 }).or_insert_with(|| iter::once(*event_hash).collect());
         }
         self.get_known_event_mut(event_hash)
-            .map(|ref mut event| event.valid_blocks_carried = valid_blocks_carried)
+            .map(|ref mut event| event.interesting_content = interesting_content)
+    }
+
+    // Any payloads which this event sees as "interesting".  If this returns a non-empty set, then
+    // this event is classed as an interesting one.
+    fn interesting_content(&self, event_hash: &Hash) -> Result<BTreeSet<T>, Error> {
```
pierrechevalier83: What's the plan related to this return type becoming generic (will have implications until PARSEC's API)?
Are we keeping it for a subsequent PR?

Fraser999: I've thought about it a bit, and don't see much point in changing what we currently have actually.  I'm happy making any related changes part of this PR if others disagree though.

pierrechevalier83: Happy to keep it in mind for later

PR #81, opened 2018-09-04
MAID-2970 feat/parsec: parse event graph info from dot file

maqi:
Reading a dumped dot file and parse event graph info from it.
This is to be used for re-construct a parsec for testing purpose.

### DIFF
--- a/src/dot_parser.rs
+++ b/src/dot_parser.rs
@@ -1,0 +1,410 @@
+// Copyright 2018 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use gossip::{Cause, Event};
+use hash::Hash;
+use meta_vote::{BoolSet, MetaVote, Step};
+use mock::{PeerId, Signature, Transaction};
+use observation::Observation;
+use std::collections::{BTreeMap, BTreeSet};
+use std::fs::File;
+use std::io::{self, Read};
+use std::iter;
+use std::path::Path;
+use vote::Vote;
+
+#[allow(unused)]
+/// The event graph and associated info that were parsed from the dumped dot file.
+pub struct ParsedContents {
+    events: BTreeMap<Hash, Event<Transaction, PeerId>>,
+    events_order: Vec<Hash>,
+    meta_votes: BTreeMap<Hash, BTreeMap<PeerId, Vec<MetaVote>>>,
+}
+
+/// Reading a dumped dot file and return with parsed event graph and associated info.
+pub fn parse_dot_file(filename: &Path) -> io::Result<ParsedContents> {
+    read(File::open(filename)?)
+}
+
+#[derive(Clone, Debug)]
+struct ParsingEvent {
+    creator: String,
+    index: u64,
+    cause: String,
+    self_parent: Option<String>,
+    other_parent: Option<String>,
+    interesting_content: BTreeSet<Transaction>,
+    last_ancestors: BTreeMap<PeerId, u64>,
+    observations: BTreeSet<PeerId>,
+}
+
+impl ParsingEvent {
+    #[cfg_attr(feature = "cargo-clippy", allow(too_many_arguments))]
+    fn new(
+        creator: String,
+        index: u64,
+        cause: String,
+        self_parent: Option<String>,
+        other_parent: Option<String>,
+        blocks_carried_string: &str,
+        last_ancestors_string: &str,
+        observations: BTreeSet<PeerId>,
+    ) -> Self {
+        let split_vbc = extract_between(blocks_carried_string, "{", "}");
+        let interesting_content = if split_vbc.is_empty() {
+            BTreeSet::new()
+        } else {
+            split_vbc
+                .split(',')
+                .map(|s| Transaction::new(extract_between(s, "(", ")")))
+                .collect::<BTreeSet<_>>()
+        };
+
+        let mut last_ancestors = BTreeMap::new();
+        let split_la = extract_between(last_ancestors_string, "{", "}");
+        if !split_la.is_empty() {
+            for entry in split_la.split(',').map(|s| s.trim()) {
+                let split_entry = entry.split(": ").collect::<Vec<_>>();
+                let _ = last_ancestors
+                    .insert(PeerId::new(split_entry[0]), unwrap!(split_entry[1].parse()));
+            }
+        }
+
+        ParsingEvent {
+            creator,
+            index,
+            cause,
+            self_parent,
+            other_parent,
+            interesting_content,
+            last_ancestors,
+            observations,
+        }
+    }
+}
+
+fn read(mut file: File) -> io::Result<ParsedContents> {
+    let mut contents = String::new();
+    let _ = file.read_to_string(&mut contents)?;
+
+    let mut parsing_events = parse_event_graph(&contents);
+    let mut parsing_mvs = parse_meta_votes(&contents);
+
+    let mut events = BTreeMap::new();
+    let mut events_order = Vec::new();
+    let mut name_hash_map: BTreeMap<String, Hash> = BTreeMap::new();
+    let mut meta_votes = BTreeMap::new();
+    let length = parsing_events.len();
+
+    while !parsing_events.is_empty() {
+        let ordered_event = next_ordered_event(&parsing_events);
+        if ordered_event == String::default() {
+            assert!(parsing_events.is_empty());
+            break;
+        }
+        let event = unwrap!(parsing_events.remove(&ordered_event));
+
+        let cause = construct_cause(event.clone(), &name_hash_map);
+
+        let interesting_content = event
+            .interesting_content
+            .iter()
+            .map(|transction| Observation::OpaquePayload(transction.clone()))
+            .collect::<BTreeSet<_>>();
+
+        let parsed_event: Event<Transaction, PeerId> = Event::new_from_input(
+            PeerId::new(&event.creator),
+            cause,
+            event.index,
+            Signature::new(""),
+            event.last_ancestors,
+            interesting_content,
+            BTreeSet::new(),
+        );
+
+        let hash = *parsed_event.hash();
+
+        if let Some(meta_vote) = parsing_mvs.remove(&ordered_event) {
+            let _ = meta_votes.insert(hash, meta_vote);
+        } else if let Some(parent_meta_vote) = parsed_event
+            .self_parent()
+            .and_then(|self_parent| meta_votes.get(self_parent))
+            .and_then(|meta_vote| Some(meta_vote.clone()))
+        {
+            let _ = meta_votes.insert(hash, parent_meta_vote);
+        }
+
+        let _ = events.insert(hash, parsed_event);
+        events_order.push(hash);
+        let _ = name_hash_map.insert(ordered_event.to_string(), hash);
+    }
+
+    assert_eq!(events.len(), length);
+
+    Ok(ParsedContents {
+        events,
+        events_order,
+        meta_votes,
+    })
+}
+
+fn parse_event_graph(contents: &str) -> BTreeMap<String, ParsingEvent> {
+    let split_events = split_events(contents);
+    let meta_votes = split_meta_votes(contents);
+
+    let mut parsing_events = BTreeMap::new();
+    for (name, events) in &split_events.nodes {
+        for (index, event) in events.iter().enumerate() {
+            if let Some(info) = meta_votes.get(event) {
+                let self_parent: Option<String> = if index == 0 {
+                    None
+                } else {
+                    Some(events[index - 1].to_string())
+                };
+                let other_parent: Option<String> = split_events
+                    .other_parents
+                    .get(event)
+                    .and_then(|s| Some(s.to_string()));
+
+                let _ = parsing_events.insert(
+                    event.to_string(),
+                    ParsingEvent::new(
+                        name.to_string(),
+                        index as u64,
+                        info.cause.to_string(),
+                        self_parent,
+                        other_parent,
+                        info.interesting_content,
+                        info.last_ancestors,
+                        BTreeSet::new(),
+                    ),
+                );
+            }
+        }
+    }
+    parsing_events
+}
+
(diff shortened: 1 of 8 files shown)


src/dot_parser.rs
```diff
@@ -0,0 +1,370 @@
...
+    parsing_mvs
+}
+
+fn parse_meta_vote(line: &str) -> MetaVote {
+    // in the format of `{ 0/0, est:{f} bin:{t, f} aux:{f} dec:{} }`
+    let split_line = line.split(':').collect::<Vec<_>>();
+    let round: usize = split_line[0]
+        .split('/')
+        .nth(0)
+        .unwrap()
```
pierrechevalier83: True for the entire file: we prefer `unwrap!` to `unwrap()`

src/dot_parser.rs
```diff
@@ -0,0 +1,370 @@
+// Copyright 2018 MaidSafe.net limited.
```
pierrechevalier83: Could we split the long functions (`read` and `parse_event_graph` jump out) into no more than 25 lines? It's not a big change but it would really help readability (so I can focus my attention on one scope at a time).

src/meta_vote.rs
```diff
@@ -17,6 +17,16 @@ pub(crate) enum Step {
     GenuineFlip,
 }
 
+impl Step {
+    pub(crate) fn from_index(index: u8) -> Self {
+        match index {
+            1 => Step::ForcedFalse,
```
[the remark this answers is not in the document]
maqi: it's only used for testing purpose, so I think we shall panic directly?
I will list all option then raise a panic.

src/gossip/event.rs
```diff
@@ -38,6 +38,31 @@ pub(crate) struct Event<T: NetworkEvent, P: PublicId> {
 }
 
 impl<T: NetworkEvent, P: PublicId> Event<T, P> {
+    // Creates a new event using the input parameters directly
+    #[allow(unused)]
```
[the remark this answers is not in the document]
maqi: It was to avoid complaining when compile non-test build.
however, with moving dat_parser test from integration_test into unit test, this can now be put inside `cfg[test]` guard and no longer need the `unused`.

src/dot_parser.rs
```diff
@@ -0,0 +1,381 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::path::Path;
+
+    #[ignore]
+    #[test]
+    fn dot_parser() {
+        assert!(parse_dot_file(&Path::new("./input.dot")).is_ok());
+    }
```
octol: I takit it this is temporary until the real unit tests land?

src/lib.rs
```diff
@@ -92,6 +92,12 @@ mod peer_list;
 mod round_hash;
 mod vote;
 
+#[doc(hidden)]
```
Fraser999: We don't need this `#[doc(hidden)]` since the following comments won't appear in the generated docs anyway (it's a private mod, and also docs don't normally get generated by building with `test` feature enabled).

src/lib.rs
```diff
@@ -92,6 +92,12 @@ mod peer_list;
...
+/// **NOT FOR PRODUCTION USE**
```
Fraser999: This can get removed too, since it now can't be used for production use.

src/mock.rs
```diff
@@ -32,6 +38,15 @@ impl PeerId {
...
         Self { id: id.to_string() }
     }
+
+    pub fn from_initial(initial: char) -> Self {
+        for name in NAMES.iter() {
+            if name.starts_with(initial) {
+                return PeerId::new(name);
+            }
+        }
+        PeerId::new("")
```
Fraser999: Wouldn't it be better to panic here?

src/vote.rs
```diff
@@ -21,6 +21,11 @@ pub struct Vote<T: NetworkEvent, P: PublicId> {
 }
 
 impl<T: NetworkEvent, P: PublicId> Vote<T, P> {
+    /// Creates a `Vote` with input parameters.
+    pub fn new_from_input(payload: Observation<T, P>, signature: P::Signature) -> Self {
```
Fraser999: This should be annotated with `#[cfg(test)]` and the comment updated to say that it's only used by the dot_parser.

src/meta_vote.rs
```diff
@@ -53,7 +65,7 @@ impl BoolSet {
         *self == BoolSet::Empty
     }
 
-    fn insert(&mut self, val: bool) -> bool {
+    pub(crate) fn insert(&mut self, val: bool) -> bool {
```
Fraser999: I don't think we need this change.

src/meta_vote.rs
```diff
@@ -82,7 +94,7 @@ impl BoolSet {
         }
     }
 
-    fn from_bool(val: bool) -> Self {
+    pub(crate) fn from_bool(val: bool) -> Self {
```
Fraser999: I think we should revert this and in `parse_boolset()` just use:
```rust
match line {
    "{t}" => BoolSet::Single(true),
    "{f}" => BoolSet::Single(false),
...
```

src/meta_vote.rs
```diff
@@ -17,6 +17,18 @@ pub(crate) enum Step {
     GenuineFlip,
 }
 
+#[cfg(test)]
+impl Step {
```
Fraser999: This could be implemented inside `dot_parser::parse_meta_vote` like:
```rust
let step = {
    match step_index {
        0 => Step::ForcedTrue,
        1 => Step::ForcedFalse,
        2 => Step::GenuineFlip,
        _ => panic!("Improper Step index {:?}", step_index),
    }
};
```
and I think it belongs there personally.

src/gossip/cause.rs
```diff
@@ -9,11 +9,12 @@
 use hash::Hash;
 use id::PublicId;
 use network_event::NetworkEvent;
+use std::fmt::{self, Display, Formatter};
 use vote::Vote;
 
 #[serde(bound = "")]
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
-pub(super) enum Cause<T: NetworkEvent, P: PublicId> {
+pub(crate) enum Cause<T: NetworkEvent, P: PublicId> {
```
Fraser999: Rather than increasing the scope of `Cause` like this, I think we could restrict the impact by handling this all inside the context of `Event` which is already handling special cases for `dump-graphs`.

For writing the `cause()` to the dot file, we don't really need the actual `Cause`, we could expose a fn in `Event` restricted to `#[cfg(feature = "dump-graphs")]` which outputs the cause as a `String`, or maybe better still, does all the stuff required by `dump_graph::write_full()` so that in there it just calls `event.dump()` and that writes the three lines to a String or the `writer`.

src/gossip/event.rs
```diff
@@ -30,7 +30,7 @@ pub(crate) struct Event<T: NetworkEvent, P: PublicId> {
     // Sequential index of this event: this event is the `index`-th one made by its creator.
     index: u64,
     // Index of each peer's latest event that is an ancestor of this event.
-    last_ancestors: BTreeMap<P, u64>,
+    pub last_ancestors: BTreeMap<P, u64>,
```
Fraser999: This should be reverted since it's only needed for `dump-graphs`.  I'd favour handling all dumping inside a new method so we don't have to expose `last_ancestors` or `cause`.

If not, we can keep this member private, and have a pub getter which is gaurded with `#[cfg(feature = "dump-graphs")]`.

src/gossip/event.rs
```diff
@@ -155,6 +155,11 @@ impl<T: NetworkEvent, P: PublicId> Event<T, P> {
         }
     }
 
+    #[allow(unused)]
```
Fraser999: This should be `#[cfg(feature = "dump-graphs")]` instead, but I'd favour not exposing `cause` at all (see other comments).

src/gossip/cause.rs
```diff
@@ -32,3 +33,18 @@ pub(super) enum Cause<T: NetworkEvent, P: PublicId> {
...
+
+impl<T: NetworkEvent, P: PublicId> Display for Cause<T, P> {
+    fn fmt(&self, formatter: &mut Formatter) -> fmt::Result {
+        write!(
+            formatter,
+            "{}",
+            match &self {
+                Cause::Request { .. } => "Request".to_string(),
+                Cause::Response { .. } => "Response".to_string(),
+                Cause::Observation { vote, .. } => format!("Observation({:?})", vote.payload()),
```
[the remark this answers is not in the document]
maqi: then the print out of a `cause` will be:
Request/Response/Initial/Genesis(_)/Add(_)/Remove(_)/Genesis(_)

which is kind of mixing two levels of info into one?

src/gossip/cause.rs
maqi: The increased scope of `Cause` is mainly not because of the dump, but due to the `parser` ?

src/gossip/cause.rs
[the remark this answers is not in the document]
Fraser999: True.  I'd thought the enum type got printed too, so we'd see e.g. `Observation(Observation::Initial)`, but you're correct.

src/gossip/cause.rs
Fraser999: Meh - that's pretty disappointing :(  I don't know of a way to have different scopes depending on which features are enabled.

Seems like we could still avoid this by having `Event::new_from_input()` handle the work being done by `dot_parser::construct_cause()`?

src/dot_parser.rs
```diff
@@ -0,0 +1,387 @@
...
+                }
+            }
+
+            (name, events)
+        }).collect::<Vec<_>>();
+    (nodes, other_parents)
+}
+
+// Returns with: BTreeMap<event_hash, (cause, interesting_content, last_ancestors)>
+fn split_meta_votes<'a>(contents: &'a str) -> BTreeMap<&'a str, (&'a str, &'a str, &'a str)> {
```
madadam: Again, I'd use struct instead of the `(&str, &str, &str)` tuple.

src/dot_parser.rs
```diff
@@ -0,0 +1,387 @@
...
+    let mut meta_votes = BTreeMap::new();
+    let _ = contents
+        .split("/// {")
+        .filter(|s| s.contains("/// }"))
+        .map(|s| unwrap!(s.split("/// }").next()))
+        .map(|s| s.trim())
+        .map(|s| {
+            let split_info = s.split("///").map(|s| s.trim()).collect::<Vec<_>>();
+            let _ = meta_votes.insert(split_info[0], (split_info[1], split_info[2], split_info[3]));
+        }).collect::<Vec<_>>();
```
madadam: Instead of `collect`ing into `Vec` which is then thrown away, why not just iterate using `for` ?

src/dot_parser.rs
```diff
@@ -0,0 +1,387 @@
...
+fn split_meta_votes<'a>(contents: &'a str) -> BTreeMap<&'a str, (&'a str, &'a str, &'a str)> {
+    let mut meta_votes = BTreeMap::new();
+    let _ = contents
+        .split("/// {")
+        .filter(|s| s.contains("/// }"))
+        .map(|s| unwrap!(s.split("/// }").next()))
+        .map(|s| s.trim())
+        .map(|s| {
+            let split_info = s.split("///").map(|s| s.trim()).collect::<Vec<_>>();
+            let _ = meta_votes.insert(split_info[0], (split_info[1], split_info[2], split_info[3]));
```
madadam: You can avoid allocating the `Vec` if you instead do something like this:
```rust
meta_votes.insert(split_info.next().unwrap(), (split_info.next().unwrap(), split_info.next().uwnrap(), split_info.next().unwrap()));
```

src/dot_parser.rs
madadam: Or alternatively `collect` into the `BTreeMap` directly.

src/dot_parser.rs
```diff
@@ -0,0 +1,387 @@
...
+use mock::{PeerId, Signature, Transaction};
+use observation::Observation;
+use std::collections::{BTreeMap, BTreeSet};
+use std::fs::File;
+use std::io::{self, Read};
+use std::iter;
+use std::path::Path;
+use vote::Vote;
+
+#[allow(unused)]
```
madadam: I'd consider moving this whole file into the `dev_utils` sub-crate. Then this `allow(unused)` wouldn't be necessary.

src/dot_parser.rs
```diff
@@ -0,0 +1,389 @@
...
+                parse_meta_vote(split_line[1])
+            } else {
+                parse_meta_vote(line)
+            };
+            let _ = parsing_mvs
+                .entry(event_name.clone())
+                .and_modify(|peer_votes| {
+                    let _ = peer_votes
+                        .entry(peer_id.clone())
+                        .and_modify(|meta_votes| meta_votes.push(meta_vote.clone()))
```
madadam: Is the `clone` necessary here (also on the following lines)?

src/dot_parser.rs
```diff
@@ -0,0 +1,389 @@
...
+        .map(|s| {
+            let split_info = s.split("///").map(|s| s.trim()).collect::<Vec<_>>();
+            let _ = meta_votes.insert(split_info[0], (split_info[1], split_info[2], split_info[3]));
+        }).collect::<Vec<_>>();
+    meta_votes
+}
+
+fn parse_meta_vote(line: &str) -> MetaVote {
+    // in the format of `{ 0/0, est:{f} bin:{t, f} aux:{f} dec:{} }`
+    let split_line = line.split(':').collect::<Vec<_>>();
```
madadam: You can avoid the `Vec` allocation here here if you replace `split_line[n]` with `split_line.next().unwrap()` in the following lines.

src/dot_parser.rs
maqi: yes

src/dot_parser.rs
maqi: this is only used for testing, so I will prefer `collect then indexing` to `next` as less code complex.

src/dot_parser.rs
maqi: that will cause more stuff to be exposed out of the crate.

src/dot_parser.rs
maqi: this is only used for testing, so I will prefer `collect then indexing` to `next` as less code complex.

src/dot_parser.rs
madadam: I see. It's a limitation of using lambdas here. Could be avoided by rewriting the code using match and handling the `Occupied` and `Vacant` cases explicitly. Would make the code bit more verbose though. Don't feel strongly about this, so I'm leaving it up to you. I'll accept the PR regardless.

src/dot_parser.rs
madadam: It could be argued that performance is important in tests too, because we want the test suites to finish as quickly as possible. In this case the actual performance difference is probably negligible, so if you like this approach better, I'm OK with it too.

PR #82, opened 2018-09-04
MAID-3139: Send the Schedule output to lower log level

octol:
Send the Schedule output to a lower log level to avoid hiding the
directory where the graphs go from the output.

Note that the Schedule output is still logged to schedule.txt when
dumping graphs.

MAID-3139

### DIFF
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -54,6 +54,8 @@
     variant_size_differences
 )]
 
+#[macro_use]
+extern crate log;
 extern crate maidsafe_utilities;
 extern crate parsec;
 extern crate rand;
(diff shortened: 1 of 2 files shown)


tests/integration_tests.rs
```diff
@@ -271,6 +273,7 @@ proptest! {
         },
         ..Default::default()
     }) {
+        let _ = maidsafe_utilities::log::init(true);
```
maqi: just to confirm, the other two `random_schedule` tests don't need the schedule to be output to trace, right?

octol: I don't think so, but I could be mistaken...

PR #83, opened 2018-09-05
MAID-3016:  add benchmarks using criterion

madadam:
This PR adds a minimal benchmark based on the `minimal_network` integration tests. It should serve as a foundation upon which to build more sophisticated benchmarks later on. 

It also extracts the `utils` module from tests into its own sub-crate called dev_utils. This is so the module can be reused in both integration tests and benchmarks.


dev_utils/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+homepage = "https://maidsafe.net"
+license = "GPL-3.0"
+name = "parsec_dev_utils"
+repository = "https://github.com/maidsafe/parsec"
+version = "0.0.1"
+
+[dependencies]
+log = "~0.3.8"
+maidsafe_utilities = "~0.17.0"
+parsec = { path = ".." }
```
maqi: shall this be changed to use a proper version before merge?

madadam: I'd say it should always use the latest version. It's an internal crate, not to be published to crates.io anyway, so I'd say it's fine like this.

dev_utils/src/environment.rs
```diff
@@ -42,23 +42,17 @@ impl Environment {
...
         seed: RngChoice,
     ) -> Self {
-        let rng: Box<RngDebug> = match seed {
+        let mut rng: Box<RngDebug> = match seed {
             RngChoice::SeededRandom => Box::new(SeededRng::new()),
             RngChoice::Seeded(seed) => Box::new(SeededRng::from_seed(seed)),
-            RngChoice::SeededXor(seed) => Box::new(XorShiftRng::from_seed(seed)),
+            RngChoice::SeededXor(seed) => {
+                let rng = Box::new(XorShiftRng::from_seed(seed));
+                println!("Using {:?}", rng);
```
maqi: this will be printed during the normal test as well?
shall it be removed?

Fraser999: If we're using `XorShiftRng` and a test fails, we don't get the seed printed, so I think this should stay.

PR #88, opened 2018-09-07
MAID-3035: add test for the from_existing constructor

madadam:


src/parsec.rs
```diff
@@ -939,3 +941,28 @@ impl<T: NetworkEvent, S: SecretId> Drop for Parsec<T, S> {
...
+    #[test]
+    fn construct_from_existing() {
+        let mut peers = mock::create_ids(10);
+        let our_id = unwrap!(peers.pop());
+        let peers = peers.into_iter().collect();
+
+        let parsec =
+            Parsec::<Transaction, _>::from_existing(our_id.clone(), &peers, is_supermajority);
+
+        // Existing group + us
```
maqi: `Existing section` ? better be consistent with the wording in the `from_existing` function?

src/parsec.rs
```diff
@@ -120,8 +120,10 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         section: &BTreeSet<S::PublicId>,
         is_interesting_event: IsInterestingEventFn<S::PublicId>,
     ) -> Self {
+        let our_public_id = our_id.public_id().clone();
```
pierrechevalier83: Not sure a variable is necessary here. If we do want one, I would move it to right above the point at which it's used (i.e: swap these 2 lines)

madadam: I think it needs to be like it is, because `our_id` is consumed on the next line.

PR #91, opened 2018-09-11
MAID-3013: handle removing peers

Fraser999:


src/parsec.rs
```diff
@@ -323,8 +342,14 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
 
     fn unpack_and_add_events(
         &mut self,
+        src: &S::PublicId,
         packed_events: Vec<PackedEvent<T, S::PublicId>>,
     ) -> Result<(), Error> {
+        self.confirm_self_not_removed()?;
+        if self.peer_list.is_removed(src) {
```
madadam: Might be worth wrapping this into a `confirm_peer_not_removed` function, for symmetry with the `confirm_self_not_removed` function.

src/parsec.rs
```diff
@@ -362,7 +387,18 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
                 payload_hash
             );
             self.consensus_history.push(payload_hash);
+            let observation = block.payload().clone();
             self.consensused_blocks.push_back(block);
+            match observation {
```
madadam: If we moved this block above the `consensused_blocks.push_back(...)` line, we could avoid having to clone the payload.

src/peer_list.rs
```diff
@@ -73,12 +76,11 @@ impl<S: SecretId> PeerList<S> {
...
     }
 
-    /// Adds a peer into the map. If the peer has already been added as inactive
-    /// before, activate it.
+    /// Adds a peer into the map. If the peer has already been added as inactive, activate it.
     pub fn add_peer(&mut self, peer_id: S::PublicId) {
         match self.peers.entry(peer_id.clone()) {
             Entry::Occupied(mut entry) => {
-                entry.get_mut().active = true;
+                entry.get_mut().state = PeerState::Active;
```
madadam: Should we check for attempt to add a previously removed peer? Or do we want to allow it?

src/peer_list.rs
```diff
@@ -132,22 +146,42 @@ impl<S: SecretId> PeerList<S> {
...
+
     fn new_active() -> Self {
         Self {
-            active: true,
+            state: PeerState::Active,
             events: BTreeMap::new(),
         }
     }
 
     fn new_inactive() -> Self {
```
madadam: Maybe rename this to `new_pending` for consistency? Also check if we use the term `inactive` elsewhere and change that too.

src/parsec.rs
Fraser999: I toyed with doing that, but since it's only called twice I didn't feel it was worthwhile :)  I'll make the change though - it's probably slightly cleaner right enough.

src/parsec.rs
Fraser999: I can't do that since we want to return from this method if it's block indicating our own removal (line 396 below), but still need to push that block onto the consensused list regardless.

src/peer_list.rs
Fraser999: I think checks like that may make sense, but in theory it shouldn't happen in production code since the section shouldn't allow votes for adding a previously-removed peer.  Either way, I think it's probably outwith the remit of this PR?

src/peer_list.rs
```diff
@@ -132,22 +146,42 @@ impl<S: SecretId> PeerList<S> {
...
 pub(crate) struct Peer {
-    pub active: bool,
+    state: PeerState,
     pub events: BTreeMap<u64, Hash>,
 }
 
 impl Peer {
+    pub fn is_active(&self) -> bool {
+        match self.state {
+            PeerState::Active => true,
```
pierrechevalier83: I would have the enum be `PartialEq` to save a few lines (ymmv)

pierrechevalier83: I would have the enum be `PartialEq` to save a few lines (ymmv)

PR #92, opened 2018-09-11
MAID-2974: Add constructor initialising Parsec using dot_parser

octol:
To help with functional testing, create an extra constructor for Parsec
tagged that can initialise a partial Parsec with a few events from a
gossip graph using the dot_parser.

MAID-2974

### DIFF
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -995,9 +995,28 @@ impl<T: NetworkEvent, S: SecretId> Drop for Parsec<T, S> {
 }
 
 #[cfg(test)]
+use dot_parser::ParsedContents;
+#[cfg(test)]
+use mock::{PeerId, Transaction};
+
+#[cfg(test)]
+impl Parsec<Transaction, PeerId> {
+    #[allow(unused)]
+    pub(crate) fn from_parsed_contents(peer_id: PeerId, parsed_contents: ParsedContents) -> Self {
+        let mut parsec = Parsec::empty(peer_id, is_supermajority);
+        parsec.events = parsed_contents.events;
+        parsec.events_order = parsed_contents.events_order;
+        parsec.meta_votes = parsed_contents.meta_votes;
+        parsec
+    }
+}
+
+#[cfg(test)]
 mod tests {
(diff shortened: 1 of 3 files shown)


src/parsec.rs
```diff
@@ -995,4 +1014,32 @@ mod tests {
...
+        let peer_id = PeerId::new("Alice");
+        let parsec = Parsec::<Transaction, _>::from_parsed_contents(peer_id, parsed_contents);
+        assert_eq!(parsed_contents_comparison.events, parsec.events);
+        assert_eq!(parsed_contents_comparison.events_order, parsec.events_order);
+        assert_eq!(parsed_contents_comparison.meta_votes, parsec.meta_votes);
+    }
+
+    #[ignore]
+    #[test]
+    #[should_panic]
```
Fraser999: This shouldn't panic I guess?

octol: Well the intention was for the comparison to fail since the files are different?

Fraser999: Yes, but you're using `assert_ne!`, so it'll only panic if they _are_ equal.

octol: ah of course! silly me :)

src/parsec.rs
```diff
@@ -995,4 +1014,31 @@ mod tests {
 
         let _ = Parsec::<Transaction, _>::from_existing(our_id, &peers, is_supermajority);
     }
+
+    #[ignore]
+    #[test]
+    fn from_parsed_contents() {
+        let input_file = Path::new("./input-graphs/Alice-001.dot");
```
maqi: if we are going to use this input file as a `testing material`, then it shall be included in this PR as well.
and the test shall not be `ignored` as the PR is about the testing.

octol: Yes I think you're probably right. I wasn't sure where to place testing data since I don't think we have any in parsec? I'll see what I can come up with! :)

pierrechevalier83: Fraser offered to tackle as part of MAID-2975 so happy to merge as is in the end :smile:

PR #94, opened 2018-09-12
MAID-2971 test/dot_parser: test dot_parser functionalities

maqi:


src/gossip/event.rs
```diff
@@ -337,6 +336,9 @@ impl<T: NetworkEvent, P: PublicId> Debug for Event<T, P> {
     }
 }
 
+#[cfg(test)]
+use mock::{PeerId, Transaction};
```
Fraser999: All `use` statements should be at the top of the module I think?

src/dev_utils/dot_parser.rs
```diff
@@ -427,13 +429,74 @@ fn extract_between<'a>(input: &'a str, left: &str, right: &str) -> &'a str {
...
+            assert_eq!(gossip_graph, parsed_result.events);
+
+            // The dumped dot file doesn't contain all the meta_votes
+            assert!(meta_votes.len() >= parsed_result.meta_votes.len());
+            for (hash, meta_vote) in &parsed_result.meta_votes {
+                let ori_meta_vote = unwrap!(meta_votes.get(hash));
+                assert_eq!(ori_meta_vote, meta_vote);
+            }
+        }
+        assert_ne!(num_of_files, 0u8);
```
pierrechevalier83: I don't see the point of creating that mut variable incrementing it etc. Why not simply assert that `!entries.is_empty()` at line 470?

PR #95, opened 2018-09-12
MAID-2975: add functional testing framework

Fraser999:
Notes for reviewers:
* This differs slightly from the task's description since it doesn't actually generate SVG files as part of running the tests (unless the `dump-graphs` feature is enabled; but the SVGs will be written to `%TEMP%` as normal, not the repo).  Instead it checks that correct SVG files exist and it outputs the command to run if this check fails.  Running this in CI should have the desired effect of ensuring valid SVG files are committed for each unit test.
* I was unsure about the value of actually checking the contents of an existing SVG file against the output of `dot -Tsvg` since there may be a valid discrepancy if different versions of `dot` are involved.  However, I think there's a fairly high chance of a copy-paste error when creating a new unit test (e.g. copy an SVG from a different test and forget to update it to match the dot file) or forgetting to change the SVG if a dot file is modified.  I feel if we're debugging a failing test with an invalid SVG, it could waste quite a lot of time, so for now I feel this check is worthwhile.
* When a peer's debug ID is empty, I changed from using `E` to `?` since I think that better reflects the scenario.  I don't think that affects any parsing logic, but thought it worth highlighting.


src/dot_parser.rs
```diff
@@ -15,20 +15,70 @@ use std::collections::{BTreeMap, BTreeSet};
...
+    let mut svg_path = dot_path.clone();
+    assert!(svg_path.set_extension("dot.svg"));
+    let mut svg = unwrap!(
+        File::open(&svg_path),
+        "\nFailed to open SVG file for {0}.  Try running:\n    dot -Tsvg {0} -O\n",
+        dot_path.display()
+    );
+
+    // If we have "dot" available, check that the SVG contents are correct.
+    if let Ok(output) = Command::new("dot")
```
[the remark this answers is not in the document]
Fraser999: Yes, that's correct.

src/gossip/event.rs
```diff
@@ -200,6 +200,15 @@ impl<T: NetworkEvent, P: PublicId> Event<T, P> {
         }
     }
 
+    /// Returns the first char of the creator's ID, followed by an underscore and the event's index.
```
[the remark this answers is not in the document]
Fraser999: I think it's good to expose it for the use of dump_graph too since that's slightly more robust.  I also thought it was a better way of representing the event's name in the `Debug` impl below too.

I could hide this behind `#[cfg(any(test, feature = "testing", feature = "dump-graph"))]`, but I don't think it's harmful to just fully expose it.

src/gossip/event.rs
```diff
@@ -381,6 +386,24 @@ impl Event<Transaction, PeerId> {
     }
 }
 
+// TODO - remove
```
pierrechevalier83: What will it take to remove it? (btw, I would also have written it in the comment to pre-emptively answer reviewer's questions ;) )

src/dot_parser.rs
```diff
@@ -15,20 +15,73 @@ use std::collections::{BTreeMap, BTreeSet};
...
+    let mut dot_path = PathBuf::from("input_graphs");
+    dot_path.push(unwrap!(thread::current().name()).replace("::", "_"));
+    dot_path.push(filename);
+    assert!(
+        dot_path.exists(),
+        "\nDot file {} doesn't exist.",
+        dot_path.display()
+    );
+
+    // Check the corresponding SVG exists.
```
pierrechevalier83: style/nitpick: I would probably split the function into a few functions rather than resorting to comments. I won't insist if you disagree.

pierrechevalier83:
I guess more importantly, I agree with the 3 conclusions highlighted in your "note for reviewers"

src/gossip/event.rs
Fraser999: I expanded the comment :)

src/dot_parser.rs
Fraser999: Yeah, this one's borderline for me.  The full function fits on my screen (which is my definition of "not too big" 😄), and I know some people prefer comments to private functions which are only called from one place, but I'm not overly fussed either way.  Happy to split this up though.

PR #96, opened 2018-09-13
MAID-3064: handle malice - unexpected genesis

madadam:


src/parsec.rs
```diff
@@ -407,6 +410,7 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
                         return Ok(());
                     }
                 }
+                Observation::Accusation { .. } => {} // TODO: kick the offender out
```
pierrechevalier83: Let's discuss this tomorrow morning. I think we may leave that to routing.

PR #98, opened 2018-09-14
MAID-3152: Use safe_crypto signing key pair

octol:
To be able to properly test other components relying on signature
verification, plug in safe_crypto into PeerId by adding a public and
private signing pair. We want to keep the traits inside id.rs as they
are.

MAID-3152

### DIFF
--- a/src/mock.rs
+++ b/src/mock.rs
@@ -20,6 +20,13 @@ const NAMES: &[&str] = &[
     "Lucy", "Mike", "Nina", "Oran", "Paul", "Quin", "Rose", "Stan", "Tina",
 ];
 
+lazy_static! {
+    static ref PEERS: Vec<PeerId> = NAMES
+        .iter()
+        .map(|name| PeerId::new_with_random_keypair(name))
+        .collect();
+}
+
 /// **NOT FOR PRODUCTION USE**: Mock signature type.
 #[derive(Clone, Hash, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 pub struct Signature(SafeSignature);
@@ -42,6 +49,14 @@ pub struct PeerId {
 
 impl PeerId {
     pub fn new(id: &str) -> Self {
+        PEERS
+            .iter()
+            .find(|peer| peer.id == id)
+            .unwrap_or(&PeerId::new_with_random_keypair(id))
+            .clone()
+    }
+
+    fn new_with_random_keypair(id: &str) -> Self {
         let (pub_sign, sec_sign) = gen_sign_keypair();
         Self {
             id: id.to_string(),
--- a/src/mock.rs
+++ b/src/mock.rs
@@ -9,7 +9,11 @@
 use id::{PublicId, SecretId};
 use network_event::NetworkEvent;
 use rand::{Rand, Rng};
+use safe_crypto::Signature as SafeSignature;
+use safe_crypto::{gen_sign_keypair, PublicSignKey, SecretSignKey};
+use std::cmp::Ordering;
 use std::fmt::{self, Debug, Display, Formatter};
+use std::hash::{Hash, Hasher};
 
 const NAMES: &[&str] = &[
     "Alice", "Bob", "Carol", "Dave", "Eric", "Fred", "Gina", "Hank", "Iris", "Judy", "Kent",
@@ -17,20 +21,33 @@ const NAMES: &[&str] = &[
 ];
 
 /// **NOT FOR PRODUCTION USE**: Mock signature type.
-#[derive(Clone, Hash, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize, Debug)]
-pub struct Signature(String);
+#[derive(Clone, Hash, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct Signature(SafeSignature);
+
+impl Debug for Signature {
+    fn fmt(&self, formatter: &mut Formatter) -> fmt::Result {
+        write!(formatter, "..")
+    }
+}
 
 /// **NOT FOR PRODUCTION USE**: Mock type implementing `PublicId` and `SecretId` traits.  For
 /// non-mocks, these two traits must be implemented by two separate types; a public key and secret
 /// key respectively.
-#[derive(Clone, Hash, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+#[derive(Clone, Serialize, Deserialize)]
 pub struct PeerId {
     id: String,
+    pub_sign: PublicSignKey,
+    sec_sign: SecretSignKey,
 }
 
 impl PeerId {
     pub fn new(id: &str) -> Self {
-        Self { id: id.to_string() }
+        let (pub_sign, sec_sign) = gen_sign_keypair();
+        Self {
+            id: id.to_string(),
+            pub_sign,
+            sec_sign,
+        }
     }
 
     // Only being used by the dot_parser.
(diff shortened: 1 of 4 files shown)


src/mock.rs
```diff
@@ -54,10 +65,30 @@ impl Debug for PeerId {
...
+    fn cmp(&self, other: &PeerId) -> Ordering {
+        self.id.cmp(&other.id)
+    }
+}
+
 impl PublicId for PeerId {
     type Signature = Signature;
-    fn verify_signature(&self, _signature: &Self::Signature, _data: &[u8]) -> bool {
-        true
+    fn verify_signature(&self, signature: &Self::Signature, _data: &[u8]) -> bool {
```
octol: I'll update `_data` -> `data`

src/mock.rs
```diff
@@ -9,28 +9,60 @@
...
 pub struct PeerId {
     id: String,
+    pub_sign: PublicSignKey,
+    sec_sign: SecretSignKey,
 }
 
 impl PeerId {
     pub fn new(id: &str) -> Self {
-        Self { id: id.to_string() }
+        for name in NAMES.iter() {
```
pierrechevalier83: This `for + if` pattern can probably be replaced with `.find()`

src/mock.rs
```diff
@@ -9,28 +9,60 @@
...
 
 const NAMES: &[&str] = &[
     "Alice", "Bob", "Carol", "Dave", "Eric", "Fred", "Gina", "Hank", "Iris", "Judy", "Kent",
     "Lucy", "Mike", "Nina", "Oran", "Paul", "Quin", "Rose", "Stan", "Tina",
 ];
 
+lazy_static! {
+    static ref KEYPAIRS: HashMap<&'static str, (PublicSignKey, SecretSignKey)> = {
+        let mut m = HashMap::new();
+        for name in NAMES.iter() {
```
pierrechevalier83: `NAMES.map(...).collect()` would be more idiomatic

src/mock.rs
```diff
@@ -9,28 +9,60 @@
...
 use std::fmt::{self, Debug, Display, Formatter};
+use std::hash::{Hash, Hasher};
 
 const NAMES: &[&str] = &[
     "Alice", "Bob", "Carol", "Dave", "Eric", "Fred", "Gina", "Hank", "Iris", "Judy", "Kent",
     "Lucy", "Mike", "Nina", "Oran", "Paul", "Quin", "Rose", "Stan", "Tina",
 ];
 
+lazy_static! {
+    static ref KEYPAIRS: HashMap<&'static str, (PublicSignKey, SecretSignKey)> = {
```
pierrechevalier83: I would probably keep `NAMES` as is but replace this `KEYPAIRS` with a `Vec<PeerId>`. It would seem simpler and the complexity really doesn't matter since it's a fixed ~20 elements anyway.

src/mock.rs
```diff
@@ -9,28 +9,60 @@
...
+    pub_sign: PublicSignKey,
+    sec_sign: SecretSignKey,
 }
 
 impl PeerId {
     pub fn new(id: &str) -> Self {
-        Self { id: id.to_string() }
+        for name in NAMES.iter() {
+            if id == *name {
+                let (pub_sign, sec_sign) = (KEYPAIRS.get(name)).unwrap().clone();
```
octol: We're not supposed to use `.unwrap()`, but since this function is not marked `test` or `testing` we don't have access to the macro...

pierrechevalier83: Uhm. We're not supposed to use the macro or the function in production code. Only normal error handling (i.e: return a Result). That's one of our (slightly contentious) coding guidelines

src/mock.rs
pierrechevalier83: Or `log_or_panic`

src/mock.rs
```diff
@@ -9,28 +9,53 @@
...
 
 const NAMES: &[&str] = &[
     "Alice", "Bob", "Carol", "Dave", "Eric", "Fred", "Gina", "Hank", "Iris", "Judy", "Kent",
     "Lucy", "Mike", "Nina", "Oran", "Paul", "Quin", "Rose", "Stan", "Tina",
 ];
 
+lazy_static! {
+    static ref PEERS: Vec<PeerId> = NAMES
+        .iter()
+        .map(|name| PeerId::new_without_reuse(name))
```
pierrechevalier83: I think it would be the same if this called `new`

src/mock.rs
```diff
@@ -9,28 +9,53 @@
...
 impl PeerId {
     pub fn new(id: &str) -> Self {
-        Self { id: id.to_string() }
+        if let Some(peer) = PEERS.iter().find(|peer| peer.id == id) {
+            return peer.clone();
+        }
+        PeerId::new_without_reuse(id)
+    }
+
+    pub fn new_without_reuse(id: &str) -> Self {
```
pierrechevalier83: I had much rather see this function being private unless necessary.

src/mock.rs
```diff
@@ -9,28 +9,53 @@
...
 pub struct PeerId {
     id: String,
+    pub_sign: PublicSignKey,
+    sec_sign: SecretSignKey,
 }
 
 impl PeerId {
     pub fn new(id: &str) -> Self {
-        Self { id: id.to_string() }
+        if let Some(peer) = PEERS.iter().find(|peer| peer.id == id) {
```
pierrechevalier83: style: better
```
if x {
  a
} else {
  b
}
```
than
```
if x {
  return a;
}
b
```

The former is slightly more idiomatic: by not using `return`, you show you didn't corner yourself into a situation where you actually had to use an early return statement.

src/mock.rs
```diff
@@ -54,10 +79,37 @@ impl Debug for PeerId {
     }
 }
 
+impl Hash for PeerId {
```
pierrechevalier83: Do you still need this now you're using a `Vec`?

src/mock.rs
```diff
@@ -54,10 +79,37 @@ impl Debug for PeerId {
...
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.id.hash(state);
+        self.pub_sign.hash(state);
+    }
+}
+
```
pierrechevalier83: We may need them, but please double check that we do need these traits (`PartialEq`, `Eq`, `PartialOrd`, `Ord`)

src/mock.rs
```diff
@@ -54,10 +79,37 @@ impl Debug for PeerId {
...
+impl PartialEq for PeerId {
+    fn eq(&self, other: &PeerId) -> bool {
+        self.id == other.id && self.pub_sign == other.pub_sign && self.sec_sign == other.sec_sign
```
Fraser999: The point of implementing it by hand was to ensure it matched the custom impl for `Hash`, i.e. it should only consider `id` and `pub_sign`.  See the [clippy lint description](https://github.com/rust-lang-nursery/rust-clippy/wiki#derive_hash_xor_eq) for further info.  Same goes for `PartialOrd`.

src/mock.rs
pierrechevalier83: Or, as @Fraser999 pointed out, monads ftw :D (`unwrap_or_else`)

src/mock.rs
octol: As per the discussion yesterday, let's keep them :)

src/mock.rs
```diff
@@ -36,6 +43,14 @@ pub struct PeerId {
...
 impl PeerId {
     pub fn new(id: &str) -> Self {
+        PEERS
+            .iter()
+            .find(|peer| peer.id == id)
+            .unwrap_or(&PeerId::new_with_random(id))
+            .clone()
+    }
+
+    fn new_with_random(id: &str) -> Self {
```
pierrechevalier83: nitpick: `new_with_random_keypair` would be clearer imo

octol: Will update!

src/mock.rs
```diff
@@ -9,28 +9,54 @@
...
 
 impl PeerId {
     pub fn new(id: &str) -> Self {
-        Self { id: id.to_string() }
+        PEERS
+            .iter()
+            .find(|peer| peer.id == id)
+            .unwrap_or(&PeerId::new_with_random_keypair(id))
+            .clone()
+    }
```
octol: Yeah I must say this `unwrap_or` looks so much nicer than the previous `if` block!

src/mock.rs
```diff
@@ -9,28 +9,54 @@
...
+    static ref PEERS: Vec<PeerId> = NAMES
+        .iter()
+        .map(|name| PeerId::new_with_random_keypair(name))
+        .collect();
+}
+
 /// **NOT FOR PRODUCTION USE**: Mock signature type.
 #[derive(Clone, Hash, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize, Debug)]
-pub struct Signature(String);
+pub struct Signature(SafeSignature);
```
Fraser999: I think we should implement `Debug` for `Signature`, since what we see now is something like:
```
Signature(Signature { signature: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 39, 212, 57, 37, 204, 78, 139, 243] })
```
I don't think it's normally useful to see the actual value of a mock signature, so something as simple as the following might do:
```rust
impl Debug for Signature {
    fn fmt(&self, formatter: &mut Formatter) -> fmt::Result {
        write!(formatter, "..")
    }
}
```

octol: Makes sense! I'll get to it

PR #100, opened 2018-09-17
chore/test: testing env support add remove peer

maqi:


src/dev_utils/network.rs
```diff
@@ -26,6 +26,8 @@ struct QueueEntry {
 
 pub struct Network {
     pub peers: Vec<Peer>,
+    pub allow_add_error: bool,
```
pierrechevalier83: Still unsure what the purpose of these 2 flags is. Could you explain?

tests/integration_tests.rs
```diff
@@ -173,6 +173,53 @@ mod test {
...
+            &mut env,
+            &ScheduleOptions {
+                deterministic_failures: failures,
+                ..Default::default()
+            },
+        );
+        env.network.execute_schedule(schedule);
+
+        let result = env.network.blocks_all_in_sequence();
+        assert!(result.is_ok(), "{:?}", result);
```
Fraser999: Just `unwrap!(env.network.blocks_all_in_sequence());` would be equivalent and more succinct here.  Same below.

src/mock.rs
```diff
@@ -107,6 +107,11 @@ pub fn create_ids(count: usize) -> Vec<PeerId> {
     NAMES.iter().take(count).cloned().map(PeerId::new).collect()
 }
 
+pub fn get_id(pos: usize) -> PeerId {
+    assert!(pos < names_len());
```
Fraser999: Not sure there's much point to this line, since the following would panic with a fairly explicit message anyway.

PR #101, opened 2018-09-18
Maid 3034: test from_genesis

pierrechevalier83:


pierrechevalier83:
I'll assume @madadam's approval stands after only a `cargo fmt` of diff :smile:

PR #102, opened 2018-09-18
MAID-3159 chore/dot_parser: construct peer_list from file

maqi:


src/dev_utils/dot_parser.rs
```diff
@@ -212,13 +214,31 @@ fn read(mut file: File) -> io::Result<ParsedContents> {
 
     assert_eq!(events.len(), length);
 
+    let parsing_peer_states = parse_peer_states(&contents);
```
pierrechevalier83: s/parsing_peer_states/peer_states ?

src/dev_utils/dot_parser.rs
```diff
@@ -429,6 +449,9 @@ fn parse_option_bool(line: &str) -> Option<bool> {
 }
 
 fn extract_between<'a>(input: &'a str, left: &str, right: &str) -> &'a str {
+    println!("input {:?}", input);
```
pierrechevalier83: I don't think you want to keep these 3 lines.

src/dump_graph.rs
```diff
@@ -28,13 +29,14 @@ pub(crate) fn init() {
 /// printed to stdout.  The function will never panic, and hence is suitable for use in creating
 /// these files after a thread has already panicked, e.g. in the case of a test failure.  No-op for
 /// case where `dump-graphs` feature not enabled.
-pub(crate) fn to_file<T: NetworkEvent, P: PublicId>(
-    _owner_id: &P,
-    _gossip_graph: &BTreeMap<Hash, Event<T, P>>,
-    _meta_votes: &BTreeMap<Hash, BTreeMap<P, Vec<MetaVote>>>,
+pub(crate) fn to_file<T: NetworkEvent, S: SecretId>(
+    _owner_id: &S::PublicId,
```
pierrechevalier83: Rather than having all these underscores, I would probably prefer duplicating the function signature with and without `dump-graphs`.

(I realise it's not actually new here but had rather see that fixed still)

input_graphs/parsec_tests_from_parsed_contents/0.dot
```diff
@@ -1,6 +1,7 @@
 digraph GossipGraph {
   splines=false
   rankdir=BT
+/// PeerStates { (Alice, "Active")(Bob, "Active")(Carol, "Active")(Dave, "Active") }
```
pierrechevalier83: This syntax is inconsistent with other items in these comments such as `last_ancestors`. It's also a surprising syntax: almost looks like Lisp (burn!)

I would go for
```
peer_states: {Alice: "Active", Bob: "Active, Carol: "Active", Dave: "Active"}
```
for consistency

src/dump_graph.rs
```diff
@@ -413,4 +424,16 @@ mod detail {
...
         writeln!(writer, "}}")
     }
+
+    fn write_peerlist_to_dot<S: SecretId>(
+        writer: &mut Write,
+        peer_list: &PeerList<S>,
+    ) -> io::Result<()> {
+        write!(writer, "/// PeerStates {{ ")?;
+        for (peer_name, peer_state) in peer_list.iter() {
+            write!(writer, "({:?}, {:?})", peer_name, peer_state.print_state())?;
```
pierrechevalier83: this is where the consistency is lacking (see my comment on the dot syntax above). Maybe by being consistent, you can just reuse a function you've already written (or that you can extract) for printing a BTreeMap in general.

src/peer_list.rs
```diff
@@ -146,6 +148,46 @@ impl<S: SecretId> PeerList<S> {
     }
 }
 
+#[cfg(test)]
+impl PeerList<PeerId> {
+    // Creates a new PeerList using the input parameters directly
+    pub(super) fn new_from_dot_input(
+        events_graph: &BTreeMap<Hash, Event<Transaction, PeerId>>,
+        parsing_peer_states: &BTreeMap<PeerId, String>,
```
pierrechevalier83: again, `peer_states` would do as a name.

src/peer_list.rs
```diff
@@ -146,6 +148,46 @@ impl<S: SecretId> PeerList<S> {
...
+                "Active" => PeerState::Active,
+                "Removed" => PeerState::Removed,
+                _ => panic!("wrong state string: {:?}", state_str),
+            };
+            let _ = peers.insert(peer.clone(), Peer { state, events });
+            peer_id_hashes.push((Hash::from(serialise(peer).as_slice()), peer.clone()))
+        }
+
+        PeerList {
+            our_id: peer_id_hashes[0].1.clone(),
```
pierrechevalier83: It seems unexpected to me that the list of peers should always be sequenced in such a way that my id is first.

I had much rather have an extra comment line with:
```
our_id: Carol
```
than always having to reorder the list of peer states to fit that property. Also not sure they're dumped with me first, but even if they are I had rather be more explicit.

src/peer_list.rs
```diff
@@ -146,6 +148,46 @@ impl<S: SecretId> PeerList<S> {
...
+                }
+            }
+            let state = match state_str.as_ref() {
+                "Pending" => PeerState::Pending,
+                "Active" => PeerState::Active,
+                "Removed" => PeerState::Removed,
+                _ => panic!("wrong state string: {:?}", state_str),
+            };
+            let _ = peers.insert(peer.clone(), Peer { state, events });
+            peer_id_hashes.push((Hash::from(serialise(peer).as_slice()), peer.clone()))
```
pierrechevalier83: ~Maybe I would like a panic for instance if after doing all this, some events in the graph don't have a peer from the "peer states" line that created them.~

Actually a panic may not do (for instance if we want to test an `InvalidGossiCreator`). I would like for it to be logged then.

src/dev_utils/dot_parser.rs
```diff
@@ -212,13 +214,31 @@ fn read(mut file: File) -> io::Result<ParsedContents> {
...
         events_order,
         meta_votes,
+        peer_list,
     })
 }
 
+fn parse_peer_states(contents: &str) -> BTreeMap<PeerId, String> {
+    let states_string = extract_between(contents, "/// PeerStates", "/// {");
+    let mut peer_states = BTreeMap::new();
+    let _ = extract_between(states_string, "{", "}")
```
pierrechevalier83: this would change too if we make the syntax more consistent.

src/dump_graph.rs
maqi: The function will be called from the Parsec, and duplicate means we still need those `underscores` for the `non-dump-graphs` scenario.
If do want to remove the `underscore` totally, I think that means we will have to put the feature guard to couple of places in the Parsec production code.
Which I think is `messier` ?

src/peer_list.rs
maqi: This is a testing facilities, so I think `logging` doesn't make sense.
I will prefer leave it as it is and let the later on production code raises panic/error_report once checking.

src/dump_graph.rs
pierrechevalier83: Uhm. It's probably my lack of experience with rust showing, or maybe I just miscommunicated my suggestion. Would the following code have any impact on the calling site?

```
 #[cfg(feature = "dump-graphs")]
 pub(crate) fn to_file<T: NetworkEvent, S: SecretId>(
    owner_id: &S::PublicId,
    gossip_graph: &BTreeMap<Hash, Event<T, S::PublicId>>,
    meta_votes: &BTreeMap<Hash, BTreeMap<S::PublicId, Vec<MetaVote>>>,
    peer_list: &PeerList<S>,
) {
    detail::to_file(owner_id, gossip_graph, meta_votes, peer_list)
}
#[cfg(not(feature = "dump-graphs"))]
 pub(crate) fn to_file<T: NetworkEvent, S: SecretId>(_: &S::PublicId, _: &BTreeMap<Hash, Event<T, S::PublicId>>, _: &BTreeMap<Hash, BTreeMap<S::PublicId, Vec<MetaVote>>>, _: &PeerList<S>) {}
```

src/peer_list.rs
pierrechevalier83: My idea here is simply to help the test implementer notice that their input may not be as expected. I'm just saying it can't panic because it could be on purpose, but may make sense to log it to facilitate investigations.

src/peer_list.rs
```diff
@@ -146,6 +148,48 @@ impl<S: SecretId> PeerList<S> {
...
+        }
+
+        PeerList {
+            // The population of `our_id` here is just to avoid the value to be void.
```
pierrechevalier83: It is better with the comment, but I would still be in favour of passing our id down the call stack until here so we can rely on this struct being always in a consistent state rather than forcing the caller to patch it up after the fact.

src/dev_utils/dot_parser.rs
```diff
@@ -212,13 +226,25 @@ fn read(mut file: File) -> io::Result<ParsedContents> {
...
     Ok(ParsedContents {
         events,
         events_order,
         meta_votes,
+        peer_list,
     })
 }
 
+fn parse_peer_states(contents: &str) -> BTreeMap<PeerId, String> {
+    let states_string = extract_between(contents, "/// PeerStates", "/// {");
```
pierrechevalier83: This makes an assumption on where exactly the `PeerStates` line is in the comments.
Probably more robust to give `"/// PeerStates {"` and `"}"` as the boundaries and remove the `extract_between` from `parse_entries`. (Of course, it will also mean doing the same thing in other callers of `parse_entries`.

src/peer_list.rs
maqi: `our_id` is in the type of `S` and passing it down means it needs to penetrate through `dot_parser` as well. which means there are many places of type anonnation need to be changed.

src/peer_list.rs
```diff
@@ -146,6 +148,48 @@ impl<S: SecretId> PeerList<S> {
...
+            // The actual population happens at `Parsec::from_parsed_contents`.
+            our_id: peer_id_hashes[0].1.clone(),
+            peers,
+            peer_id_hashes,
+        }
+    }
+
+    // Update `our_id`
+    pub(super) fn set_id(&mut self, our_id: PeerId) {
```
Fraser999: I don't believe we should provide this setter, even for testing purposes.  I think it'd be better to pass the `PeerId` through `dot_parser::parse_test_dot_file()` rather than `Parsec::from_parsed_contents()`, so it can just be set properly at initialisation.

src/dev_utils/dot_parser.rs
maqi: `parse_entries` is also used for `interesting_content`, which contains `}` and will have extra code to remove it.
I think the whole parser made many assumptions of `locations`. As this is only a testing tool and the dot file shall only be generated using our own dump_graph toll, so I will prefer to leave it.

src/peer_list.rs
pierrechevalier83: Still feels better than partially constructing and relying on magic at the call sire imho...

maqi: will try to dump/parse our_id via the dot file.

input_graphs/parsec_tests_from_parsed_contents/0.dot
```diff
@@ -1,6 +1,8 @@
 digraph GossipGraph {
   splines=false
   rankdir=BT
+/// OurId: Alice
```
pierrechevalier83: OK. one last minor comment:
Should we make this syntax: `our_id` and `peer_states` rather than `OurId` and `PeerState`? We've used snake case consistently for the other fields, so my OCD field cringes at mixing styles...

PR #104, opened 2018-09-19
MAID-3014 test/remove_peer: functional tests of remove peer

maqi:
When consensus is reached on a vote for a node's removal, it should be
properly removed from peer_manager.
Attempts to call functions such as create_gossip on a node that was
removed must result in error.
Direct gossip from removed nodes should be rejected.


src/parsec.rs
```diff
@@ -1194,4 +1194,50 @@ mod tests {
...
+            Parsec::<Transaction, _>::from_existing(judy_id.clone(), &section, is_supermajority);
+        // Peer state is 'pending' when created from existing. Need to call 'add_peer' to update
+        // the state to 'active'.
+        for peer_id in section {
+            judy.peer_list.add_peer(peer_id);
+        }
+        let request = unwrap!(judy.create_gossip(Some(PeerId::new("Alice"))));
+        // Message from a removed peer shall result in error.
+        expected_error = unwrap_err!(alice.handle_request(&judy_id, request));
+        if let Error::RemovedPeer = expected_error {
```
pierrechevalier83: This is a strange way to put it. I would find it much more intuitive to make `Error` equality comparable (I'm assuming it's not) and use `==` or `!=`

input_graphs/parsec_functional_tests_remove_peer/remove_judy.dot
```diff
@@ -0,0 +1,4560 @@
+digraph GossipGraph {
```
pierrechevalier83: ~ 4500 lines of graphviz is quite far from the original pitch for functional tests.

The original idea was to be able to manually draw the specific section of the graph that's relevant, parse it and test the exact point that matters.
If we won't manually generate it, can we at least try and trim > 80% of the file to only keep the bits that are necessary for the test?

I will never read this dot file and there will be overhead of me looking for the interesting bit in the generated `.svg`.

src/parsec.rs
```diff
@@ -1194,4 +1194,50 @@ mod tests {
...
+
+        // Add event now which shall result in Alice removing Judy.
+        unwrap!(alice.add_event(a_40));
+        assert!(alice.peer_list.is_removed(&&judy_id));
+
+        // Try calling `create_gossip()` for Judy shall result in error.
+        let mut expected_error = unwrap_err!(alice.create_gossip(Some(judy_id.clone())));
+        if let Error::RemovedPeer = expected_error {
+        } else {
+            panic!("Expected UnknownPeer, but got {:?}", expected_error)
```
Fraser999: The panic message should mention`RemovedPeer`, not `UnknownPeer`.  Similar error in the one below too.

input_graphs/parsec_functional_tests_remove_peer/remove_judy.dot
maqi: will reduce the network size to 5 and re-generate the dot file.
A futher trim down of just keeping last few events will be investigated as a separate task.

PR #106, opened 2018-09-20
MAID-3036: add functional test for add_peer

Fraser999:


src/gossip/event.rs
```diff
@@ -364,6 +364,13 @@ impl Event<Transaction, PeerId> {
                 let content = unwrap!(unwrap!(cause.split('(').nth(2)).split(')').next());
                 let observation = if cause.contains("OpaquePayload") {
                     Observation::OpaquePayload(Transaction::new(content))
+                } else if cause.contains("Genesis") {
+                    // `content` will contain e.g. "{Alice, Bob, Carol, Dave, Eric}"
+                    let peer_ids = content[1..content.len() - 1]
+                        .split(", ")
+                        .map(PeerId::new)
+                        .collect();
+                    Observation::Genesis(peer_ids)
```
maqi: will the first and last peer_id to be `{Alice` and `Eric}`, or the curly bracket has been taken away already?

src/parsec.rs
```diff
@@ -372,7 +376,24 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         self.peer_list.add_event(&event)?;
         let event_hash = *event.hash();
         let is_initial = event.is_initial();
-        self.events_order.push(event_hash);
+        if !self.peer_list.is_active(self.our_pub_id()) && event.creator() != self.our_pub_id() {
+            // We're handling an event before we've been made active.  If it's not one of our own
```
maqi: what if the node was `removed` then receive events from others?
will this also make the event be inserted `jump the queue`?

src/parsec.rs
```diff
@@ -1002,7 +1023,16 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
             .skip_while(move |hash| !last_ancestors_hashes.contains(hash))
-            .filter_map(move |hash| self.get_known_event(hash).ok()))
+            .filter_map(move |hash| self.get_known_event(hash).ok())
+            .filter(move |event| {
+                // We only need to include events newer than those indicated by
+                // `peer_last_event.last_ancestors()`
+                peer_last_event
+                    .last_ancestors()
+                    .get(event.creator())
+                    .map(|&last_ancestor_index| event.index() > last_ancestor_index)
```
maqi: Is this an optimisation, or related to the add peer feature?

src/parsec.rs
```diff
@@ -1208,6 +1272,89 @@ mod functional_tests {
         assert_ne!(parsed_contents_other.meta_votes, parsec.meta_votes);
     }
 
+    #[test]
+    fn add_peer() {
+        let mut parsed_contents = parse_test_dot_file("add_fred.dot");
+        // Split out the events Eric would send to Alice.  These are the last seven events listed in
+        // `parsed_contents.events_order`, i.e. B_14, C_14, D_14, D_15, B_15, C_15, E_14, and E_15.
```
maqi: If this test is such coupled to the output graph. Will it worth to mention how the graph got generated (example with which config and seed)?

src/gossip/event.rs
Fraser999: I'm stripping the braces here `content[1..content.len() - 1]`

src/parsec.rs
Fraser999: We've already done `self.confirm_self_not_removed()?;` before calling this in both places, and I think that should always be the case before calling `add_event()`, so I'd rather not check inside here as well.  However, maybe to be as safe as possible I could put in a `is_pending()` getter for `peer_list::Peer` and use that here?

src/parsec.rs
Fraser999: Yes, it's an optimisation to avoid sending more data than is needed.

src/parsec.rs
Fraser999: I did consider that, and like the idea.  But I used a modified version of the example, so until the example can accept more args (e.g. to say make the observation "Add(Fred)", I don't think we should add even more code comments here.

I'm also not sure how painful it will be if we do provide a seed, but because of code changes the seed no longer generates a similar graph.

maqi: Yeah, maybe providing a seed is a overshoot.
However, I think some short description like: using example with network size of 5 and voting observation of `Add(Fred)` , shall be enough and convenient for later on re-generation of the graph?

Fraser999: I would say that's the point of the SVG file being included.

pierrechevalier83: Actually, in a different scope, I was considering adding the seed to the graph (like in a title or something). So when you see a .svg, you know how to reproduce it

PR #108, opened 2018-09-21
WIP MAID-3177 test/parsec: trim down the dot file

maqi:
Trim down the dot file for the functional test to contain only
the interested section of graph.


src/gossip/event.rs
```diff
@@ -354,13 +354,21 @@ impl Event<Transaction, PeerId> {
 
         let cause = match cause {
             "cause: Initial" => Cause::Initial,
-            "cause: Request" => Cause::Request {
-                self_parent: unwrap!(self_parent),
-                other_parent: unwrap!(other_parent),
+            "cause: Request" => if self_parent.is_none() || other_parent.is_none() {
```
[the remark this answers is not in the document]
maqi: I am ok with adding a comment.

Regarding `Observation`, it requires a mandatory self_parent, which means we cannot create such dangling observation event in the partial graph dot graph (i.e. it's self parent event must be included in the dot file.)

src/dev_utils/dot_parser.rs
```diff
@@ -338,21 +333,44 @@ fn parse_event_graph(contents: &str) -> BTreeMap<String, ParsingEvent> {
...
+            .meta_votes
+            .entry(peer_id)
+            .and_modify(|meta_votes| meta_votes.push(meta_vote.clone()))
+            .or_insert_with(|| iter::once(meta_vote).collect());
+    }
+}
+
+fn parse_meta_votes(contents: &str) -> BTreeMap<String, ParsingMetaVotes> {
     // The last dot file generated by a node contains no meta_votes.
     let mvs_lines = if let Some(content) = contents.split("\n\n").find(|s| s.contains("aux")) {
```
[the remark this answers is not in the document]
maqi: as you suggestted, I will add the replacement to UNIX ending immediately after reading the contents in.

src/gossip/event.rs
```diff
@@ -354,13 +354,23 @@ impl Event<Transaction, PeerId> {
 
         let cause = match cause {
             "cause: Initial" => Cause::Initial,
-            "cause: Request" => Cause::Request {
-                self_parent: unwrap!(self_parent),
-                other_parent: unwrap!(other_parent),
+            // For the dot file contains only partial graph, we have to manually change the info of
+            // ancestor to null for some events.
```
Fraser999: I'd say that the comment doesn't make it clear that we're deliberately interpreting events with missing parents as `Initial` events, even if their stated `cause` indicates otherwise.

I'm actually wondering about the validity of this approach, since the resulting event's hash is different when we interpret it as `Initial` rather than a `Request` or `Response`.  Maybe we should keep the events as the correct type, and work around the missing parents another way?

src/gossip/event.rs
```diff
@@ -354,13 +354,15 @@ impl Event<Transaction, PeerId> {
 
         let cause = match cause {
             "cause: Initial" => Cause::Initial,
+            // For the dot file contains only partial graph, we have to manually change the info of
+            // ancestor to null for some events. In that case, populate ancestors with empty hash.
             "cause: Request" => Cause::Request {
-                self_parent: unwrap!(self_parent),
-                other_parent: unwrap!(other_parent),
+                self_parent: self_parent.unwrap_or_else(Hash::all_zero),
```
Fraser999: I still think we should be doing this for `Observation`s too.

src/parsec.rs
```diff
@@ -1208,8 +1208,25 @@ impl<T: NetworkEvent, S: SecretId> Drop for Parsec<T, S> {
 
 #[cfg(test)]
 impl Parsec<Transaction, PeerId> {
+    fn populate_interesting_events(parsec: &mut Self, event: &Event<Transaction, PeerId>) {
```
Fraser999: I'm not really seeing the need for this to be a separate function, but if it stays separate, I think we should use the more canonical (&mut self, ...) form.

src/dev_utils/dot_parser.rs
```diff
@@ -338,21 +336,44 @@ fn parse_event_graph(contents: &str) -> BTreeMap<String, ParsingEvent> {
     parsing_events
 }
 
-fn parse_meta_votes(contents: &str) -> BTreeMap<String, BTreeMap<PeerId, Vec<MetaVote>>> {
+struct ParsingMetaVotes {
```
pierrechevalier83: Not a fan of the name... I don't necessarily have a great idea for a better name right now as I'm not sure what the index represents exactly... Maybe we can have a chat about it.

PR #112, opened 2018-09-26
MAID-3191: Fix error reporting in integration tests

madadam:


pierrechevalier83:
We discussed it and thought we could come up with a better solution wherein the actually divergent lists of blocks are always reported. Dropping this PR.

PR #118, opened 2018-09-28
MAID-3153 feat/parsec: store multiple elections in parallel

maqi:


src/block.rs
```diff
@@ -49,6 +51,11 @@ impl<T: NetworkEvent, P: PublicId> Block<T, P> {
         &self.proofs
     }
 
+    /// Returns the hash of this block.
+    pub fn hash(&self) -> Hash {
```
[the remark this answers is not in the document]
maqi: I'd prefer rename it. There is only one place so far, however may will have more places to be called later on with the parsec algorithm to be refactored to utilize the parallel storage.

src/parsec.rs
```diff
@@ -40,6 +40,73 @@ pub fn is_supermajority<P: Ord>(voters: &BTreeSet<&P>, current_peers: &BTreeSet<
...
+            .and_modify(|votes| {
+                let _ = votes.insert(event_hash, meta_votes);
+            });
+    }
+
+    fn last_meta_votes(&self) -> &MetaVotes<S::PublicId> {
+        if let Some(votes) = self.meta_votes.get(&self.last_stable_block_hash) {
+            votes
+        } else {
+            panic!(
```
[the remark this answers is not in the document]
maqi: if changes to `log_or_panic`, this function requires a return of reference to BTreeMap of meta_votes,
which we cannot in that case :(
any better idea istead of changing the return type to `Option`?

src/meta_vote.rs
```diff
@@ -422,3 +425,156 @@ impl MetaVoteCounts {
...
+    pub(super) fn last_meta_votes(&self) -> &MetaVotes<P> {
+        &unwrap!(self.meta_elections.last()).meta_votes
+    }
+
+    #[cfg(not(any(test, feature = "testing", feature = "dump-graphs")))]
+    pub(super) fn last_meta_votes(&self) -> &MetaVotes<P> {
+        if let Some(election) = self.meta_elections.last() {
+            &election.meta_votes
+        } else {
+            panic!("MetaVotesMap is empty!")
```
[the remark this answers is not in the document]
maqi: This function will be used in the test when dumping to dot_file during panic.
And panic during panic handling isn't preferred.  :(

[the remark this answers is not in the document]
fizyk20: But it will still panic in exactly the same circumstances as the other function, no? The other one unwrap, so it will panic if `last()` is `None`, and this one... will panic when `last()` is `None`.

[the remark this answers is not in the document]
maqi: Yeah, you are right. 
As mentioned at https://github.com/maidsafe/parsec/pull/118#discussion_r221287576
This is only trying to show another approach, however seems a `panic` is unavoidable then  :(

src/meta_vote.rs
```diff
@@ -422,3 +425,156 @@ impl MetaVoteCounts {
...
+            .filter_map(|election| {
+                if election.block_hash == Hash::all_zero() {
+                    None
+                } else {
+                    Some(election.block_hash)
+                }
+            }).collect()
+    }
+
+    pub(super) fn push_last_block_hash(&mut self, payload_hash: Hash) {
```
Fraser999: I'd rename this to `new_election()`.

src/meta_vote.rs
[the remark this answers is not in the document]
fizyk20: The function could be made to return an `Option`, which just moves the problem of unwrapping somewhere else, but maybe it would help?

src/meta_vote.rs
```diff
@@ -422,3 +425,156 @@ impl MetaVoteCounts {
...
+        self.meta_elections
+            .push(MetaElection::<P>::new(payload_hash))
+    }
+
+    pub(super) fn insert_into_last_election(
```
Fraser999: Just a nitpick, but I think "latest" is better than "last" for all these methods.  Since the one we're referring to is the current election, "last" _could_ be interpreted as the one before the current one.  Not a big deal though.  I think we should aim for consistency though, so in my suggestions above about e.g. `latest_meta_votes()`, if we're sticking with "last" then that too would be `last_meta_votes()`.

src/meta_vote.rs
```diff
@@ -422,3 +425,156 @@ impl MetaVoteCounts {
...
+    fn new(block_hash: Hash) -> Self {
+        MetaElection {
+            block_hash,
+            meta_votes: BTreeMap::new(),
+            round_hashes: BTreeMap::new(),
+        }
+    }
+}
+
+pub(super) struct MetaVotesMap<P: PublicId> {
```
Fraser999: I'd prefer if this were called `MetaElections`.  Seems slightly odd to me that as we go "up levels", we go `MetaVote` → `MetaVotes` → `MetaElection` → `MetaVoteMap`.

Also, having the inner type named as `meta_elections` seems a bit like stuttering to me (although since it's private we'd never actually be calling e.g. `meta_elections.meta_elections` outside of any unit tests here).  Probably making it unnamed is fine here?

src/meta_vote.rs
```diff
@@ -422,3 +425,156 @@ impl MetaVoteCounts {
...
+}
+
+impl<'a, P: 'a + PublicId> MetaVotesMap<P> {
+    pub(super) fn new() -> Self {
+        MetaVotesMap {
+            meta_elections: iter::once(MetaElection::<P>::new(Hash::all_zero())).collect(),
+        }
+    }
+
+    pub(super) fn get_mvs_from_last_election(
```
Fraser999: I think it'd be good to have a couple of (probably private) getters for `self.meta_elections.last()` and `self.meta_elections.last_mut()` which don't return an option and use `log_or_panic!`.  Then we can use those in all these "..._last_..." methods.  For the "log" case, we can just return an empty map.

src/meta_vote.rs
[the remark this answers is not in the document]
maqi: the function is being used in couple of place, and using a returned Option makes the code little bit messier.
It's more about a choice we face: allow panic for a actually won't happen situation , or allow a little bit messy code

src/meta_vote.rs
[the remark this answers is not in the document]
maqi: I will switch to `meta_votes_of_current_election()` and `round_hashes_of_current_election()` then

src/meta_vote.rs
maqi: as discussed, will use `current` to replace all `last` and `latest`

src/meta_vote.rs
maqi: As discussed, I will change MetaVoteMap/meta_vote_map to MetaVoteElections/meta_elections

src/meta_vote.rs
maqi: the return of `last()` or `last_mut()` is not a map but a `MetaElection`, which we can't provide an empty map to replace it?
Also, if we return an empty map on a `last_mut()`, won't this be a misleading?

src/meta_vote.rs
```diff
@@ -467,113 +467,103 @@ impl<'a, P: 'a + PublicId> MetaVotesMap<P> {
...
-            panic!("MetaVotesMap is empty!")
+            panic!("MetaElections is empty!")
+        }
+    }
+
+    fn current_election_mut(&mut self) -> &mut MetaElection<P> {
+        if let Some(election) = self.meta_elections.last_mut() {
+            election
+        } else {
+            panic!("MetaElections is empty!")
```
fizyk20: Maybe if it's empty, we could just create a new election for hash all_zero and return a reference to that? I'm fine with merging as is, though, this shouldn't happen and if it does, we will at least know about it ;)

maqi: I don't think we can return a reference to a new created election due to the lifetime complain :(

Fraser999:
Closed in favour of #129 which includes the commits from this PR (having been squashed first).

PR #120, opened 2018-09-28
Expand the add_peers test

madadam:


tests/integration_tests.rs
```diff
@@ -232,14 +234,25 @@ fn random_schedule_no_delays() {
 
 #[test]
 fn add_peers() {
+    use parsec::dev_utils::ObservationEvent::*;
+
+    let mut names = NAMES.iter();
     let mut env = Environment::new(SEED);
-    let schedule = Schedule::new(
```
pierrechevalier83: I don't know about replacing: I think it's also nice to have one that's more random to possibly catch weird edge cases in soak testing

PR #126, opened 2018-10-03
MAID-3065: functional test for unexpected genesis

madadam:


src/parsec.rs
```diff
@@ -1026,13 +1028,8 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
 
     fn restart_consensus(&mut self, latest_block_hash: &Hash) {
-        self.round_hashes = self
-            .peer_list
-            .all_ids()
-            .map(|peer_id| {
-                let round_hash = RoundHash::new(peer_id, *latest_block_hash);
-                (peer_id.clone(), vec![round_hash])
-            }).collect();
+        self.initialise_round_hashes(latest_block_hash);
```
octol: Reading the code, this is a pure refactoring right without any logic changes?

madadam: Yes. I guess that since the refactored parts have nothing to do with the actual purpose of this PR, maybe I should take them out?

src/dev_utils/dot_parser.rs
```diff
@@ -39,6 +52,17 @@ impl ParsedContents {
 
         Some(event)
     }
+
+    /// Insert event into the `ParsedContents`. Note this does not perform any
+    /// validations whatsoever, so this is useful for simulating all kinds of
```
maqi: better set the editor's line limiter to be 100 chars?

src/dump_graph.rs
```diff
@@ -167,6 +167,14 @@ mod detail {
         {
             let _ = child.wait();
         }
+
+        // Create symlink so it's easier to find the latest graphs.
+        // TODO: make it work on Windows too.
+        #[cfg(unix)]
```
maqi: I'd prefer avoid OS dependent code as much as possible unless really necessary.
Better make it work on Windows as well in this PR ?

madadam: I agree, but there is no multiplatform symlink function in the std, at the moment (actually, there is one, but it's deprecated). I can try to add a Windows block too, but I have no way to test it, as I'm not on Windows.

src/parsec.rs
octol: I kind of like having it in the same PR but as a separate commit. But this works too!

src/dev_utils/dot_parser.rs
```diff
@@ -347,6 +371,7 @@ fn parse_event_graph(contents: &str) -> BTreeMap<String, ParsedEvent> {
     parsing_events
 }
 
+#[derive(Debug)]
```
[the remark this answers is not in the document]
madadam: It's been useful to me when debugging some issues I was having. So maybe not required, but nice to have. In any case, it doesn't hurt to have it, does it?

src/parsec.rs
madadam: I usually try to put refactorings into separate commits. This one somehow slipped by me :grimacing:

src/dev_utils/dot_parser.rs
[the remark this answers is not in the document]
maqi: just try to minimize non-mandatory diffs  :)

PR #127, opened 2018-10-03
MAID-3061: Handle malice: DuplicateVote

madadam:


src/parsec.rs
```diff
@@ -1172,24 +1176,53 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+
+    // Detect that if the event carries a vote, there is already one or more
+    // votes with the same observation by the same creator.
+    fn detect_duplicate_vote(&self, event: &Event<T, S::PublicId>) -> Option<Malice> {
+        let payload = event.vote().map(|v| v.payload())?;
+
+        let mut duplicates = self
+            .peer_list
+            .peer_events(event.creator())
+            .rev()
```
Fraser999: Doesn't this mean we favour the newest events rather than the oldest as per the task spec?

madadam: No. Consider this example: an event is added to the graph. Then some time later, another event carrying the same vote is added. We detect the duplicate and raise the accusation. Then, even later another event with the same vote is added. We detect there are now three duplicates. That means that we must have already detected the duplicate when there were only two and already raised the accusation, so we won't raise it again. The accusation will always contain the oldest two events.

The `rev` there is a pure optimization (we are more likely to find the duplicates when starting from newest rather that oldest). It doesn't affect the result of the algorithm.

Fraser999: Ah, yes - you're right.  Thanks :)

src/parsec.rs
```diff
@@ -1172,24 +1176,53 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
             .map_or(true, |self_parent| !self_parent.is_initial())
+        {
+            Some(Malice::UnexpectedGenesis(*event.hash()))
+        } else {
+            None
+        }
+    }
+
+    // Detect that if the event carries a vote, there is already one or more
+    // votes with the same observation by the same creator.
```
maqi: looks like the line limit for comments is not 100 chars? the `votes` shall be able to be moved into line 1199?
just a trivial comment anyway.

madadam: Hm. I have my editor configured to show a vertical line at 80 chars, and I use it to manually wrap comments. I should probably change it to 100. Still, it would be best if rustfmt auto-wrapped comment lines too.

Anyway, probably not worth amending the PR?

Fraser999: I agree about not changing this PR, but it'd be good if you could update your editor until rustfmt does that 👍

madadam: Already did!

PR #132, opened 2018-10-08
MAID-3210: Change malice handling

madadam:
This PR introduces the following changes to malice handling:
1. Malice detection now happens before the event is added to the graph
2. Accusations are raised only after all events from the gossip message has been processed
3. Exception is the `IncorrectGenesis` malice, which raises the accusation immediately and interrupts processing of all the events in the message to prevent populating the graph with potentially large number of invalid events.

There is also a minor refactoring in a separate commit which adds type alias for `Result<T, Error>` to reduce verbosity.

### DIFF
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -10,7 +10,7 @@ use block::Block;
 #[cfg(test)]
 use dev_utils::ParsedContents;
 use dump_graph;
-use error::Error;
+use error::{Error, Result};
 use gossip::{Event, PackedEvent, Request, Response};
 use hash::Hash;
 use id::SecretId;
@@ -23,6 +23,7 @@ use peer_list::{PeerList, PeerState};
 use serialise;
 use std::collections::{BTreeMap, BTreeSet, VecDeque};
 use std::iter;
+use std::mem;
 use std::u64;
 use vote::Vote;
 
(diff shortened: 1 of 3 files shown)


src/parsec.rs
```diff
@@ -1113,60 +1120,47 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-        }
-
-        if self.detect_incorrect_genesis(event) {
-            malices.push(Malice::IncorrectGenesis(*event.hash()));
+    // Detect if the event carries an `Observation::Genesis` that doesn't match what we'd expect.
+    fn detect_incorrect_genesis(&mut self, event: &Event<T, S::PublicId>) -> Result<(), Error> {
+        if let Some(Observation::Genesis(ref group)) = event.vote().map(Vote::payload) {
+            if group.iter().collect::<BTreeSet<_>>() != self.genesis_group() {
+                // Raise the accusation immediately and return an error, to prevent accepting
+                // potentially large number of invalid / spam events into out graph.
```
Fraser999: "out" → "our"

src/parsec.rs
```diff
@@ -1113,60 +1120,47 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-        // TODO: detect other forms of malice here
-
-        malices
+        Ok(())
     }
 
     // Detect whether the event carries unexpected `Observation::Genesis`.
-    fn detect_unexpected_genesis(&self, event: &Event<T, S::PublicId>) -> Option<Malice> {
-        let payload = event.vote().map(Vote::payload)?;
+    fn detect_unexpected_genesis(&mut self, event: &Event<T, S::PublicId>) -> Result<(), Error> {
```
Fraser999: I don't think it's worth returning a `Result` from this method - no return value is needed at all really.  Same comment applies to `detect_missing_genesis()` and `detect_duplicate_vote()`.

madadam: I wanted to keep somehow uniform signature of all those `detect_*` methods. But you are probably right that it's not worth it.

PR #133, opened 2018-10-08
MAID-3058 feature/malice: detect StaleOtherParent

maqi:


src/parsec.rs
```diff
@@ -1245,6 +1249,27 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
             false
         }
     }
+
+    // Detect if the event's other_parent older than first ancestor of self_parent.
+    fn detect_stale_other_parent(&self, event: &Event<T, S::PublicId>) -> bool {
+        let (other_parent_index, other_parent_creator) = if let Some(other_parent) =
+            event.other_parent().and_then(|hash| self.events.get(hash))
```
madadam: We have a `Parsec::self_parent(&self, event: &Event) -> Option<&Event>` method already. Might be worth adding similar one for `other_parent` too and using it here, for symmetry and consistency.

Just a suggestion. I won't block the PR over this.

src/parsec.rs
```diff
@@ -1245,6 +1249,27 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+        {
+            (other_parent.index(), other_parent.creator())
+        } else {
+            return false;
+        };
+        let self_parent_ancestor_index = if let Some(index) = event
+            .self_parent()
+            .and_then(|hash| self.events.get(hash))
```
madadam: Use `self.self_parent(event)` instead of `event.self_parent().and_then(|hash| self.events.get(hash))`.

PR #134, opened 2018-10-08
MAID-3062: add test for duplicate vote handling

Fraser999:


src/gossip/event.rs
```diff
@@ -220,17 +220,18 @@ impl<T: NetworkEvent, P: PublicId> Event<T, P> {
...
-        {
-            (index, last_ancestors)
-        } else {
-            log_or_panic!(
-                "{:?} constructed an invalid event.",
-                peer_list.our_id().public_id()
-            );
-            (0, BTreeMap::new())
-        };
+        let (index, last_ancestors) =
```
maqi: any particular reason we change from `if let Ok` to `match` ?

src/parsec.rs
```diff
@@ -1230,7 +1230,7 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
 
         self.accuse(
             event.creator().clone(),
-            Malice::DuplicateVote(*event.hash(), other_hash),
+            Malice::DuplicateVote(other_hash, *event.hash()),
```
maqi: if we prefer the old hash being the first one, or the vice versa, shall we at least make the comment in Malice enum definition more explicit?

src/parsec.rs
```diff
@@ -1797,6 +1798,49 @@ mod functional_tests {
...
+        let carols_valid_vote_hash =
+            *unwrap!(find_event_by_short_name(alice.events.values(), "C_7")).hash();
+        unwrap!(alice.add_event(first_duplicate_clone));
+        let expected_accusations = vec![(
+            carol.our_pub_id().clone(),
+            Malice::DuplicateVote(carols_valid_vote_hash, first_duplicate_hash),
+        )];
+        assert_eq!(alice.pending_accusations, expected_accusations);
+
+        // Check that the second one doesn't trigger any further accusation.
```
maqi: in case of there is no accusation, shall we also check whether the duplicate event got inserted into graph?
and in above, where accusation raised, the same check of whether the duplicate event is inserted or not?

src/gossip/event.rs
Fraser999: Just so that I could print the error type in the failure branch.

src/parsec.rs
Fraser999: This was done to make it match the task description (the test exposed the difference).  I don't think there's any great justification for preferring the old hash first, so I don't really see a need for a comment (all I could put there would be something like "old hash first because we decided arbitrarily that way").  Happy to add a comment if there's more to it than that though?

src/parsec.rs
Fraser999: Fair points - the duplicates should both be inserted.  I was thinking `unwrap!` on `add_event()` to check that would be enough, but your suggestion is more robust - I'll add explicit checks.

PR #135, opened 2018-10-09
MAID-3046 feature/malice: detect fork

maqi:


src/parsec.rs
```diff
@@ -1259,6 +1259,29 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         }
     }
 
+    // Detect whether the event incurs a fork.
+    fn detect_fork(&mut self, event: &Event<T, S::PublicId>) {
```
Fraser999: I don't think this is detecting a fork (i.e. two events with the same `self_parent`).  Not sure if there's a more efficient way to do it, but essentially we want to iterate all `event.creator()`s events, checking that none have `self_parent() == self_parent_hash`.

src/parsec.rs
```diff
@@ -1259,6 +1259,29 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+        let (self_parent_index, self_parent_hash) =
+            if let Some(self_parent) = self.self_parent(event) {
+                (self_parent.index(), *self_parent.hash())
+            } else {
+                return;
+            };
+        let latest_peer_index = if let Some(latest_peer_event) = self
+            .peer_list
+            .last_event_hash(event.creator())
+            .and_then(|hash| self.events.get(hash))
```
Fraser999: We should use `self.get_known_event(hash)` for things like this, so we get the benefit of `log_or_panic!`.  It'll probably not apply for this exact code, but I'd expect we'll use it when checking each of the event creator's events' self-parents.

madadam:
I think this PR might be a bit premature, as Parsec currently cannot handle forks. In particular, the per-peer events are stored in a map keyed by event index, but forked events would have the same index, so only one would be actually stored there. Not sure how much of a problem that is, if any or whether there are other fork-related issues. Anyway, we should probably write a simple test case first, which would try to create a graph with forked events and verify that the events are actually stored in the graph. This may or might not require modification to the dot-parser.

src/parsec.rs
Fraser999: OK - after a discussion, I do believe this _is_ detecting a fork given that we enter the events into the peer_list in the right order.  We could make it somewhat more efficient by just expecting the new event, if it has a self-parent hash to be the same as out latest hash for that peer in the peer_list.

src/parsec.rs
```diff
@@ -1259,6 +1259,24 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         }
     }
 
+    // Detect whether the event incurs a fork.
+    fn detect_fork(&mut self, event: &Event<T, S::PublicId>) {
+        let self_parent_hash = if let Some(self_parent) = self.self_parent(event) {
```
Fraser999: This whole method could be simplified to:
```rust
if self.peer_list.last_event_hash(event.creator()) != event.self_parent() {
    if let Some(self_parent_hash) = event.self_parent() {
        self.accuse(event.creator().clone(), Malice::Fork(*self_parent_hash));
    }
}
```

PR #136, opened 2018-10-09
MAID-3059: add test for stale other-parent

Fraser999:


src/parsec.rs
```diff
@@ -1845,34 +1875,4 @@ mod functional_tests {
...
-    fn add_genesis_group<'a, S, I>(peer_list: &mut PeerList<S>, genesis: I)
-    where
-        S: SecretId,
-        S::PublicId: 'a,
-        I: IntoIterator<Item = &'a S::PublicId>,
-    {
-        for peer_id in genesis {
-            peer_list.add_peer(peer_id.clone(), PeerState::active());
-        }
-    }
```
octol: I was actually wondering what our convention is for helpers like this. I think I prefer at the top but having them at the bottom I guess makes the code read from top-to-bottom like Clean Code advocates?

Fraser999: There's no official guideline afaik, but we do tend to group test helpers at the start of the `tests` module.  In this case, I thought we'd all also benefit since there are still quite a few more test cases to be added to this file, and I think it's generally easier to find your new test if it's right at the bottom of the module.

src/dev_utils/dot_parser.rs
```diff
@@ -393,17 +393,13 @@ impl ParsedMetaVotes {
...
-    let mut parsing_mvs: BTreeMap<String, ParsedMetaVotes> = BTreeMap::new();
-    let mut event_name = String::default();
+    let mut parsing_mvs = BTreeMap::new();
+    let mut event_name = String::new();
     let mut peer_id = PeerId::new("");
-    for line in mvs_lines {
+    for line in contents
+        .lines()
+        .skip_while(|&line| line != "/// meta-vote section")
+    {
```
madadam: I like this. Much cleaner than before. But I would maybe `trim` the line before comparing it, in case there is some stray whitespace.

src/parsec.rs
```diff
@@ -1391,6 +1391,36 @@ mod functional_tests {
         };
     }
 
+    // Returns iterator over all votes cast by the given node.
+    fn our_votes<T: NetworkEvent, S: SecretId>(
```
madadam: I think I prefer the helpers at the end, so the more important stuff (the tests themselves) is at the top. But it's a minor issue which I can live with, so I won't block the PR over this.

PR #137, opened 2018-10-09
MAID-3071: Add test for MissingGenesis

octol:
MissingGenesis malice is when a node that is part of the genesis group creates an event with no Observation::Genesis immediately after its initial event.

### DIFF
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -1838,6 +1838,76 @@ mod functional_tests {
     }
 
     #[test]
+    fn handle_malice_missing_genesis_event() {
+        let alice_id = PeerId::new("Alice");
+        let dave_id = PeerId::new("Dave");
+
+        // Create Alice where the first event is not a genesis event (malice)
+        let mut alice_contents = ParsedContents::new(alice_id.clone());
+        alice_contents
+            .peer_list
+            .add_peer(alice_id.clone(), PeerState::active());
+        alice_contents
+            .peer_list
+            .add_peer(dave_id.clone(), PeerState::active());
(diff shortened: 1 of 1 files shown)


src/parsec.rs
```diff
@@ -1798,6 +1798,72 @@ mod functional_tests {
...
+        let d_1 = Event::<Transaction, _>::new_from_observation(
+            d_0_hash,
+            Observation::Genesis(genesis),
+            &dave_contents.events,
+            &dave_contents.peer_list,
+        );
+        dave_contents.add_event(d_1);
+        let mut dave = Parsec::from_parsed_contents(dave_contents);
+
+        // Send sync event from Alice to Dave
```
madadam: Minor, but we are not sending the sync event. The sync event is created by the recipient after processing all the events in the gossip message. So the comment should be something like "send gossip from Alice to Dave".

octol: Ah yes good point! I'll update

src/parsec.rs
```diff
@@ -1807,6 +1807,72 @@ mod functional_tests {
...
+                .filter_map(|payload| match payload {
+                    Observation::Accusation {
+                        ref offender,
+                        malice: Malice::MissingGenesis(hash),
+                    } => Some((offender, hash)),
+                    _ => None,
+                }).next()
+        );
+        assert_eq!(*offender, alice_id);
+        assert_eq!(*hash, a_1_hash);
```
maqi: Do we also need to verify the events (a_0 and a_1) reveived via gossip have been added into receiver's graph?

octol: Added in the latest commit, thanks :)

PR #142, opened 2018-10-12
MAID-3219: Fix issues with per-peer membership list

madadam:
The fixes:

1. The decided payload was incorrectly computed when deciding old meta-elections, because the payload computation relies on `interesting_events` and `interesting_content`, but those properties are not per-meta-election, and are reset when the current meta-election is decided. The fix is to put the decided payload into the meta-election itself when it is decided for the first time, then reuse on subsequent decidings.

2. Similar to the above issue, but caused by the number of voters being different between when the meta-election is decided for the first time and for the subsequent times. Happened when deciding observations that change the number of voters (`AddPeer`, ...). Fix is similar to point 1 - the number of voters is stored in the meta-election.

Other changes:

1. Keep our membership list empty - populating it would be redundant as our peers are already stored in `PeerList`.
2. Do not put a peer into their own membership list. This would again be redundant as everybody implicitly knows about themselves.

WIP. Requires tests first.

### DIFF
--- a/src/meta_voting/meta_elections.rs
+++ b/src/meta_voting/meta_elections.rs
@@ -9,9 +9,12 @@
 use super::{MetaVote, MetaVotes};
 use hash::Hash;
 use id::PublicId;
+use network_event::NetworkEvent;
+use observation::Observation;
 use round_hash::RoundHash;
 use std::collections::btree_map::Entry;
 use std::collections::{BTreeMap, BTreeSet};
+use std::fmt::{self, Debug};
 use std::{iter, mem, usize};
 
 /// Handle that uniquely identifies a `MetaElection`.
@@ -23,16 +26,29 @@ impl MetaElectionHandle {
     pub const CURRENT: Self = MetaElectionHandle(usize::MAX);
 }
 
-struct MetaElection<P: PublicId> {
+impl Debug for MetaElectionHandle {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "MetaElectionHandle");
+
+        if *self == Self::CURRENT {
+            write!(f, "::CURRENT")
+        } else {
+            write!(f, "({})", self.0)
+        }
+    }
+}
+
+struct MetaElection<T: NetworkEvent, P: PublicId> {
     meta_votes: MetaVotes<P>,
     // The "round hash" for each set of meta votes.  They are held in sequence in the `Vec`, i.e.
     // the one for round `x` is held at index `x`.
     round_hashes: BTreeMap<P, Vec<RoundHash>>,
     // Set of peers who haven't decided this election yet.
     undecided_peers: BTreeSet<P>,
+    outcome: Option<Outcome<T, P>>,
 }
 
-impl<P: PublicId> MetaElection<P> {
+impl<T: NetworkEvent, P: PublicId> MetaElection<T, P> {
     fn new<'a, I>(voters: I) -> Self
     where
         I: IntoIterator<Item = &'a P>,
@@ -42,6 +58,7 @@ impl<P: PublicId> MetaElection<P> {
             meta_votes: BTreeMap::new(),
             round_hashes: BTreeMap::new(),
             undecided_peers: voters.into_iter().cloned().collect(),
+            outcome: None,
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/peer_list.rs
```diff
@@ -135,6 +144,28 @@ impl<S: SecretId> PeerList<S> {
         }
     }
 
+    /// Add `other_peer_id` to `peer_id`'s membership list.
+    pub fn add_peers_peer(&mut self, peer_id: &S::PublicId, other_peer_id: S::PublicId) {
+        if *peer_id == other_peer_id {
```
Fraser999: These two `if`s could be combined with `||`.

src/peer_list.rs
```diff
@@ -135,6 +144,28 @@ impl<S: SecretId> PeerList<S> {
         }
     }
 
+    /// Add `other_peer_id` to `peer_id`'s membership list.
```
Fraser999: I think the comment should briefly mention when this is a no-op.

src/parsec.rs
```diff
@@ -532,21 +531,26 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
         );
 
-        info!(
-            "{:?} got consensus on block {} with payload {:?} and payload hash {:?}",
-            self.our_pub_id(),
-            self.meta_elections.consensus_history().len(),
-            payload,
-            payload_hash
-        );
+        let payload = self.consensused_blocks.back().map(|block| block.payload());
```
Fraser999: `|block| block.payload()` → `Block::payload`?

src/meta_voting/meta_elections.rs
```diff
@@ -108,21 +139,47 @@ impl<P: PublicId> MetaElections<P> {
         self.get(handle).and_then(|e| e.round_hashes.get(peer_id))
     }
 
+    /// Payload decided by the given meta-election, if any.
+    pub fn decided_payload(&self, handle: MetaElectionHandle) -> Option<&Observation<T, P>> {
+        self.get(handle)
```
Fraser999: I wonder if we should change `get()` and `get_mut()` to internally do the sort of check where we `log_or_panic!()` if the election doesn't exist?

src/meta_voting/meta_elections.rs
```diff
@@ -91,9 +116,15 @@ impl<P: PublicId> MetaElections<P> {
...
-        self.decided.keys().cloned()
+    /// Elections that were already decided by us, but not by the given peer.
+    pub fn undecided_by<'a, 'b: 'a, 'c: 'a>(
+        &'b self,
+        peer_id: &'c P,
+    ) -> impl Iterator<Item = MetaElectionHandle> + 'a {
+        self.decided
+            .iter()
+            .filter(move |(_, election)| election.undecided_peers.contains(peer_id))
+            .map(|(handle, _)| *handle)
```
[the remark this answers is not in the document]
Fraser999: Nope - I'm fine with it this way if it's cleaner.  (I expect there's little or no difference performance-wise too).

src/meta_voting/meta_elections.rs
```diff
@@ -23,16 +26,29 @@ impl MetaElectionHandle {
...
+}
+
+struct MetaElection<T: NetworkEvent, P: PublicId> {
     meta_votes: MetaVotes<P>,
     // The "round hash" for each set of meta votes.  They are held in sequence in the `Vec`, i.e.
     // the one for round `x` is held at index `x`.
     round_hashes: BTreeMap<P, Vec<RoundHash>>,
     // Set of peers who haven't decided this election yet.
     undecided_peers: BTreeSet<P>,
+    decided: Option<Decided<T, P>>,
```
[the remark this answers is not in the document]
madadam: `outcome` sounds good. Not sure it's worth to name the type `MetaElectionOutcode` as opposed to just `Outcode` as it's never exposed outside of its module?

Also, there have been some doubts raised about the name `decided` in `MetaElections::decided` because it's apparently confusing, which I sort of agree with. I'm thinking of renaming it back to `old` which it was called before my PR. Thoughts?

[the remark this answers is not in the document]
pierrechevalier83: Ok for simply `Outcome` for the type.

`old` could work, and I agree it is better. Maybe `previous_elections` is even clearer? (I think sometimes it's OK to be a bit more repetitive than necessary if it helps understanding. For instance, the name: `history` in `MetaElections` could as well be used to describe `previous_elections`... Maybe we should rename it too (`consensused_blocks` ?).

src/meta_voting/meta_elections.rs
```diff
@@ -87,10 +87,10 @@ struct Decided<T: NetworkEvent, P: PublicId> {
 pub(crate) struct MetaElections<T: NetworkEvent, P: PublicId> {
     // Current ongoing meta-election.
     current: MetaElection<T, P>,
-    // Meta-elections decided by us, but not by all peers.
-    decided: BTreeMap<MetaElectionHandle, MetaElection<T, P>>,
+    // Meta-elections that are already decided by us, but not by all the other peers.
+    previous: BTreeMap<MetaElectionHandle, MetaElection<T, P>>,
```
pierrechevalier83: I feel like a massive PITA, but I still have the objection that this lacks descriptivity: `previous_elections` would go a long way.
For instance if I'm reading line 106: `previous: BTreeMap::new()`, you force me to go lookup this line to know what `previous` represents. Could also be called `elections_history` for consistency, I guess. My main point is to avoid forcing coders to lookup the type every time they encounter the variable.

PR #145, opened 2018-10-15
MAID-3214: initialise new peer membership list

madadam:
This PR attempts to fix an existing issue where a peer added via consensus on `Observation::Add` would not have their membership list properly initialized (see https://maidsafe.atlassian.net/browse/MAID-3214 for more detailed description of why this happens). The fix is to wait for the first sync-event by the new peer and then fetch the creator of the other-parent of that sync-event and use their historic membership list at the time the other-parent was created. To achieve this, we keep track of all the changes to the membership lists so we can reconstruct them at any given point in time.

- Why can we copy the other-parent's creator's membership list? Because the new peer (let's call him Eric) is guaranteed to know all the peers the other-parent's creator (let's call him Dave) knew at the time David created the other-parent.

- Why do we have to use historic list (as opposed to the current one)? Because at the time we receive Eric's sync-event, we might have already received events from Dave's that change his membership list, but Eric does now yet know about them.

This PR also includes some refactorings.

### DIFF
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -722,13 +722,13 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         match *payload {
             Observation::Add(ref other_peer_id) => self
                 .peer_list
-                .add_peers_peer(peer_id, other_peer_id.clone()),
+                .add_to_membership_list(peer_id, other_peer_id.clone()),
             Observation::Remove(ref other_peer_id) => self
                 .peer_list
-                .remove_peers_peer(peer_id, other_peer_id.clone()),
-            Observation::Accusation { ref offender, .. } => {
-                self.peer_list.remove_peers_peer(peer_id, offender.clone())
-            }
+                .remove_from_membership_list(peer_id, other_peer_id.clone()),
+            Observation::Accusation { ref offender, .. } => self
+                .peer_list
+                .remove_from_membership_list(peer_id, offender.clone()),
             _ => (),
         }
     }
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -82,7 +82,10 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         for peer_id in genesis_group {
             parsec
                 .peer_list
-                .add_peer(peer_id.clone(), PeerState::active(), genesis_group);
+                .add_peer(peer_id.clone(), PeerState::active());
+            parsec
+                .peer_list
+                .initialise_peer_membership_list(peer_id, genesis_group.iter().cloned())
         }
 
         parsec
@@ -146,26 +149,35 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         let our_public_id = our_id.public_id().clone();
         let mut parsec = Self::empty(our_id, genesis_group, is_interesting_event);
 
+        // Add ourselves.
         parsec
             .peer_list
-            .add_peer(our_public_id, PeerState::RECV, genesis_group);
+            .add_peer(our_public_id.clone(), PeerState::RECV);
 
+        // Add the genesis group.
         for peer_id in genesis_group {
-            parsec.peer_list.add_peer(
-                peer_id.clone(),
-                PeerState::VOTE | PeerState::SEND,
-                genesis_group,
-            )
+            parsec
+                .peer_list
+                .add_peer(peer_id.clone(), PeerState::VOTE | PeerState::SEND)
         }
 
+        // Add the current section members.
         for peer_id in section {
             if genesis_group.contains(peer_id) {
                 continue;
             }
 
+            parsec.peer_list.add_peer(peer_id.clone(), PeerState::SEND)
+        }
+
+        // Initialise everyone's membership list.
+        for peer_id in iter::once(&our_public_id)
+            .chain(genesis_group)
+            .chain(section)
+        {
             parsec
                 .peer_list
-                .add_peer(peer_id.clone(), PeerState::SEND, genesis_group)
+                .initialise_peer_membership_list(peer_id, genesis_group.iter().cloned())
         }
 
         parsec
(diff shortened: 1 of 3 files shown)


src/peer_list.rs
```diff
@@ -423,10 +421,69 @@ impl<P: PublicId> Peer<P> {
...
+    /// Returns the current membership list.
+    #[allow(unused)]
+    pub fn peers(&self) -> &BTreeSet<P> {
+        &self.peers
+    }
+
+    /// Returns the historic membership list at the time the event at `index` was the last event of
+    /// this peer.
+    #[allow(unused)]
+    pub fn peers_at(&self, event_index: u64) -> BTreeSet<P> {
```
fizyk20: I'm a bit worried if this will work in the presence of forks - an index doesn't identify an event uniquely, then. It might me much more complex to do this with hashes, though...

src/meta_voting/meta_elections.rs
```diff
@@ -59,18 +76,26 @@ impl<P: PublicId> MetaElection<P> {
     }
 }
 
-pub(crate) struct MetaElections<P: PublicId> {
+// Info carried by a meta-election decided by us.
+struct Decided<T: NetworkEvent, P: PublicId> {
+    // Payload decided by this election
+    payload: Observation<T, P>,
+    // Number of voters at the time this election was decided.
+    voter_count: usize,
```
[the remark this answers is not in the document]
pierrechevalier83: I was actually wondering the same thing when reviewing PR 142, but assumed it must be right (well, we'll know if the tests pass :smile: ) Happy for Adam to comment though :wink:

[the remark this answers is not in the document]
madadam: Do you mean as opposed to storing the full set of voters? For some reason I thought that the count should be enough, but I'm not so convinced anymore :grimacing: I'll try to spend some more time thinking about it.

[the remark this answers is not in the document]
madadam: OK, after discussion, this issue is going to be addressed in a separate PR.

src/parsec.rs
```diff
@@ -649,13 +649,13 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         match *payload {
             Observation::Add(ref other_peer_id) => self
                 .peer_list
-                .add_peers_peer(peer_id, other_peer_id.clone()),
+                .add_to_membership_list(peer_id, other_peer_id.clone()),
```
pierrechevalier83: I would prefer `add_to_peers_membership_list`. It makes it more obvious what both arguments mean. As is, you could think you're adding to peers to some membership list.

pierrechevalier83:
```suggestion
                .add_to_peers_membership_list(peer_id, other_peer_id.clone()),
```
It would make it more obvious what both arguments mean. As is, you could think you're adding two peers to some membership list.

src/parsec.rs
```diff
@@ -649,13 +649,13 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
             Observation::Remove(ref other_peer_id) => self
                 .peer_list
-                .remove_peers_peer(peer_id, other_peer_id.clone()),
-            Observation::Accusation { ref offender, .. } => {
-                self.peer_list.remove_peers_peer(peer_id, offender.clone())
-            }
+                .remove_from_membership_list(peer_id, other_peer_id.clone()),
```
pierrechevalier83: Same as above: I would prefer `remove_from_peers_membership_list`

src/peer_list.rs
```diff
@@ -169,6 +169,50 @@ impl<S: SecretId> PeerList<S> {
         }
     }
 
+    /// Initialise the membership list of `peer_id` with the given `other_peer_ids`.
+    pub fn initialise_membership_list<I>(&mut self, peer_id: &S::PublicId, other_peer_ids: I)
```
pierrechevalier83: Would also call this `initialise_peers_membership_list` for consistency if you rename both others.

src/parsec.rs
```diff
@@ -766,10 +766,40 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-    // Initialise the membership list of the creator of the given event.
-    // Do nothing if already initialised.
-    fn initialise_membership_list(&mut self, _event_hash: &Hash) {
-        // TODO
+    // Initialise the membership list of the creator of the given event to the same membership list
+    // the creator of the other-parent had at the time of the other-parent's creation. Do nothing if
+    // the event is not a request or response or if the membership list if already initialised.
+    fn initialise_membership_list(&mut self, event_hash: &Hash) {
+        let (creator, other_parent_creator, other_parent_index) = {
+            let event = if let Some(event) = self.get_known_event(event_hash).ok() {
```
pierrechevalier83: Do we really want to swallow this error?

src/peer_list.rs
```diff
@@ -169,6 +169,50 @@ impl<S: SecretId> PeerList<S> {
...
+        // Do not populate our membership list, as it would be redundant.
+        if *peer_id == *self.our_id.public_id() {
+            return;
+        }
+
+        if let Some(peer) = self.peers.get_mut(peer_id) {
+            for other_peer_id in other_peer_ids {
+                // Do not add the peer into their own membership list, as it too would be
+                // redundant.
+                if other_peer_id == *peer_id {
```
pierrechevalier83: Nitpick: this would feel more idiomatic with a filter (especially because other_peers_ids is already an iterator)

src/peer_list.rs
```diff
@@ -169,6 +169,50 @@ impl<S: SecretId> PeerList<S> {
...
+    pub fn initialise_membership_list<I>(&mut self, peer_id: &S::PublicId, other_peer_ids: I)
+    where
+        I: IntoIterator<Item = S::PublicId>,
+    {
+        // Do not populate our membership list, as it would be redundant.
+        if *peer_id == *self.our_id.public_id() {
+            return;
+        }
+
+        if let Some(peer) = self.peers.get_mut(peer_id) {
```
pierrechevalier83: `else`, shouldn't we return an error? (To make this function harder to misuse without noticing)

PR #146, opened 2018-10-16
MAID-3212: handle multiple meta-elections and peers' peer lists

Fraser999:


Fraser999:
Will be replaced by a subsequent PR to tackle this task.

PR #147, opened 2018-10-16
MAID-3222 and MAID-3223: Expand the public API to support routing integration

madadam:
- add method to check whether there are any not yet consensused observations in the graph
- add method to retrieve all unpolled observations created by us

### DIFF
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -53,6 +53,8 @@ pub struct Parsec<T: NetworkEvent, S: SecretId> {
     events: BTreeMap<Hash, Event<T, S::PublicId>>,
     // The sequence in which all gossip events were added to this `Parsec`.
     events_order: Vec<Hash>,
+    // Information about observations stored in the graph, mapped to their hashes.
+    observations: BTreeMap<Hash, ObservationInfo>,
     // The hashes of events for each peer that have a non-empty set of `interesting_content`.
     interesting_events: BTreeMap<S::PublicId, VecDeque<Hash>>,
     // Consensused network events that have not been returned via `poll()` yet.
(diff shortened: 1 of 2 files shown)


src/parsec.rs
```diff
@@ -513,6 +559,11 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
 
             let block = self.create_block(payload.clone())?;
             self.consensused_blocks.push_back(block);
+
```
pierrechevalier83: it seems `cargo fmt` didn't like this line jump

src/parsec.rs
```diff
@@ -320,6 +323,43 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+    }
+
+    fn our_consensused_observations(&self) -> impl Iterator<Item = &Observation<T, S::PublicId>> {
+        self.consensused_blocks
+            .iter()
+            .filter(move |block| {
+                block
+                    .proofs()
+                    .iter()
+                    .any(|proof| proof.public_id() == self.our_pub_id())
```
pierrechevalier83: So `our_consensused_observations` only returns consensused observations of which we are a signatory.

I think this is more restrictive than it should be. The question is have I voted for this payload, which is different from did my vote for this payload get considered when creating this block?

src/parsec.rs
```diff
@@ -513,6 +568,13 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
 
             let block = self.create_block(payload.clone())?;
             self.consensused_blocks.push_back(block);
+
+            if let Some(hash) = self.meta_elections.consensus_history().last() {
+                if let Some(info) = self.observations.get_mut(hash) {
+                    info.consensused = true;
+                }
```
maqi: shall there at least be a log_or_panic for the `else` situation here?
also, why not use the `block.create_payload_hash()`, instead of fetching hash from meta_elections?

PR #148, opened 2018-10-18
MAID-3225: meta-election data

madadam:
This PR takes all data that is used for deciding meta-election and puts them under meta-election, so when multiple meta-elections are evaluated in parallel, the don't interfere with each other. Summary of changes:

- Introduced a struct called `MetaEvent` to contain all per-event, per-meta-election data.
- Added map of `Hash` -> `MetaEvent` to `MetaElection`.
- Moved `meta_votes` to `MetaEvent`.
- Moved `observations` from `Event` to `MetaEvent`.
- Moved `interesting_content` from `Event` to `MetaEvent`.
- Moved `interesting_events` from `Parsec` to `MetaElection`.
- A full list of voters is now stored in decided `MetaElection`s, instead of just the count.

This PR also includes some refactoring.

### DIFF
--- a/src/dev_utils/dot_parser.rs
+++ b/src/dev_utils/dot_parser.rs
@@ -24,7 +24,7 @@ pub(crate) struct ParsedContents {
     pub our_id: PeerId,
     pub events: BTreeMap<Hash, Event<Transaction, PeerId>>,
     pub events_order: Vec<Hash>,
-    pub meta_events: BTreeMap<Hash, MetaEvent<PeerId>>,
+    pub meta_events: BTreeMap<Hash, MetaEvent<Transaction, PeerId>>,
     pub peer_list: PeerList<PeerId>,
 }
 
@@ -250,13 +250,13 @@ fn read(mut file: File) -> io::Result<ParsedContents> {
             other_parent,
             mv.event_index,
             event.last_ancestors,
-            event.interesting_content,
         );
 
         let hash = *parsed_event.hash();
-        if !mv.meta_votes.is_empty() {
+        if !event.interesting_content.is_empty() || !mv.meta_votes.is_empty() {
             let mut meta_event = MetaEvent::new();
             meta_event.meta_votes = mv.meta_votes;
+            meta_event.interesting_content = event.interesting_content;
 
             let _ = meta_events.insert(hash, meta_event);
         }
--- a/src/dev_utils/dot_parser.rs
+++ b/src/dev_utils/dot_parser.rs
@@ -8,7 +8,7 @@
 
 use gossip::Event;
 use hash::Hash;
-use meta_voting::{BoolSet, MetaVote, Step};
+use meta_voting::{BoolSet, MetaEvent, MetaVote, Step};
 use mock::{PeerId, Transaction};
 use observation::Observation;
 use peer_list::{PeerList, PeerState};
@@ -24,7 +24,7 @@ pub(crate) struct ParsedContents {
     pub our_id: PeerId,
     pub events: BTreeMap<Hash, Event<Transaction, PeerId>>,
     pub events_order: Vec<Hash>,
-    pub meta_votes: BTreeMap<Hash, BTreeMap<PeerId, Vec<MetaVote>>>,
+    pub meta_events: BTreeMap<Hash, MetaEvent<PeerId>>,
     pub peer_list: PeerList<PeerId>,
 }
 
@@ -37,7 +37,7 @@ impl ParsedContents {
             our_id,
             events: BTreeMap::new(),
             events_order: Vec::new(),
-            meta_votes: BTreeMap::new(),
+            meta_events: BTreeMap::new(),
             peer_list,
         }
     }
@@ -48,6 +48,7 @@ impl ParsedContents {
         let event = self.events.remove(&hash)?;
 
         self.peer_list.remove_event(&event);
+        let _ = self.meta_events.remove(&hash);
 
         Some(event)
     }
@@ -59,6 +60,7 @@ impl ParsedContents {
 
         let hash = *event.hash();
         let _ = self.events.insert(hash, event);
+        let _ = self.meta_events.insert(hash, MetaEvent::new());
         self.events_order.push(hash);
     }
 }
@@ -101,7 +103,6 @@ struct ParsedEvent {
     other_parent: Option<String>,
     interesting_content: Vec<Observation<Transaction, PeerId>>,
     last_ancestors: BTreeMap<PeerId, u64>,
-    observations: BTreeSet<PeerId>,
 }
 
 impl ParsedEvent {
@@ -112,7 +113,6 @@ impl ParsedEvent {
         other_parent: Option<String>,
         interesting_content_string: &str,
         last_ancestors_string: &str,
-        observations: BTreeSet<PeerId>,
     ) -> Self {
         let interesting_content = parse_interesting_content(interesting_content_string);
         let last_ancestors = parse_peer_entries(last_ancestors_string)
@@ -126,7 +126,6 @@ impl ParsedEvent {
             other_parent,
             interesting_content,
             last_ancestors,
-            observations,
         }
     }
 }
@@ -227,7 +226,7 @@ fn read(mut file: File) -> io::Result<ParsedContents> {
     let mut events = BTreeMap::new();
     let mut events_order = Vec::new();
     let mut name_hash_map: BTreeMap<String, Hash> = BTreeMap::new();
-    let mut meta_votes = BTreeMap::new();
+    let mut meta_events = BTreeMap::new();
     let length = parsing_events.len();
 
     while !parsing_events.is_empty() {
@@ -256,7 +255,10 @@ fn read(mut file: File) -> io::Result<ParsedContents> {
 
         let hash = *parsed_event.hash();
         if !mv.meta_votes.is_empty() {
-            let _ = meta_votes.insert(hash, mv.meta_votes);
+            let mut meta_event = MetaEvent::new();
+            meta_event.meta_votes = mv.meta_votes;
+
+            let _ = meta_events.insert(hash, meta_event);
         }
 
         let _ = events.insert(hash, parsed_event);
@@ -273,7 +275,7 @@ fn read(mut file: File) -> io::Result<ParsedContents> {
         our_id,
         events,
         events_order,
-        meta_votes,
+        meta_events,
         peer_list,
     })
 }
@@ -360,7 +362,6 @@ fn parse_event_graph(contents: &str) -> BTreeMap<String, ParsedEvent> {
                         other_parent,
                         info.interesting_content,
                         info.last_ancestors,
-                        BTreeSet::new(),
                     ),
                 );
             }
@@ -649,10 +650,10 @@ mod tests {
             assert_eq!(gossip_graph, parsed_result.events);
 
             // The dumped dot file doesn't contain all the meta_votes
-            assert!(meta_votes.len() >= parsed_result.meta_votes.len());
-            for (hash, meta_vote) in &parsed_result.meta_votes {
+            assert!(meta_votes.len() >= parsed_result.meta_events.len());
+            for (hash, meta_event) in &parsed_result.meta_events {
                 let ori_meta_vote = unwrap!(meta_votes.get(hash));
-                assert_eq!(ori_meta_vote, meta_vote);
+                assert_eq!(*ori_meta_vote, meta_event.meta_votes);
             }
         }
         assert_ne!(num_of_files, 0u8);
--- a/src/dev_utils/dot_parser.rs
+++ b/src/dev_utils/dot_parser.rs
@@ -646,9 +646,6 @@ mod tests {
             let parsed_result = unwrap!(parse_dot_file(&dot_file_path));
 
             assert_eq!(gossip_graph.len(), parsed_result.events_order.len());
-            for event in &mut gossip_graph.values_mut() {
-                event.observations.clear();
-            }
             assert_eq!(gossip_graph, parsed_result.events);
 
             // The dumped dot file doesn't contain all the meta_votes
--- a/src/dev_utils/dot_parser.rs
+++ b/src/dev_utils/dot_parser.rs
@@ -655,6 +655,10 @@ mod tests {
             assert_eq!(gossip_graph, parsed_result.events);
 
             // The dumped dot file doesn't contain all the meta_votes
+            // NOTE: We only serialise meta-votes in the core file, not complete meta-events.
+            // This is why we compare the number of meta-votes (from the core) to the number of
+            // meta-events (in the parsed result) here. There is, however, one-to-one mapping
+            // between meta-events and sets of meta-votes, so this is correct.
             assert!(meta_votes.len() >= parsed_result.meta_events.len());
             for (hash, meta_event) in &parsed_result.meta_events {
                 let ori_meta_vote = unwrap!(meta_votes.get(hash));
--- a/src/dev_utils/dot_parser.rs
+++ b/src/dev_utils/dot_parser.rs
@@ -8,7 +8,7 @@
 
 use gossip::Event;
 use hash::Hash;
-use meta_voting::{BoolSet, MetaEvent, MetaVote, Step};
+use meta_voting::{BoolSet, MetaElectionHandle, MetaEvent, MetaVote, Step};
 use mock::{PeerId, Transaction};
 use observation::Observation;
 use peer_list::{PeerList, PeerState};
@@ -59,8 +59,10 @@ impl ParsedContents {
         unwrap!(self.peer_list.add_event(&event));
 
         let hash = *event.hash();
+        let meta_event = MetaEvent::build(MetaElectionHandle::CURRENT, &event).finish();
+
         let _ = self.events.insert(hash, event);
-        let _ = self.meta_events.insert(hash, MetaEvent::new());
+        let _ = self.meta_events.insert(hash, meta_event);
         self.events_order.push(hash);
     }
 }
@@ -254,9 +256,12 @@ fn read(mut file: File) -> io::Result<ParsedContents> {
 
         let hash = *parsed_event.hash();
         if !event.interesting_content.is_empty() || !mv.meta_votes.is_empty() {
-            let mut meta_event = MetaEvent::new();
-            meta_event.meta_votes = mv.meta_votes;
-            meta_event.interesting_content = event.interesting_content;
+            let meta_event = {
+                let mut builder = MetaEvent::build(MetaElectionHandle::CURRENT, &parsed_event);
+                builder.set_interesting_content(event.interesting_content);
+                builder.set_meta_votes(mv.meta_votes);
+                builder.finish()
+            };
 
             let _ = meta_events.insert(hash, meta_event);
         }
--- a/src/gossip/event.rs
+++ b/src/gossip/event.rs
@@ -36,8 +36,6 @@ pub(crate) struct Event<T: NetworkEvent, P: PublicId> {
     index: u64,
     // Index of each peer's latest event that is an ancestor of this event.
     last_ancestors: BTreeMap<P, u64>,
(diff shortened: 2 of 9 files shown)


src/parsec.rs
```diff
@@ -459,20 +445,57 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+            );
+
+            false
+        }
+    }
+
+    fn has_supermajority_observations(
+        &self,
+        election: MetaElectionHandle,
+        meta_event: &MetaEvent<T, S::PublicId>,
```
fizyk20: Wouldn't it make more sense for this function to still take an event, and delegate getting the corresponding meta-event to the MetaElections struct?

src/parsec.rs
```diff
@@ -459,20 +445,57 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-    fn has_supermajority_observations(&self, event: &Event<T, S::PublicId>) -> bool {
-        self.peer_list.is_super_majority(event.observations.len())
-    }
-
-    fn is_observer(&self, event: &Event<T, S::PublicId>) -> bool {
+    fn is_observer(
+        &self,
+        election: MetaElectionHandle,
+        event: &Event<T, S::PublicId>,
+        meta_event: &MetaEvent<T, S::PublicId>,
```
fizyk20: We shouldn't need to pass the meta-event here - it should be determined by the election handle and event itself. As it is currently, we could pass an event with a non-matching meta-event and shoot ourselves in the foot. Or am I missing something?

src/parsec.rs
```diff
@@ -21,12 +21,10 @@ use network_event::NetworkEvent;
...
 use peer_list::{PeerList, PeerState};
 use std::collections::{BTreeMap, BTreeSet, VecDeque};
-use std::iter;
-use std::mem;
-use std::u64;
+use std::{iter, mem, u64};
 use vote::Vote;
 
-pub type IsInterestingEventFn<P> = fn(voters: &BTreeSet<&P>, current_peers: &BTreeSet<&P>) -> bool;
+pub type IsInterestingEventFn<P> = fn(did_vote: &BTreeSet<P>, can_vote: &BTreeSet<P>) -> bool;
```
pierrechevalier83: Not sure about using verbs in variable names. Slightly unconventional and could be confusing.
What about:
`actual_voters`, `valid_voters`?

src/parsec.rs
```diff
@@ -582,25 +601,16 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-                    info.consensused = true;
-                } else {
-                    log_or_panic!(
-                        "{:?} doesn't know about observation with hash {:?}",
-                        self.peer_list.our_pub_id(),
-                        hash
-                    );
-                }
+            if let Some(info) = self.observations.get_mut(&payload_hash) {
+                info.consensused = true;
```
fizyk20: Aren't observations supposed to be stored in meta-elections after this PR? Or do we need to keep them here as well to initialise new meta-elections properly?

src/parsec.rs
```diff
@@ -795,170 +820,127 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-                .collect()
-        };
-        self.get_known_event_mut(event_hash)
-            .map(|event| event.observations = observations)
-    }
+    fn set_meta_votes(
+        &self,
+        election: MetaElectionHandle,
+        event: &Event<T, S::PublicId>,
+        meta_event: &mut MetaEvent<T, S::PublicId>,
```
fizyk20: Again, I don't think that passing a meta-event along with the event seems like a good idea, unless I'm missing something.

src/meta_voting/meta_elections.rs
```diff
@@ -28,18 +28,21 @@ impl MetaElectionHandle {
...
 
         if *self == Self::CURRENT {
-            write!(f, "::CURRENT")
+            write!(f, "CURRENT")?
         } else {
-            write!(f, "({})", self.0)
+            write!(f, "{}", self.0)?
         }
+
+        write!(f, ")")
```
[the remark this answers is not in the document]
madadam: I disagree. Yours reads marginally better, but mine doesn't require allocations. I usually prefer to avoid them if possible.

src/meta_voting/meta_elections.rs
```diff
@@ -135,6 +153,30 @@ impl<T: NetworkEvent, P: PublicId> MetaElections<T, P> {
         self.get(handle).and_then(|e| e.meta_votes.get(event_hash))
     }
 
+    pub fn add_meta_event(&mut self, handle: MetaElectionHandle, event_hash: Hash) {
+        if let Some(election) = self.get_mut(handle) {
+            let _ = election.meta_events.insert(event_hash, MetaEvent::new());
+        }
```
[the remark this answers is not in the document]
madadam: There is `log_or_panic` in `get` and `get_mut` already.

src/parsec.rs
```diff
@@ -459,20 +459,47 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-                true
-            },
-            |parent| !self.has_supermajority_observations(parent),
-        )
+
+        if !self.has_supermajority_observations(meta_event) {
+            return false;
+        }
+
+        let self_parent = if let Some(event) = self.self_parent(event) {
```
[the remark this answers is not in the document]
madadam: Ok about the shadowing, but I though early return is preferred?

src/dev_utils/dot_parser.rs
```diff
@@ -649,10 +650,10 @@ mod tests {
             assert_eq!(gossip_graph, parsed_result.events);
 
             // The dumped dot file doesn't contain all the meta_votes
-            assert!(meta_votes.len() >= parsed_result.meta_votes.len());
-            for (hash, meta_vote) in &parsed_result.meta_votes {
+            assert!(meta_votes.len() >= parsed_result.meta_events.len());
```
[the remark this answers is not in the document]
madadam: The lengths should be the same though. Previously `meta_votes` were stored per event-hash. Now they are stored in `MetaEvent` which is stored per event-hash.

src/meta_voting/meta_elections.rs
```diff
@@ -88,18 +86,25 @@ struct Outcome<T: NetworkEvent, P: PublicId> {
...
 pub(crate) struct MetaEvent<P> {
     // The set of peers for which this event can strongly-see an event by that peer which carries a
     // valid block.  If there are a supermajority of peers here, this event is an "observer".
     pub observations: BTreeSet<P>,
+    pub meta_votes: MetaVotes<P>,
 }
 
 impl<P: PublicId> MetaEvent<P> {
-    fn new() -> Self {
+    pub fn new() -> Self {
```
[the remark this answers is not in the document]
madadam: The struct itself is `pub(crate)` which should be enough to prevent accidental exposure, I think.

src/meta_voting/meta_elections.rs
```diff
@@ -145,18 +150,35 @@ impl<T: NetworkEvent, P: PublicId> MetaElections<T, P> {
...
-        event_hash: &Hash,
-    ) -> Option<&BTreeMap<P, Vec<MetaVote>>> {
-        self.get(handle).and_then(|e| e.meta_votes.get(event_hash))
-    }
+        event_hash: Hash,
+        meta_event: MetaEvent<P>,
+    ) {
+        let election = if let Some(election) = self.get_mut(handle) {
+            election
+        } else {
```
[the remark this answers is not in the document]
madadam: See https://github.com/maidsafe/parsec/pull/148#discussion_r226551369

src/parsec.rs
madadam: The reason I changed it is that we use the term `voters` to mean all peers that *can* vote. So it would be very confusing if we use `voters` to suddenly mean those that *did* vote. Your suggestion is slightly better than how it used to be, but I think it's still confusing.

If the verbs bother you, what about: `peers_that_can_vote` and `peers_that_did_vote`. It's verbose, but should be clearer.

src/gossip/event.rs
```diff
@@ -295,16 +285,8 @@ impl<T: NetworkEvent, P: PublicId> Event<T, P> {
...
-        writeln!(writer, "/// cause: {}", self.content.cause)?;
-        writeln!(
-            writer,
-            "/// interesting_content: {:?}",
-            self.interesting_content
-        )?;
-        writeln!(writer, "/// last_ancestors: {:?}", self.last_ancestors)?;
-        writeln!(writer, "/// }}")
+    pub fn write_cause_to_dot_format(&self, writer: &mut Write) -> io::Result<()> {
+        writeln!(writer, "/// cause: {}", self.content.cause)
```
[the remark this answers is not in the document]
madadam: That's been moved to `dump_graph::write_event_to_dot`. Id move the `cause` there as well, but that would require exposing `Cause` outside of the `gossip` module which some might not like.

src/parsec.rs
madadam: Hm. We have a name overload problem I think. There are two types of "observations":
1. In `Parsec::observations` which is a kind of cache for all `Observation`s we have and was added in a different PR (to implement the API routing-integration team requested).
2. In `MetaEvent::observations` which is a set of peer ids of creators of interesting events strongly seen by that event.

I think we should rename one of them to something else to prevent confusion. I'm open to suggestion what the name should be.

src/meta_voting/meta_elections.rs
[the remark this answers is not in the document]
pierrechevalier83: Fair enough. I generally favour readability over performance until optimisation time, but this is such a minor style comment that I'm happy for you to disagree if you prefer your original style. Feel free to ignore.

src/meta_voting/meta_elections.rs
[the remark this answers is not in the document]
pierrechevalier83: uhm. ok.

src/dev_utils/dot_parser.rs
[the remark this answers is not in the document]
pierrechevalier83: uhm. Makes sense. Could you add one line of comment to explain this? At a glance, it feels like comparing 2 things of different cardinality, so my internal static analyser gives a warning :smile:

src/meta_voting/meta_elections.rs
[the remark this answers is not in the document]
pierrechevalier83: Interesting. Not sure what's the convention for methods of a crate that's `pub (crate)`. I would still be for making this `pub (crate)` so that this line as a standalone is clearly "not so public"

src/parsec.rs
```diff
@@ -919,81 +929,82 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-                }).cloned()
-                .collect()
-        };
-
-        if let Some(event) = self.meta_elections.meta_event_mut(election, event_hash) {
-            event.observations = observations;
-        }
+    fn set_observations(
+        &self,
+        _election: MetaElectionHandle,
```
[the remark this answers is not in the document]
pierrechevalier83: It did become used one commit later, btw

src/parsec.rs
pierrechevalier83: `peers_that_can_vote` and `peers_that_did_vote` sounds good to me. In doubt, better to err on the side of extra clarity/verbosity imho :wink:

src/parsec.rs
pierrechevalier83: suggestion:

2. `MetaEvent::strongly_seen_interesting_event_creators`

Verbose but clear.

Better suggestions welcome :)

src/meta_voting/meta_event.rs
```diff
@@ -23,15 +25,60 @@ pub(crate) struct MetaEvent<T: NetworkEvent, P: PublicId> {
...
+            meta_event: MetaEvent {
+                observations: BTreeSet::new(),
+                interesting_content: Vec::new(),
+                meta_votes: MetaVotes::new(),
+            },
         }
     }
+}
+
+pub(crate) struct MetaEventBuilder<'a, T: NetworkEvent + 'a, P: PublicId + 'a> {
```
pierrechevalier83: Not sure about the name (except if it's the widely accepted meaning of "Builder" in the OO/ design pattern world and I'm showing my lack of awareness of that world), but I like the general idea of making this "coupling" explicit.

src/meta_voting/meta_elections.rs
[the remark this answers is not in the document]
octol: Anyone know how good the optimizer is for these things, will it transform away the allocation (in Perre's version)? I'm guessing not?

src/meta_voting/meta_elections.rs
```diff
@@ -146,17 +240,69 @@ impl<T: NetworkEvent, P: PublicId> MetaElections<T, P> {
...
+    pub fn is_interesting_content_candidate(
+        &self,
+        handle: MetaElectionHandle,
+        creator: &P,
+        payload: &Observation<T, P>,
+    ) -> bool {
+        let election = if let Some(election) = self.get(handle) {
+            election
+        } else {
+            return false;
```
maqi: shall we have a log_or_panic here?

src/meta_voting/meta_elections.rs
```diff
@@ -146,17 +240,69 @@ impl<T: NetworkEvent, P: PublicId> MetaElections<T, P> {
...
+
+        // Already interesting?
+        if election.is_already_interesting_content(creator, payload) {
+            return false;
+        }
+
+        // Already consensused?
+        let hash = payload.create_hash();
+        if self.consensus_history()[..election.consensus_len].contains(&hash) {
+            return false;
```
maqi: I think `if ... { false } else { true}` will be preferred?

src/meta_voting/meta_elections.rs
```diff
@@ -167,19 +313,19 @@ impl<T: NetworkEvent, P: PublicId> MetaElections<T, P> {
...
     {
+        let voters: BTreeSet<_> = voters.into_iter().cloned().collect();
+
         let hash = payload.create_hash();
-        let new = MetaElection::new(voters);
-        let voter_count = new.undecided_peers.len();
+        self.consensus_history.push(hash);
+
+        let mut new = MetaElection::new(voters.clone());
+        new.consensus_len = self.consensus_history.len();
```
maqi: why not put consensus_len as a prameter of the constructor?

src/meta_voting/meta_elections.rs
madadam: See https://github.com/maidsafe/parsec/pull/148#discussion_r226551369.

src/meta_voting/meta_event.rs
madadam: I called it Builder because I think it's very similar to the "Builder pattern" that is used widely in the rust ecosystem. I'm open to suggestions for a better name though.

src/meta_voting/meta_elections.rs
```diff
@@ -127,12 +169,64 @@ impl<T: NetworkEvent, P: PublicId> MetaElections<T, P> {
...
+        let election = if let Some(election) = self.get_mut(handle) {
+            election
+        } else {
+            return;
+        };
+
+        // Update round hashes.
+        for (peer_id, event_votes) in &meta_event.meta_votes {
+            for meta_vote in event_votes {
+                let hashes = if let Some(hashes) = election.round_hashes.get_mut(&peer_id) {
```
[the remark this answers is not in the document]
madadam: :+1:

src/meta_voting/meta_elections.rs
madadam: :+1:

src/parsec.rs
madadam: What about `observees` ? (I'm only half joking :) )

pierrechevalier83: `observees` would sound like a bunch of peer to me for some reason... `observed_events` ?

src/meta_voting/meta_event.rs
pierrechevalier83: I guess where it differs from my idea of a builder is that it is sometimes passed around to functions as a container, more than as a builder...

I gather you weren't a fan of `MetaEventInContext`? (I'm trying to convey composition in the name as that's my perception of this...)

src/meta_voting/meta_elections.rs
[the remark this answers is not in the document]
pierrechevalier83: No SSO for rust apparently because of `as_mut_vec`: https://internals.rust-lang.org/t/small-string-optimization-remove-as-mut-vec/1320/36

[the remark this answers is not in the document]
pierrechevalier83: (Although my point still stands about worrying about these things after code has been shown to be in the hot path ;) )

src/parsec.rs
madadam: But it is peers. The type of `observations` is `BTreeSet<P>` (where `P: PublicId`)

src/parsec.rs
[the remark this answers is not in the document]
pierrechevalier83: > but I though early return is preferred?

This one is a bit of a matter of opinion, but I prefer no early return because it helps your function be composed of self-consistent statements (or scopes) that are good candidates for extraction into smaller function if the function grows.
With early returns, you must read the entire function to understand the flow.
Without early returns, one branch is a self-consistent entity that you can read on its own.

I make an exception for error handling because I think that the `?` operator is very clear and it is type safe so it can't make a refactoring break the behaviour.

So my rule of thumb is: the `return` keyword is a code smell.

YMMV, of course. I'm not particularly trying to convert you but just expressing my views FYI :smile: Happy for you to keep this style if you prefer it; and would also be curious if you have arguments to support it (it's fun Friday, after all :tada:)

src/parsec.rs
octol: I've always found words like 'employees' taking a bit longer to parse in my head. I always have to quickly think 'employers - employees' before it clicks. But it just might be language thingie for me...

madadam:
> @octol : Should the definition of a meta event be added to the documentation?

It's not exposed outside of the crate, so I don't think it's necessary.

src/meta_voting/meta_event.rs
madadam: But it is used like builder mostly. It's passed to `set_interesting_content`, `set_observations` and `set_meta_votes` - all of those initialise some part of the `MetaEvent` under construction. The only function where it is not used like builder is `is_observer`.

So basically I think it's much closer to the builder pattern than to just a container.

src/meta_voting/meta_elections.rs
```diff
@@ -145,18 +150,35 @@ impl<T: NetworkEvent, P: PublicId> MetaElections<T, P> {
             .map(|(handle, _)| *handle)
     }
 
-    pub fn meta_votes(
-        &self,
+    pub fn add_meta_event(
+        &mut self,
```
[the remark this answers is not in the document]
madadam: I guess I could use the `MetaEventBuilder` here, but it would hit issues with the borrow checker, because the builder borrows `Event` via `get_known_event`, so it indirectly borrows `self` which would prevent me from calling `add_meta_event` because it's `&mut self`. A way around that would be to borrow the `Event` via `events.get` instead of `get_known_event`, but then I'd probably also had to add the `log_or_panic` branch explicitly. So, I'm not sure. I might consider it if you guys think would be better that way.

src/meta_voting/meta_event.rs
madadam: That said, I'm not super against renaming it to `MetaEventInContext` or something else. I just think the name Builder is more appropriate considering how it is used.

src/meta_voting/meta_elections.rs
```diff
@@ -274,10 +275,10 @@ impl<T: NetworkEvent, P: PublicId> MetaElections<T, P> {
         // Already consensused?
         let hash = payload.create_hash();
         if self.consensus_history()[..election.consensus_len].contains(&hash) {
```
pierrechevalier83: clippy is right: this can be simplified to `!self.consensus_history()[..election.consensus_len].contains(&hash)`

PR #151, opened 2018-10-25
MAID-3233: add test for unpolled and unconsensused observations

Fraser999:
I'll add the WIP tag to this since I think it should be merged after #150 to avoid causing any more pain to that much larger PR.

It should be fairly easy to regenerate the dot file for this test once the dot-writing/parsing changes are in place.  This one was generated using the dot_gen tool with the following scenario:
```rust
let _ = scenarios
    .add(
        "parsec::functional_tests::unpolled_and_unconsensused_observations",
        |env| {
                genesis: peer_ids!("Alice", "Bob", "Carol", "Dave"),
                schedule: vec![(0, AddPeer(PeerId::new("Eric")))],
            };

            Schedule::from_observation_schedule(env, &ScheduleOptions::default(), obs)
        },
    ).file("Alice", "alice.dot")
    .seed([1, 3, 1, 4]);
```

### DIFF
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -1802,19 +1802,26 @@ impl Parsec<Transaction, PeerId> {
     pub(crate) fn from_parsed_contents(parsed_contents: ParsedContents) -> Self {
         let mut parsec = Parsec::empty(parsed_contents.our_id, &BTreeSet::new(), is_supermajority);
 
-        // Populate the `observations` cache using the payloads carried by events...
-        for event in parsed_contents.events.values() {
-            if let Some(payload) = event.vote().map(Vote::payload) {
+        // Populate `observations` cache using `interesting_content`, to support partial graphs...
+        for meta_event in parsed_contents.meta_events.values() {
+            for payload in &meta_event.interesting_content {
                 let hash = payload.create_hash();
                 let _ = parsec.observations.insert(hash, ObservationInfo::default());
             }
         }
 
-        // ..and also `interesting_content`, to support partial graphs.
-        for meta_event in parsed_contents.meta_events.values() {
-            for payload in &meta_event.interesting_content {
-                let hash = payload.create_hash();
-                let _ = parsec.observations.insert(hash, ObservationInfo::default());
+        // ..and also the payloads carried by events.
+        let our_pub_id = parsec.our_pub_id().clone();
+        for event in parsed_contents.events.values() {
+            if let Some(payload) = event.vote().map(Vote::payload) {
+                let observation = parsec
+                    .observations
+                    .entry(payload.create_hash())
+                    .or_insert_with(ObservationInfo::default);
+
+                if *event.creator() == our_pub_id {
+                    observation.created_by_us = true;
+                }
             }
         }
 
@@ -2403,7 +2410,7 @@ mod functional_tests {
         let a_0_hash = alice.events_order[0];
         let a_1_hash = alice.events_order[1];
 
-        // Create Dave where the first event is a geneneis event containing both Alice and Dave.
+        // Create Dave where the first event is a genesis event containing both Alice and Dave.
         let mut dave = initialise_parsec(dave_id.clone(), genesis, None);
         assert!(!dave.events.contains_key(&a_0_hash));
         assert!(!dave.events.contains_key(&a_1_hash));
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -2130,7 +2130,7 @@ mod functional_tests {
 
         let mut alice = Parsec::from_parsed_contents(parsed_contents);
         initialise_membership_lists(&mut alice.peer_list);
-        let genesis_group: BTreeSet<_> = alice.peer_list.all_ids().into_iter().cloned().collect();
+        let genesis_group: BTreeSet<_> = alice.peer_list.all_ids().cloned().collect();
 
         let fred_id = PeerId::new("Fred");
         assert!(!alice.peer_list.all_ids().any(|peer_id| *peer_id == fred_id));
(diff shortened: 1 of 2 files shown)


src/parsec.rs
```diff
@@ -1802,19 +1802,26 @@ impl Parsec<Transaction, PeerId> {
...
+        for event in parsed_contents.events.values() {
+            if let Some(payload) = event.vote().map(Vote::payload) {
+                let observation = parsec
+                    .observations
+                    .entry(payload.create_hash())
+                    .or_insert_with(ObservationInfo::default);
+
+                if *event.creator() == our_pub_id {
+                    observation.created_by_us = true;
+                }
```
octol: Help me understand why the two blocks were swapped around, is there a dependency I'm missing? Is the main difference the `created_by_us` variable being set? Was thinking maybe this change can be it's own commit that goes before the test?

Fraser999: You're right - let me break this into two commits.  I'll explain in the commit message.

octol: Nice, makes it easier to see why it was done :)

src/parsec.rs
```diff
@@ -2123,7 +2130,7 @@ mod functional_tests {
 
         let mut alice = Parsec::from_parsed_contents(parsed_contents);
         initialise_membership_lists(&mut alice.peer_list);
-        let genesis_group: BTreeSet<_> = alice.peer_list.all_ids().into_iter().cloned().collect();
+        let genesis_group: BTreeSet<_> = alice.peer_list.all_ids().cloned().collect();
```
octol: Of course personally I think this should also be it's own commit ;)
...but it's so tiny that probably not worth extracting out at this point

src/parsec.rs
```diff
@@ -2607,4 +2614,69 @@ mod functional_tests {
...
+        let genesis = Observation::Genesis(mock::create_ids(4).into_iter().collect());
+        {
+            let mut unpolled_observations = alice.our_unpolled_observations();
+            assert_eq!(*unwrap!(unpolled_observations.next()), genesis);
+            assert_eq!(*unwrap!(unpolled_observations.next()), add_eric);
+            assert!(unpolled_observations.next().is_none());
+        }
+
+        // Call `poll()` and retry - should only return our vote for `Add(Eric)`.
+        unwrap!(alice.poll());
```
octol: Possible to check the returned block from `poll`? Would make it slightly quicker to read what's happening. That said it's still very readable what's going on...

src/parsec.rs
```diff
@@ -2607,4 +2614,69 @@ mod functional_tests {
...
+        unwrap!(alice.add_event(b_17));
+
+        // Since we haven't call `poll()` again yet, should still return our vote for `Add(Eric)`.
+        // However, `has_unconsensused_observations()` should now return false.
+        assert!(!alice.has_unconsensused_observations());
+        assert_eq!(alice.our_unpolled_observations().count(), 1);
+        assert_eq!(*unwrap!(alice.our_unpolled_observations().next()), add_eric);
+
+        // Call `poll()` and retry - should return none.
+        unwrap!(alice.poll());
```
octol: Same as the other `poll` comment

src/parsec.rs
Fraser999: I did nearly do that actually, but decided against it in order to keep the test focused on what was relevant to the task; i.e. unpolled observations and our unconsensused observations.

src/parsec.rs
```diff
@@ -2614,4 +2614,69 @@ mod functional_tests {
...
+
+    #[test]
+    fn unpolled_and_unconsensused_observations() {
+        let mut alice_contents = parse_test_dot_file("alice.dot");
+        let b_17 = unwrap!(alice_contents.remove_latest_event());
+
+        let mut alice = Parsec::from_parsed_contents(alice_contents);
+        alice.restart_consensus(); // This is needed so the Genesis observation is consensused.
+
+        // `Add(Eric)` should still be unconsensused.
```
pierrechevalier83: I would add: `since b_17 would be the first gossip event to reach consensus on Add(Eric), but it was removed from the graph`.

src/parsec.rs
```diff
@@ -2614,4 +2614,69 @@ mod functional_tests {
...
+            assert_eq!(*unwrap!(unpolled_observations.next()), genesis);
+            assert_eq!(*unwrap!(unpolled_observations.next()), add_eric);
+            assert!(unpolled_observations.next().is_none());
+        }
+
+        // Call `poll()` and retry - should only return our vote for `Add(Eric)`.
+        unwrap!(alice.poll());
+        assert!(alice.poll().is_none());
+        assert!(alice.has_unconsensused_observations());
+        assert_eq!(alice.our_unpolled_observations().count(), 1);
```
pierrechevalier83: nitpick: instead of checking the size is 1 and then checking next, I would check next twice and make sure it's none the second time to be more symmetrical with the previous check. (Super minor but i like symmetry :p )

src/parsec.rs
```diff
@@ -2614,4 +2614,69 @@ mod functional_tests {
...
+        assert_eq!(*unwrap!(alice.our_unpolled_observations().next()), add_eric);
+
+        // Have Alice process B_17 to get consensus on `Add(Eric)`.
+        unwrap!(alice.add_event(b_17));
+
+        // Since we haven't call `poll()` again yet, should still return our vote for `Add(Eric)`.
+        // However, `has_unconsensused_observations()` should now return false.
+        assert!(!alice.has_unconsensused_observations());
+        assert_eq!(alice.our_unpolled_observations().count(), 1);
```
pierrechevalier83: same as above

src/parsec.rs
Fraser999: Same here, but I preferred this option to having to keep the iterator in a variable since that variable needs to be destroyed once we're finished the check (hence the extra level of nesting in the previous check)

PR #153, opened 2018-10-26
chore/tutorial: indicate rust version

Fraser999:


pierrechevalier83:
I'm going to bypass the second opinion for this given how trivial it is. Merging.

PR #154, opened 2018-10-26
MAID-3050 test/malice: accomplice

maqi:


src/parsec.rs
```diff
@@ -2680,4 +2680,85 @@ mod functional_tests {
...
+                malice: Malice::Fork(d_1_hash),
+            },
+            &alice_contents.events,
+            &alice_contents.peer_list,
+        );
+        let a_11_hash = *a_11.hash();
+        println!(
+            "a_10_hash {:?} a_11_hash {:?} d_1_hash {:?}",
+            a_10_hash, a_11_hash, d_1_hash
+        );
```
octol: Is this `println` leftover?

src/parsec.rs
```diff
@@ -2680,4 +2680,85 @@ mod functional_tests {
...
+        unwrap!(bob.handle_request(alice.our_pub_id(), message));
+        assert!(bob.events.contains_key(&a_11_hash));
+
+        let mut carol = Parsec::from_parsed_contents(parse_test_dot_file("carol.dot"));
+        assert!(!carol.events.contains_key(&a_11_hash));
+
+        // Send gossip from Bob to Carol, remove the accusation event
+        let mut message = unwrap!(bob.create_gossip(Some(&PeerId::new("Carol"))));
+        let num_of_events = message.packed_events.len();
+        let accusation_event = message.packed_events.remove(num_of_events - 1);
```
octol: Maybe assert the identity of the accusation event removed?

input_graphs/parsec_functional_tests_handle_malice_accomplice/alice.dot
```diff
@@ -0,0 +1,518 @@
+digraph GossipGraph {
```
pierrechevalier83: Uhm. All the dot graphs will need to be ported to the new syntax... Order of merging PRs will matter... @Fraser999, are you writing a script to port from old style to new style?

src/parsec.rs
```diff
@@ -2607,4 +2607,85 @@ mod functional_tests {
...
```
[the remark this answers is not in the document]
maqi: I don't think an assertion required, as it's a rust function.

src/parsec.rs
maqi: I don't think an assertion required, as it's a rust function.
Event it's not removed, the later on test will fail.

input_graphs/parsec_functional_tests_handle_malice_accomplice/alice.dot
maqi: shall be easy to re-generate those dot files for this test. However not sure about all others.

src/parsec.rs
```diff
@@ -2709,26 +2710,60 @@ mod functional_tests {
...
+                    _ => None,
+                }).next()
+        );
+        assert_eq!(offender, suspect);
+        assert_eq!(hash, event_hash);
+    }
+
+    #[test]
+    #[ignore]
+    // Carol received received gossip from Bob, which shall raise an accusation against Alice but
```
pierrechevalier83: "received received"

src/parsec.rs
```diff
@@ -2741,9 +2776,81 @@ mod functional_tests {
...
-        assert_eq!(*hash, a_11_hash);
+        assert_eq!(*hash, invalid_accusation);
 
-        let (offender, hash) = unwrap!(
+        verify_accused_accomplice(&carol, bob.our_pub_id(), bob_last_hash);
+    }
+
+    #[test]
+    #[ignore]
+    // Carol received `invalid_accusation` from Alice first, then receive gossip from Bob, which
```
pierrechevalier83: then received ... which should have raised ...

src/parsec.rs
```diff
@@ -2680,4 +2680,186 @@ mod functional_tests {
...
+        assert!(
+            our_votes(&carol)
+                .filter_map(|payload| match payload {
+                    Observation::Accusation {
+                        ref offender,
+                        malice: Malice::Accomplice(hash),
+                    } => Some((offender, hash)),
+                    _ => None,
+                }).next()
+                .is_none()
```
madadam: You are not using `offender` or `hash` here, so you could simplify this:

```rust
assert!(our_votes(...).all(|payload| match payload {
    Observation::Accusation { malice: Malice::Accomplice(_), ... } => false,
    _ => true,
}))
```

src/parsec.rs
```diff
@@ -2680,4 +2680,186 @@ mod functional_tests {
...
+
+        let mut carol = Parsec::from_parsed_contents(parse_dot_file_with_test_name(
+            "carol.dot",
+            "parsec_functional_tests_handle_malice_accomplice",
+        ));
+        assert!(!carol.events.contains_key(&invalid_accusation));
+
+        // Send gossip from Bob to Carol, remove the accusation event
+        let mut message = unwrap!(bob.create_gossip(Some(&PeerId::new("Carol"))));
+        let accusation_event = unwrap!(message.packed_events.pop());
```
madadam: Might be worth asserting that the removed event does really carry `Accusation { malice: Malice::InvalidAccusation(invalid_accusation), ..}`, to make the test more future-proof.

src/parsec.rs
```diff
@@ -2680,4 +2680,186 @@ mod functional_tests {
...
+        assert!(!bob.events.contains_key(&invalid_accusation));
+
+        // Send gossip from Alice to Bob
+        let message = unwrap!(alice.create_gossip(Some(&PeerId::new("Bob"))));
+        unwrap!(bob.handle_request(alice.our_pub_id(), message));
+        assert!(bob.events.contains_key(&invalid_accusation));
+
+        // Send gossip from Bob to Carol, remove the accusation event
+        let mut message = unwrap!(bob.create_gossip(Some(&PeerId::new("Carol"))));
+        let accusation_event = unwrap!(message.packed_events.pop());
```
madadam: See https://github.com/maidsafe/parsec/pull/154/files#r229213607

src/parsec.rs
maqi: this is a `packed_event` with accusation being burried under PackedEvent::Content::Cause::Vote::payload::Observation .
Meanwhile if such event left over in the message, it shall fail the later on test.
I do think it requires penetrating too much, and doesn't worth it.

src/parsec.rs
maqi: same as above

src/parsec.rs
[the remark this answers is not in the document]
maqi: To extract  the middle part ideally it shall be done in a macro, otherwise we will need an extra enum to pass in expected malice type to verify against.
I will prefer to do it in a separate task if we do think worth it.

src/parsec.rs
```diff
@@ -2734,8 +2734,8 @@ mod functional_tests {
 
     #[test]
     #[ignore]
-    // Carol received received gossip from Bob, which shall raise an accusation against Alice but
-    // didn't.
+    // Carol received gossip from Bob, which should have raised an an accomplice accusation against
```
pierrechevalier83: an an

PR #155, opened 2018-10-29
Fix: Avoid detecting self-malice

madadam:
Also fix a bug in the invalid accusation test.


src/parsec.rs
```diff
@@ -1609,6 +1609,11 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
     }
 
     fn detect_invalid_accusation(&mut self, event: &Event<T, S::PublicId>) {
+        // We can't detect this type of malice for ourselves.
+        if event.creator() == self.our_pub_id() {
```
pierrechevalier83: This will fix the bug you found. I'm wondering if more generally we shouldn't always skip ourselves from any Malice accusations, though... We already know we're honest...

PR #157, opened 2018-10-31
Fix gossip creation in the presence of forks

madadam:


src/gossip/event.rs
```diff
@@ -142,6 +144,7 @@ impl<T: NetworkEvent, P: PublicId> Event<T, P> {
             signature: packed_event.signature,
             hash,
             index,
+            order: events.len(),
```
pierrechevalier83: Is this one-based? (i.e: has the event hash already been pushed?) If it is, I had rather use -1 and be 0-based as programmers would tend to expect in general.

src/gossip/event.rs
```diff
@@ -36,6 +36,8 @@ pub(crate) struct Event<T: NetworkEvent, P: PublicId> {
     index: u64,
     // Index of each peer's latest event that is an ancestor of this event.
     last_ancestors: BTreeMap<P, u64>,
+    // Topological order of this event relative to all other events in the gossip graph.
+    order: usize,
```
pierrechevalier83: Just a thought: this topological order will be different from each node's point of view. We need to make sure that it is never used for instance to compare events or when creating the hash of a gossip event, or it could cause issues.

pierrechevalier83: For this reason, I'm unsure if it's a good idea to add this field. I'm open to discussion though.

src/gossip/graph.rs
```diff
@@ -0,0 +1,98 @@
...
+        graph,
+        queue,
+        visited: vec![false; event.order() + 1],
+    }
+}
+
+pub(crate) struct Ancestors<'a, T: NetworkEvent + 'a, P: PublicId + 'a> {
+    graph: &'a BTreeMap<Hash, Event<T, P>>,
+    queue: BTreeMap<usize, &'a Event<T, P>>,
+    visited: Vec<bool>, // TODO: replace with bitset, for space efficiency
```
pierrechevalier83: Did you mean to do this TODO?

src/gossip/graph.rs
```diff
@@ -0,0 +1,98 @@
...
+use network_event::NetworkEvent;
+use std::collections::BTreeMap;
+
+/// Iterator over all ancestors of the given event (including itself) in reverse topological order.
+pub(crate) fn ancestors<'a, T: NetworkEvent, P: PublicId>(
+    graph: &'a BTreeMap<Hash, Event<T, P>>,
+    event: &'a Event<T, P>,
+) -> Ancestors<'a, T, P> {
+    let mut queue = BTreeMap::new();
+    let _ = queue.insert(event.order(), event);
```
pierrechevalier83: At first read, I don't get it: we only insert one event?

src/parsec.rs
```diff
@@ -1409,45 +1409,24 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-        // the early events in `self.events_order` to be skipped mistakenly. To avoid this, if there
-        // are any peers for which `peer_id` doesn't have a `last_ancestors` entry, add those peers'
-        // oldest events we know about to the list of hashes.
-        for (id, peer) in self.peer_list.iter() {
-            if !peer_last_event.last_ancestors().contains_key(id) {
-                for hash in peer.events_by_index(0) {
-                    let _ = last_ancestors_hashes.insert(hash);
-                }
-            }
+        let mut mask = vec![true; self.events.len()];
```
pierrechevalier83: As I'm reading this, on first read I don't know what this mask represents. Could we try and give it a name that gives some more context?

src/gossip/graph.rs
```diff
@@ -0,0 +1,98 @@
...
+    graph: &'a BTreeMap<Hash, Event<T, P>>,
+    queue: BTreeMap<usize, &'a Event<T, P>>,
+    visited: Vec<bool>, // TODO: replace with bitset, for space efficiency
+}
+
+impl<'a, T: NetworkEvent, P: PublicId> Iterator for Ancestors<'a, T, P> {
+    type Item = &'a Event<T, P>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
```
pierrechevalier83: Why loop if we return at the end of the loop (or am I missing something?)

src/parsec.rs
pierrechevalier83: `event_ancestors_mask`?

src/gossip/event.rs
madadam: No, it's zero-based. This event hasn't yet been added to the graph when this is called.

src/gossip/event.rs
madadam: The field is pretty essential to the `ancestors` algorithm, so the events are returned in the same order as they were inserted into our graph (well, in reverse). I think it's not that hard to make sure it's never used when hashing or comparison. Might be worth writing a test for it though.

src/gossip/graph.rs
madadam: No. Just something we can consider in the performance optimization phase. Also, as there is no bitset in the std, it would require external crate.

src/gossip/graph.rs
madadam: I think this should become clearer if you read the `next` method. Or should I add a comment with a short explanation of the algorithm?

src/gossip/graph.rs
madadam: To skip already visited events.

src/parsec.rs
madadam: I don't think it's necessary to qualify the name because there is only one "mask" and the method returns (an iterator over) events. So it should be obvious that it is a mask of events?

But maybe the name "mask" is not clear? What about "presence"?

Fraser999: Yeah, +1 for `presence`.

src/gossip/event.rs
Fraser999: Maybe `topo_index` would be more meaningful?  If we did that, I guess there'd be reason then to also change `index` to e.g. `index_by_creator` or something?  I'm not strong on this though - just think it might be good to have `topo` in the name.

src/gossip/event.rs
```diff
@@ -434,22 +330,189 @@ impl Event<Transaction, PeerId> {
...
+            cache,
+        }
+    }
+}
+
+// Properties of `Event` that can be computed from its content.
+#[serde(bound = "")]
+#[derive(Serialize, Deserialize)]
+struct Cache<P: PublicId> {
+    // Event hash
```
pierrechevalier83: Can we be more specific and say:
```
// Hash of the event's content
```
?

src/gossip/event.rs
```diff
@@ -434,22 +330,189 @@ impl Event<Transaction, PeerId> {
...
+    hash: Hash,
+    // Sequential index of this event: this event is the `index`-th one made by its creator.
+    index: u64,
+    // Index of each peer's latest event that is an ancestor of this event.
+    last_ancestors: BTreeMap<P, u64>,
+    // Topological order of this event relative to all other events in the gossip graph.
+    order: usize,
```
pierrechevalier83: maybe `topological_order` would be a slightly more informative name?

madadam: This is what @Fraser999 is suggesting too. I'm not completely against it, but if we do this we should probably consider renaming `index` to something more descriptive? What about:

- `index` -> `creator_index`
- `order` -> `topological_index` (alternatively, `global_index`, `graph_index`, ...)
?

pierrechevalier83: `index` -> `index_by_creator`? `creator_index` sounds like the index of the creator (something like 0 for Alice, 1 for Bob etc.)

src/parsec.rs
```diff
@@ -1408,46 +1389,30 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-        // the early events in `self.events_order` to be skipped mistakenly. To avoid this, if there
-        // are any peers for which `peer_id` doesn't have a `last_ancestors` entry, add those peers'
-        // oldest events we know about to the list of hashes.
-        for (id, peer) in self.peer_list.iter() {
-            if !peer_last_event.last_ancestors().contains_key(id) {
-                for hash in peer.events_by_index(0) {
-                    let _ = last_ancestors_hashes.insert(hash);
-                }
-            }
+        // Events to include in the result. Initially start with including everything...
```
maqi: Events to be included in ...

src/parsec.rs
```diff
@@ -1408,46 +1389,30 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+        let mut inclusion_list = vec![true; self.events.len()];
+
+        // ...then exclude events that are ancestors of `last_event`, because that are the events
+        // that the peer already has,
```
maqi: because those are the events that the peer already has.

src/gossip/event.rs
pierrechevalier83: ^^ This is the last outstanding comment as far as I can see.

src/parsec.rs
```diff
@@ -534,34 +529,13 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-
         if let Some(observation) = event.vote().map(Vote::payload) {
             let info = self
                 .observations
                 .entry(observation.create_hash())
                 .or_insert_with(ObservationInfo::default);
 
-            if event.creator() == self.peer_list.our_pub_id() {
+            if our {
                 info.created_by_us = true;
```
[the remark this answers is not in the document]
pierrechevalier83: It's not equivalent because we don't say
```
... } else {
    info.created_by_us = false;
}
```

src/gossip/event.rs
```diff
@@ -354,9 +246,17 @@ impl<T: NetworkEvent, P: PublicId> Event<T, P> {
     }
 }
 
+impl<T: NetworkEvent, P: PublicId> PartialEq for Event<T, P> {
+    fn eq(&self, other: &Self) -> bool {
+        self.content == other.content
```
Fraser999: I'm not sure if the `signature` should be included here?  Was that deliberately rejected?

src/gossip/messages.rs
```diff
@@ -18,10 +18,10 @@ pub struct Request<T: NetworkEvent, P: PublicId> {
     pub(crate) packed_events: Vec<PackedEvent<T, P>>,
 }
 
-impl<'a, T: 'a + NetworkEvent, P: 'a + PublicId> Request<T, P> {
-    pub(crate) fn new<I: Iterator<Item = &'a Event<T, P>>>(events_iter: I) -> Self {
+impl<T: NetworkEvent, P: PublicId> Request<T, P> {
+    pub(crate) fn new(events: Vec<&Event<T, P>>) -> Self {
```
Fraser999: I expect there was a reason to change from an iterator here, but for my own benefit, can you explain briefly please?

Fraser999: OK - thanks for the explanation :)

src/gossip/event.rs
Fraser999: Hmm, I could probably agree at the conceptual level, but I believe some signing schemes will cause signatures to be different even when signing the same data with the same secret key.

In that case, we might want to treat two validly-signed `Event`s with identical content as always equal... but I'd be inclined to _not_ do that, since I don't think we have a valid use-case for a peer to sign the same event twice.

So, maybe it's better to include the `signature` in the comparison for now, to err on the side of safety?

PR #158, opened 2018-10-31
MAID-3076 feat/malice: detect other parent by same creator

maqi:


src/parsec.rs
```diff
@@ -1492,6 +1494,27 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         Ok(())
     }
 
+    // Detect if the event's other_parent has the same creator as this event.
+    fn detect_other_parent_by_same_creator(&mut self, event: &Event<T, S::PublicId>) -> Result<()> {
+        if event.other_parent().is_none() {
```
Fraser999: It might be better to have this as an `else` block of `if let Some(other_parent) = self.other_parent(event) {` below?

src/observation.rs
```diff
@@ -65,5 +65,8 @@ pub enum Malice {
     /// Contains hash of the sync event whose creator shall detect such malice however failed to
     /// raise an accusation.
     Accomplice(Hash),
+    /// Event's creator is the same to its other_parent's creator. The accusation contains the
+    /// original event so other peers can verify the accusation directly.
+    OtherParentBySameCreator(Vec<u8>),
```
Fraser999: I think this should contain a `PackedEvent` rather than a serialised `Event`.  If we don't want to add `PackedEvent` to the lib's API, I guess it could still be a `Vec<u8>` but then I'd make it a serialised `PackedEvent`.

My preference is to make `PackedEvent` public, and use that here (I'm assuming it would have to be exposed if we do that).

src/parsec.rs
pierrechevalier83: Following this line of thought, this entire code could become one statement with no early return:
```
if let Some(...) ... {
    if other_parent ...
    {
        self.create_accusation...
        Err(...)
    } else {
        Ok()
    }
} else {
    Ok()
}
````

Comment (maqi) [reply src/parsec.rs]: there is mutable borrower check issue as `self.other_parent(event)` returns a ref inside option, which requires immutable borrow of self, meanwhile the inside self.create_accusation requires a mutable borrow of self.

File: src/parsec.rs [hunk 3]
```diff
@@ -1459,6 +1461,8 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         // NOTE: `detect_incorrect_genesis` must come first.
         self.detect_incorrect_genesis(event)?;
 
+        self.detect_other_parent_by_same_creator(event)?;
+
```
pierrechevalier83: Super nitpick, but maybe we can remove the blank line after this line then?

src/parsec.rs
pierrechevalier83: OK. So this prevents my suggestion, but it doesn't block Fraser's, does it?

src/parsec.rs
maqi: the other detect functions won't return with error, i.e. later on events in the same msg will be processed.
meanwhile this function will return with an error and cause the later on events in the same msg be `discarded`.

so this extra blank line is left purposely.

src/parsec.rs
maqi: Yes, So I have moved it into the else block

src/parsec.rs
pierrechevalier83: Uhm. OK. Since you've got a reasoning behind it, I won't insist.

PR #160, opened 2018-11-01
MAID-3212: update test dot files to new format

Fraser999:


src/parsec.rs
```diff
@@ -1807,6 +1807,13 @@ impl Parsec<Transaction, PeerId> {
             }
         }
 
+        for consensused in parsed_contents.meta_elections.consensus_history() {
```
pierrechevalier83: This could be its in own commit with descriptive message for visibility.

src/dump_graph.rs
```diff
@@ -522,4 +216,550 @@ mod detail {
...
+                    .to_string(),
+            );
+        }
+        for (peer, meta_votes) in meta_votes {
+            let mut prefix = format!("{}: ", first_char(peer).unwrap_or('?'));
+            for mv in meta_votes {
+                let est = mv.estimates.as_short_string();
+                let bin = mv.bin_values.as_short_string();
+                let aux = match mv.aux_value {
+                    None => "-",
```
maqi: this map `Option<bool>` could be a function to be used for all?

src/dump_graph.rs
```diff
@@ -522,4 +216,550 @@ mod detail {
...
+                        attr.to_string()
+                    ))?;
+                    event.write_to_dot_format(&mut self.file)?;
+                    self.writeln(format_args!(""))?;
+                }
+            }
+            Ok(())
+        }
+
+        fn write_meta_elections(&mut self) -> io::Result<()> {
```
maqi: this function is really large.
could the part inside `for handle in self.meta_elections.all() {` be extracted as a separate function?
and even the part inside `for (short_name, mev) in meta_events.values() {` & `for (peer, hashes) in &election.round_hashes {` be extracted as well?

src/hash.rs
```diff
@@ -29,6 +30,18 @@ impl Hash {
         }
         Ordering::Equal
     }
+
+    pub fn from_bytes(bytes: [u8; HASH_LEN]) -> Self {
+        Hash(bytes)
+    }
+
+    pub fn as_full_string(&self) -> String {
```
maqi: do we need thse two new functions for the production code?

src/dump_graph.rs
Fraser999: Yes, it is large, I agree.

However, I'd prefer to not change this now that it has been coded this way.  I don't think splitting this will improve readablility since the logic inside it is all very straightforward - we're just iterating all the parts of meta-elections and dumping them to the `lines` vector.

PR #163, opened 2018-11-04
MAID-3049: detect accomplice malice

octol:
Detect when a peer relays us malice that they themselves have not raised any accusations for.

### DIFF
--- a/src/functional_tests.rs
+++ b/src/functional_tests.rs
@@ -485,6 +485,7 @@ mod handle_malice {
     use id::SecretId;
     use mock::Transaction;
     use observation::Malice;
+    use observation::UnprovableMalice;
     use peer_list::PeerList;
     use peer_list::PeerState;
     use std::collections::BTreeMap;
@@ -969,10 +970,8 @@ mod handle_malice {
 
     fn create_invalid_accusation() -> (EventHash, TestParsec<Transaction, PeerId>) {
         // Generated with RNG seed: [3932887254, 691979681, 2029125979, 3359276664]
-        let mut alice_contents = parse_dot_file_with_test_name(
-            "alice.dot",
-            "functional_tests_handle_malice_accomplice",
-        );
+        let mut alice_contents =
+            parse_dot_file_with_test_name("alice.dot", "functional_tests_handle_malice_accomplice");
 
         let a_26_hash = *unwrap!(find_event_by_short_name(&alice_contents.graph, "A_26")).hash();
         let d_1_hash = *unwrap!(find_event_by_short_name(&alice_contents.graph, "D_1")).hash();
@@ -1004,7 +1003,7 @@ mod handle_malice {
                 .filter_map(|payload| match payload {
                     Observation::Accusation {
                         ref offender,
-                        malice: Malice::Accomplice(hash),
+                        malice: Malice::Unprovable(UnprovableMalice::Accomplice(hash)),
                     } => Some((offender, hash)),
                     _ => None,
                 }).next()
@@ -1014,7 +1013,6 @@ mod handle_malice {
     }
 
     #[test]
-    #[ignore]
     // Carol received gossip from Bob, which should have raised an accomplice accusation against
     // Alice but didn't.
     fn accomplice() {
@@ -1063,7 +1061,6 @@ mod handle_malice {
     }
 
     #[test]
-    #[ignore]
     // Carol received `invalid_accusation` from Alice first, then received gossip from Bob,
     // which should have raised an accomplice accusation against Alice but didn't.
     fn accomplice_separate() {
@@ -1103,7 +1100,6 @@ mod handle_malice {
     }
 
     #[test]
-    #[ignore]
     // Carol received `invalid_accusation` from Alice first, then receive gossip from Bob, which
     // doesn't contain the malice of Alice. Carol shall not raise accusation against Bob.
     fn accomplice_negative() {
@@ -1133,7 +1129,7 @@ mod handle_malice {
         // Verify that Carol didn't accuse Bob of `Accomplice`.
         assert!(our_votes(&carol).all(|payload| match payload {
             Observation::Accusation {
-                malice: Malice::Accomplice(_),
+                malice: Malice::Unprovable(UnprovableMalice::Accomplice(_)),
                 ..
             } => false,
             _ => true,
--- a/src/functional_tests.rs
+++ b/src/functional_tests.rs
@@ -968,17 +968,18 @@ mod handle_malice {
     }
 
     fn create_invalid_accusation() -> (EventHash, TestParsec<Transaction, PeerId>) {
+        // Generated with RNG seed: [3932887254, 691979681, 2029125979, 3359276664]
         let mut alice_contents = parse_dot_file_with_test_name(
             "alice.dot",
-            "parsec_functional_tests_handle_malice_accomplice",
+            "functional_tests_handle_malice_accomplice",
         );
 
-        let a_10_hash = *unwrap!(find_event_by_short_name(&alice_contents.graph, "A_10")).hash();
+        let a_26_hash = *unwrap!(find_event_by_short_name(&alice_contents.graph, "A_26")).hash();
         let d_1_hash = *unwrap!(find_event_by_short_name(&alice_contents.graph, "D_1")).hash();
 
         // Create an invalid accusation from Alice
-        let a_11 = Event::<Transaction, _>::new_from_observation(
-            a_10_hash,
+        let a_27 = Event::<Transaction, _>::new_from_observation(
+            a_26_hash,
             Observation::Accusation {
                 offender: PeerId::new("Dave"),
                 malice: Malice::Fork(d_1_hash),
@@ -986,11 +987,11 @@ mod handle_malice {
             &alice_contents.graph,
             &alice_contents.peer_list,
         );
-        let a_11_hash = *a_11.hash();
-        alice_contents.add_event(a_11);
+        let a_27_hash = *a_27.hash();
+        alice_contents.add_event(a_27);
         let alice = TestParsec::from_parsed_contents(alice_contents);
-        assert!(alice.graph().contains(&a_11_hash));
-        (a_11_hash, alice)
+        assert!(alice.graph().contains(&a_27_hash));
+        (a_27_hash, alice)
     }
 
     fn verify_accused_accomplice(
@@ -1021,7 +1022,7 @@ mod handle_malice {
 
         let mut bob = TestParsec::from_parsed_contents(parse_dot_file_with_test_name(
             "bob.dot",
-            "parsec_functional_tests_handle_malice_accomplice",
+            "functional_tests_handle_malice_accomplice",
         ));
         assert!(!bob.graph().contains(&invalid_accusation));
 
@@ -1032,7 +1033,7 @@ mod handle_malice {
 
         let mut carol = TestParsec::from_parsed_contents(parse_dot_file_with_test_name(
             "carol.dot",
-            "parsec_functional_tests_handle_malice_accomplice",
+            "functional_tests_handle_malice_accomplice",
         ));
         assert!(!carol.graph().contains(&invalid_accusation));
 
@@ -1070,7 +1071,7 @@ mod handle_malice {
 
         let mut carol = TestParsec::from_parsed_contents(parse_dot_file_with_test_name(
             "carol.dot",
-            "parsec_functional_tests_handle_malice_accomplice",
+            "functional_tests_handle_malice_accomplice",
         ));
         assert!(!carol.graph().contains(&invalid_accusation));
 
@@ -1081,7 +1082,7 @@ mod handle_malice {
 
         let mut bob = TestParsec::from_parsed_contents(parse_dot_file_with_test_name(
             "bob.dot",
-            "parsec_functional_tests_handle_malice_accomplice",
+            "functional_tests_handle_malice_accomplice",
         ));
         assert!(!bob.graph().contains(&invalid_accusation));
 
@@ -1110,7 +1111,7 @@ mod handle_malice {
 
         let mut carol = TestParsec::from_parsed_contents(parse_dot_file_with_test_name(
             "carol.dot",
-            "parsec_functional_tests_handle_malice_accomplice",
+            "functional_tests_handle_malice_accomplice",
         ));
         assert!(!carol.graph().contains(&invalid_accusation));
 
@@ -1121,7 +1122,7 @@ mod handle_malice {
 
         let bob = TestParsec::from_parsed_contents(parse_dot_file_with_test_name(
             "bob.dot",
-            "parsec_functional_tests_handle_malice_accomplice",
+            "functional_tests_handle_malice_accomplice",
         ));
         assert!(!bob.graph().contains(&invalid_accusation));
 
--- a/src/gossip/event.rs
+++ b/src/gossip/event.rs
@@ -166,7 +166,7 @@ impl<T: NetworkEvent, P: PublicId> Event<T, P> {
 
     // Returns the index-by-creator of the last ancestor of this event created by the given peer.
     pub fn last_ancestor_by(&self, peer: &P) -> LastAncestor {
-        if self.cache.forking_peers.contains(peer) {
+        if self.is_forking_peer(peer) {
             LastAncestor::Fork
         } else {
             self.cache
@@ -175,6 +175,10 @@ impl<T: NetworkEvent, P: PublicId> Event<T, P> {
                 .map(|last_index| LastAncestor::Some(*last_index))
                 .unwrap_or(LastAncestor::None)
         }
+    }
+
+    pub(crate) fn is_forking_peer(&self, peer: &P) -> bool {
+        self.cache.forking_peers.contains(peer)
     }
 
     /// Returns `Some(vote)` if the event is for a vote of network event, otherwise returns `None`.
--- a/src/gossip/split_on_first.rs
+++ b/src/gossip/split_on_first.rs
@@ -1,0 +1,65 @@
+// Copyright 2018 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use std::mem;
+
+/// Split a Vec<T> into chunks, where the element matched by the predicate is the first in each
+/// chunk.
+pub fn split_on_first<F, T>(items: Vec<T>, predicate: F) -> SplitOnFirst<T, F>
+where
+    F: FnMut(&T) -> bool,
+{
+    SplitOnFirst { items, predicate }
+}
+
+pub struct SplitOnFirst<T, P>
+where
+    P: FnMut(&T) -> bool,
+{
+    items: Vec<T>,
+    predicate: P,
+}
+
+impl<T, P> Iterator for SplitOnFirst<T, P>
+where
+    P: FnMut(&T) -> bool,
+{
+    type Item = Vec<T>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.items.is_empty() {
+            return None;
+        }
+
+        // Since the matched element is the first in the second chunk, we need to skip index == 0.
+        let mut index = 1;
+        while index < self.items.len() {
+            if (self.predicate)(&self.items[index]) {
+                break;
+            } else {
+                index += 1;
+            }
+        }
+
+        let new_items = self.items.split_off(index);
+        Some(mem::replace(&mut self.items, new_items))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::super::split_on_first;
+
+    #[test]
+    fn split_vec_of_numbers() {
+        assert_eq!(
+            split_on_first(vec![1, 2, 3, 4], |n| n % 2 == 0).collect::<Vec<Vec<_>>>(),
+            vec![vec![1], vec![2, 3], vec![4]]
+        )
+    }
+}
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -11,6 +11,7 @@ use block::Block;
 use dev_utils::ParsedContents;
 use dump_graph;
 use error::{Error, Result};
+use gossip::split_on_first;
 use gossip::{
     Event, EventHash, EventIndex, Graph, IndexedEventRef, PackedEvent, Request, Response,
     UnpackedEvent,
@@ -32,6 +33,8 @@ use std::{mem, usize};
 use vote::Vote;
 
 type PendingAccusations<T, P> = Vec<(P, Malice<T, P>)>;
+#[cfg(feature = "malice-detection")]
+type Accusations<T, P> = Vec<(P, Malice<T, P>)>;
 
 /// Returns whether `small` is more than two thirds of `large`.
 pub fn is_more_than_two_thirds(small: usize, large: usize) -> bool {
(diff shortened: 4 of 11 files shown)


src/parsec.rs
```diff
@@ -487,6 +487,7 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         }
     }
 
+    #[cfg_attr(feature = "cargo-clippy", allow(needless_pass_by_value))]
```
octol: I'll try to get rid of this

src/parsec.rs
```diff
@@ -1712,6 +1748,85 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+                let event_sees_malice = match malice {
+                    Malice::MissingGenesis(hash)
+                    | Malice::IncorrectGenesis(hash)
+                    | Malice::StaleOtherParent(hash)
+                    | Malice::InvalidAccusation(hash)
+                    | Malice::InvalidGossipCreator(hash)
+                    | Malice::Accomplice(hash) => {
+                        Some(current_event.sees(self.get_known_event(hash).unwrap()))
+                    }
+                    _ => None,
```
octol: There are a few categories, most notably `Fork` which still needs to be dealt with

src/parsec.rs
```diff
@@ -1712,6 +1748,85 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+    ) -> PendingAccusations<T, S::PublicId> {
+        let our_accusations = self
+            .peer_list
+            .our_events()
+            .filter_map(|hash| self.get_known_event(hash).ok())
+            .filter_map(|event| event.vote().map(Vote::payload))
+            .filter(|observation| match observation {
+                Observation::Accusation { offender, .. } => offender != event_creator,
+                _ => false,
+            });
```
octol: We can probably optimise here not looking at all accusations everytime

src/parsec.rs
```diff
@@ -1712,6 +1748,85 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+        }
+    }
+
+    fn detect_accomplice_for_past_events(
+        &self,
+        event_creator: &S::PublicId,
+        current_hash: Hash,
+        src: &S::PublicId,
+        first_event: Hash,
+    ) -> PendingAccusations<T, S::PublicId> {
```
pierrechevalier83: Instead of looking at all our events and all their events, I think we should restrict this to all our events since the self_parent of their oldest event in the `Request` or `Response` and we should also only consider accusations that are contained in the `Request` or `Response`

src/parsec.rs
```diff
@@ -1712,6 +1748,85 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+        first_event_in_chunk: Hash,
+    ) {
+        let pending_accusations = self.pending_accusations.clone();
+        for (offender, malice) in &pending_accusations {
+            if offender != creator {
+                // Did the sender also create this accusation?
+                let offender = offender.clone();
+                let malice = malice.clone();
+                let acc = Observation::Accusation { offender, malice };
+                if !self.accusations_by_peer(src).any(|a| a == &acc) {
```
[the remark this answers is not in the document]
octol: Actually that reminds me, need to add a guard so that we don't raise multiple duplicate accomplice accusations

src/parsec.rs
octol: > Instead of looking at all our events and all their events, I think we should restrict this to all our events since the self_parent of their oldest event in the Request or Response and we should also only consider accusations that are contained in the Request or Response

1. How do I get the hash of the events that are not unpacked?
2. About only considering accusations part of the Reqest/Response, can't the sender avoid being accused of accomplice malice by simply chancing that we already know of some of the events and omit some old events. Do we currently detect omitted events if those are events we already have in our graph?

src/parsec.rs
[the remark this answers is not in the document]
pierrechevalier83: > The accusation gets added to pending_accusations in the first chunk that sees the malice?
That's not the part that I'm disagreeing with. I'm saying if Bob misbehaves and Alice accuses Bob, but not in a timely manner, Alice is still an accomplice. By only checking whether Alice accused Bob, you miss one way she could be an accomplice, by only accusing Bob later.
The problem with this is that bad nodes could use this situation to make you look complicit or to make you think someone is accusing them invalidly...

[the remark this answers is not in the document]
octol: > By only checking whether Alice accused Bob, you miss one way she could be an accomplice, by only accusing Bob later.

Maybe I'm missing something here, but by checking against `pending_accusations` and our previously raised accusations don't we (Alice) accuse Bob at the earliest point in the graph where the malice is visible?

src/parsec.rs
```diff
@@ -1712,6 +1748,85 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+    }
+
+    fn detect_accomplice_for_past_events(
+        &self,
+        event_creator: &S::PublicId,
+        current_hash: Hash,
+        src: &S::PublicId,
+        first_event: Hash,
+    ) -> PendingAccusations<T, S::PublicId> {
+        let our_accusations = self
```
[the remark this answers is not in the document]
octol: Well the last filter is slightly different, but I can see if I can re-chain it

src/parsec.rs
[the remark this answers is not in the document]
octol: For completeness I'll add it here as here as well. Discussed this with Pierre directly and I think I convinced him that the accusation is added in a timely manner.

src/parsec.rs
```diff
@@ -1748,6 +1795,195 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+        // Compare against the malice we've detected
+        for a in self.detect_accomplice_for_pending_events(
+            event_creator,
+            first_event_in_chunk,
+            first_event_by_peer_in_packed_event,
+        )? {
+            self.accuse(a.0, a.1);
+        }
+        // Compare against our existing accusations too
+        for a in self.detect_accomplice_for_past_events(
```
maqi: same comment as above, better use `for (xx, xx) in ...` to show what are a.0 and a.1 explicitly

src/parsec.rs
```diff
@@ -1748,6 +1795,195 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+            event_creator.clone(),
+            Malice::Accomplice(first_event_in_chunk),
+        );
+        let mut new_accusations: PendingAccusations<T, S::PublicId> = vec![];
+        for (offender, malice) in self
+            .accusations_by_peer_since(self.our_pub_id(), starting_event)?
+            .iter()
+            .filter(|(off, _)| off != event_creator)
+        {
+            let current_event = self.get_known_event(&current_hash)?;
```
maqi: isn't this `crrent_event` can be outside of the for clause?

src/parsec.rs
```diff
@@ -490,9 +492,33 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+            let mut events = BTreeMap::new();
+            let all_first_parents =
+                packed_events
+                    .iter()
+                    .filter_map(|packed_event| match packed_event.self_parent() {
+                        Some(event_hash) => Some((packed_event.creator().clone(), *event_hash)),
+                        None => None,
+                    });
+
+            for e in all_first_parents {
```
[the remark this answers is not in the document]
octol: Yes good point!

src/parsec.rs
```diff
@@ -1748,6 +1796,195 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+        loop {
+            if let Some(Observation::Accusation { offender, malice }) =
+                event_by_src.vote().map(Vote::payload)
+            {
+                events.push((offender.clone(), malice.clone()));
+            }
+            if let Some(h) = event_by_src.self_parent() {
+                event_by_src = self.get_known_event(&h)?;
+                if event_by_src.topological_index() < oldest_event_considered.topological_index() {
+                    break;
```
fizyk20: This may break down if `src` has forked (branches other than the one `last_event(src)` is on will be ignored); but maybe we don't care about such situations, anyway?

src/parsec.rs
```diff
@@ -1748,6 +1796,195 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+        Ok(())
+    }
+
+    fn detect_accomplice_for_pending_events(
+        &self,
+        event_creator: &S::PublicId,
+        first_event_in_chunk: Hash,
+        first_events_in_pack: &BTreeMap<S::PublicId, Hash>,
+    ) -> Result<PendingAccusations<T, S::PublicId>> {
+        let starting_event = match first_events_in_pack.get(event_creator) {
```
pierrechevalier83:
```
let starting_event = first_events_in_pack.get(event_creator).ok_or(Error::Logic)?;
```

src/parsec.rs
```diff
@@ -1748,6 +1796,195 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+            Some(e) => e,
+            None => return Err(Error::Logic),
+        };
+        let accomplice_accusation = (
+            event_creator.clone(),
+            Malice::Accomplice(first_event_in_chunk),
+        );
+        let mut new_accusations: PendingAccusations<T, S::PublicId> = vec![];
```
pierrechevalier83: In general, if you're creating a mutable `Vec` and pushing back in a loop, you want to somehow `map` and `collect`.
Here, it would look like this:
```
        let accusations_since_starting_event =
            self.accusations_by_peer_since(event_creator, starting_event)?;
        Ok(self
            .pending_accusations
            .iter()
            .filter(|(offender, _)| offender != event_creator)
            .filter_map(|pending_accusation| {
                // Did the event creator also create this accusation?
                // Did we already accuse the sender of being an accomplice before?
                if !accusations_since_starting_event
                    .iter()
                    .any(|accusation| accusation == pending_accusation)
                    && !self
                        .pending_accusations
                        .iter()
                        .any(|accusation| accusation == &accomplice_accusation)
                {
                    Some(accomplice_accusation.clone())
                } else {
                    None
                }
            }).collect())
```

src/parsec.rs
```diff
@@ -1748,6 +1796,195 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+    }
+
+    fn detect_accomplice_for_past_events(
+        &self,
+        event_creator: &S::PublicId,
+        current_hash: Hash,
+        first_event_in_chunk: Hash,
+        first_events_in_pack: &BTreeMap<S::PublicId, Hash>,
+    ) -> Result<PendingAccusations<T, S::PublicId>> {
+        let starting_event = match first_events_in_pack.get(event_creator) {
```
pierrechevalier83: I'm not commenting in details here but the same stylistic comments as in the previous function apply.

src/parsec.rs
```diff
@@ -506,13 +532,34 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+
                 self.add_event(event)?;
                 // We have received an event of a peer in the message. The peer can now receive
                 // gossips from us as well.
                 self.peer_list
                     .change_peer_state(&event_creator, PeerState::RECV);
+
+                // If we're at the last event in the chunk.
+                if match packed_event_iter.peek() {
+                    Some(p) => p.other_parent().is_some(),
```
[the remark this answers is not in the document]
octol: Yes when `other_parent` is not `None` we are at the end of the (linear) chunk, similarly when `peek` is `None` we are at the end of the `Request`/`Response` and also at the end of the chunk

src/gossip/event.rs
```diff
@@ -127,6 +127,20 @@ impl<T: NetworkEvent, P: PublicId> Event<T, P> {
         }))
     }
 
+    #[cfg(feature = "malice-detection")]
+    pub(crate) fn unpack_hash(packed_event: PackedEvent<T, P>) -> Result<Hash, Error> {
```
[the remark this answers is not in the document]
octol: Yes good point!

src/parsec.rs
```diff
@@ -490,9 +492,34 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         self.confirm_peer_state(src, PeerState::SEND)?;
 
         let mut prev_forking_peers = BTreeSet::new();
-        for packed_event in packed_events {
+
+        // Keep track of the earliest self_parent for each peer in packed_events as we will use
+        // them as bounds in the graph for where we look for malice accusations.
+        // NOTE: there is the assumption here that the events arrive in order.
+        let _first_event_by_peer_in_packed_event = {
+            let mut events = BTreeMap::new();
```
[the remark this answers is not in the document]
octol: Looks good, I'll adopt it!

src/parsec.rs
```diff
@@ -490,9 +492,34 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+
+            for e in all_first_parents {
+                let _ = events.entry(e.0).or_insert(e.1);
+            }
+            events
+        };
+
+        // As we process the events in linear "chunks", we keep track of the first event in each
+        // chunk so we know which event to accuse of accomplice malice.
+        let mut remember_first_event_in_chunk: Option<Hash> = None;
```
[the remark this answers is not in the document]
octol: What about the name clash that creates?

src/parsec.rs
```diff
@@ -1748,6 +1796,195 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+                Ok(false)
+            }
+            Malice::Fork(_hash) => {
+                // TODO: the Fork case needs to be treated as a special case.
+                Ok(false)
+            }
+            Malice::OtherParentBySameCreator(packed_event)
+            | Malice::SelfParentByDifferentCreator(packed_event) => {
+                let event_hash = Event::unpack_hash(*packed_event.clone())?;
+                match self.events.get(&event_hash) {
```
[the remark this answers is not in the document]
octol: Do you mean `unwrap_or(false)`?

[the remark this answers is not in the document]
pierrechevalier83: It's an option isn't it?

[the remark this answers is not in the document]
pierrechevalier83: I think `or` on an `Option` is equivalent to `unwrap_or` on `Result`, but I could be mistaken :)

src/parsec.rs
octol: So this one I didn't manage to make the same transformation since malicious_event_is_seen_by_event returns a `Result`, and cannot be moved outside the `filter_map` (as far as I can tell)

src/parsec.rs
```diff
@@ -506,13 +533,34 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
                     let _ = prev_forking_peers.insert(event.creator().clone());
                 }
                 let event_creator = event.creator().clone();
+                let _event_hash = *event.hash();
```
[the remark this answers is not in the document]
octol: Note though there are a few more lines than those that would need feature gates otherwise you get unused variable errors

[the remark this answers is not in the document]
octol: Also, needs a explicit attribute to prevent clippy from complaining

src/parsec.rs
octol: Right that's a good point, will need to ponder this a little

src/parsec.rs
```diff
@@ -506,13 +533,34 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+
+                let _first_event_in_chunk =
```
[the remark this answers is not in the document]
octol: On Pierre's suggestion I tried go down the route of putting in more feature guards, but I'm not convinced the end result is any better. Let me know what you think about the updated version

src/parsec.rs
```diff
@@ -1748,6 +1796,195 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+        let mut events = vec![];
+
+        // First let's get the event for the oldest event we're considering
+        let oldest_event_considered = self.get_known_event(oldest_event)?;
+
+        // Then get the event where we start from
+        let mut event_by_src = match self
+            .peer_list
+            .last_event(src)
+            .and_then(|e| self.get_known_event(&e).ok())
```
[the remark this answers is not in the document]
octol: I'm probably missing something, but `last_event` returns an `Option<&Hash>` while `get_known_event` returns a `Result<&Event>`. So you can't chain them unless you convert using `ok()`? Well at least the following doesn't compile (types inserted to annotate)
```
let _event_by_src0: Option<&Hash> = self.peer_list.last_event(src);
let _event_by_src1 = _event_by_src0.and_then(|e| self.get_known_event(&e));
```

src/parsec.rs
```diff
@@ -1748,6 +1796,195 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+            Malice::DuplicateVote(hash0, hash1) => {
+                if let Ok(h0) = self.get_known_event(&hash0) {
+                    if let Ok(h1) = self.get_known_event(&hash1) {
+                        return Ok(event.sees(h0) && event.sees(h1));
+                    }
+                }
+                Ok(false)
+            }
+            Malice::Fork(_hash) => {
+                // TODO: the Fork case needs to be treated as a special case.
```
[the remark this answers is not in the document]
octol: Yes I need to think a little more about how to approach the fork case. Also that raises the question if that part should or should not be part of the same PR. I actually think the first commit "chore/parsec: update accomplice test dot files to new format" should have been a separate PR as we should just have merged that one straight away, no need for it to be held up by the rest. What are your thoughts? Should a PR be tied 1-to-1 to ticket?

src/parsec.rs
```diff
@@ -1748,6 +1796,195 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+                }
+            } else {
+                break;
+            }
+        }
+
+        Ok(events)
+    }
+
+    fn malicious_event_is_seen_by_event(
```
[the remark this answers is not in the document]
octol: Using `is_descendant_of()` should accomplish that right?

src/parsec.rs
```diff
@@ -1748,6 +1796,195 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+            | Malice::IncorrectGenesis(hash)
+            | Malice::StaleOtherParent(hash)
+            | Malice::InvalidAccusation(hash)
+            | Malice::InvalidGossipCreator(hash)
+            | Malice::Accomplice(hash) => match self.events.get(hash) {
+                Some(e) => Ok(event.sees(e)),
+                None => Ok(false),
+            },
+            Malice::DuplicateVote(hash0, hash1) => {
+                if let Ok(h0) = self.get_known_event(&hash0) {
```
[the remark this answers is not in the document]
octol: Ah that looks much nicer!

src/parsec.rs
```diff
@@ -1748,6 +1792,206 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+                        .get(hash)
+                        .map(|malicious_event| event.is_forking_peer(malicious_event.creator()))
+                        .unwrap_or(false)
+            }
+            Malice::OtherParentBySameCreator(packed_event)
+            | Malice::SelfParentByDifferentCreator(packed_event) => {
+                let event_hash = Event::verify_and_hash(packed_event)?;
+                self.events
+                    .get(&event_hash)
+                    .map(|malicious_event| event.sees(malicious_event))
```
pierrechevalier83: Any particular reason this one still requires `sees` when the other ones all simply use `is_descendant_of`?

octol: Yeah that was just an oversight when I changed the other ones. Thanks for spotting it :)

src/parsec.rs
```diff
@@ -506,13 +525,38 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
                 self.add_event(event)?;
                 // We have received an event of a peer in the message. The peer can now receive
                 // gossips from us as well.
                 self.peer_list
                     .change_peer_state(&event_creator, PeerState::RECV);
+
+                #[cfg(feature = "malice-detection")]
+                {
+                    let _ = first_event_in_chunk.get_or_insert(event_hash);
+                    // If `other_parent` is not `None` where at the nd of the (linear) chunk,
```
fizyk20: Typos in the comment - I guess it should've been "we're at the end", like below?

src/parsec.rs
```diff
@@ -1849,6 +1893,206 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+            Malice::Fork(hash) => {
+                self.events
+                    .get(hash)
+                    .map(|malicious_event| event.is_descendant_of(malicious_event))
+                    .unwrap_or(false)
+                    && self
+                        .events
+                        .get(hash)
+                        .map(|malicious_event| event.is_forking_peer(malicious_event.creator()))
+                        .unwrap_or(false)
```
fizyk20: Maybe put `&&` inside the `map()` call? Like `.map(|malicious_event| event.is_descendant_of(malicious_event) && event.is_forking_peer(malicious_event.creator()))`?

octol: Good point!

src/parsec.rs
```diff
@@ -1849,6 +1893,197 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+            }
+            if let Some(h) = event_by_peer.self_parent() {
+                event_by_peer = self.get_known_event(&h)?;
+                if event_by_peer.topological_index() < oldest_event_considered.topological_index() {
+                    break;
+                }
+            } else {
+                break;
+            }
+        }
```
octol: My idea was to replace this whole loop by creating a `LinkedList`, followed by a call to `take_while()`. But I got into problems with the borrow checker. Do you folks know if it's feasible?

src/gossip/event.rs
```diff
@@ -145,6 +145,15 @@ impl<T: NetworkEvent, P: PublicId> Event<T, P> {
         }
     }
 
+    pub(crate) fn sees_including_forks<E: AsRef<Event<T, P>>>(&self, other: E) -> bool {
```
pierrechevalier83: Actually, I don't think this implementation is correct in the presence of forks:
The `last_ancestor` of an event could have index 10, and there could be a fork that happened at index 2 where the other event is not an ancestor of this one.
This function would only work in the absence of fork, but it's not what we want here.

We'll need to be more precise and I think it has to have a cost in complexity.

pierrechevalier83: Look at `Parsec::has_interesting_ancestor` for a hint on how to make this efficient. Also, you probably want to check `sees_fork` first to only do the work when needed

pierrechevalier83: I think I would name this function: `has_for_ancestor` as `sees` is an overloaded term...

src/parsec.rs
```diff
@@ -504,11 +523,35 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
                 self.peer_list
                     .change_peer_state(&event_creator, PeerState::RECV);
+
+                #[cfg(feature = "malice-detection")]
+                {
+                    let _ = first_event_in_chunk.get_or_insert(event_hash);
+                    // If `other_parent` is not `None` we're at the end of the (linear) chunk,
+                    // similarly when `peek` is `None` we're at the end of the
+                    // `Request`/`Response`.
+                    if match packed_events.peek() {
```
pierrechevalier83: Thinking of it, don't these rules allow for the possibility of framing a node as an accomplice while hiding some of its gossip?

src/parsec.rs
```diff
@@ -488,9 +490,26 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         self.confirm_peer_state(src, PeerState::SEND)?;
 
         let mut prev_forking_peers = BTreeSet::new();
-        for packed_event in packed_events {
+
+        // Keep track of the earliest self_parent for each peer in packed_events as we will use
```
maqi: `Among the packed_events, keep track of each peer's earliest events' self_parent, as...` ?

src/parsec.rs
```diff
@@ -488,9 +490,26 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+        // them as bounds in the graph for where we look for malice accusations.
+        // NOTE: there is the assumption here that the events arrive in order.
+        #[cfg(feature = "malice-detection")]
+        let first_event_by_peer_in_packed_event =
+            collect_first_self_parents::<T, S>(&packed_events);
+
+        // As we process the events in linear "chunks", we keep track of the first event in each
```
maqi: It seems this `chunk` concept is newly introduced? is it same to a gossip message ?
if not, shall we explain it bit more?

src/parsec.rs
```diff
@@ -2013,6 +2238,23 @@ pub(crate) fn assert_same_events<T: NetworkEvent, S: SecretId>(a: &Parsec<T, S>,
     assert_eq!(a, b)
 }
 
+#[cfg(feature = "malice-detection")]
+fn collect_first_self_parents<T: NetworkEvent, S: SecretId>(
+    packed_events: &[PackedEvent<T, S::PublicId>],
+) -> BTreeMap<S::PublicId, EventHash> {
+    let mut events = BTreeMap::new();
+    packed_events
```
maqi: will
```
packed_events.iter().for_each(|event| {
    if let Some(hash) = event.self_parent() {
        let _ = events.entry(event.creator().clone()).or_insert(hash);
    }
});
```
be the same?

octol: It would seem so. Thanks I'll update!

src/parsec.rs
octol: I've replaced the peekable iterator with a calling `split(packed_events)`, which should hopefully make the concept simpler

src/parsec.rs
octol: As per discussion, converted to `UnprovableAccusation`

src/parsec.rs
octol: Updated!

src/gossip/event.rs
octol: Changed name

octol: Reworked in the latest version (using the new functions introduced in the Spam PR)

src/gossip/event.rs
```diff
@@ -153,12 +162,12 @@ impl<T: NetworkEvent, P: PublicId> Event<T, P> {
...
-            .last_ancestors
-            .get(other.as_ref().creator())
-            .map_or(false, |last_index| {
-                *last_index >= other.as_ref().index_by_creator()
-            })
+        self.sees_including_forks(other)
+    }
+
+    #[cfg(feature = "malice-detection")]
+    pub(crate) fn is_forking_peer(&self, peer: &P) -> bool {
```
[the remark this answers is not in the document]
octol: Done!

src/gossip/split_on_first.rs
```diff
@@ -0,0 +1,51 @@
...
+    P: FnMut(&T) -> bool,
+{
+    type Item = Vec<T>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.items.is_empty() {
+            return None;
+        }
+
+        let mut index = 1;
```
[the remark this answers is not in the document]
octol: Correct, that would be an identity transformation and potentially an infinite loop. But now that you point it out, I'll add some quick unit tests to this code. And also a comment.

src/functional_tests.rs
```diff
@@ -1013,15 +1013,14 @@ mod handle_malice {
...
-    #[ignore]
     // Carol received gossip from Bob, which should have raised an accomplice accusation against
     // Alice but didn't.
     fn accomplice() {
         let (invalid_accusation, alice) = create_invalid_accusation();
 
         let mut bob = TestParsec::from_parsed_contents(parse_dot_file_with_test_name(
             "bob.dot",
-            "parsec_functional_tests_handle_malice_accomplice",
+            "functional_tests_handle_malice_accomplice",
```
fizyk20: Is this necessary? Why not just use `parse_test_dot_file`, so that you don't have to supply the test name?

src/functional_tests.rs
```diff
@@ -1062,15 +1061,14 @@ mod handle_malice {
...
-    #[ignore]
     // Carol received `invalid_accusation` from Alice first, then received gossip from Bob,
     // which should have raised an accomplice accusation against Alice but didn't.
     fn accomplice_separate() {
         let (invalid_accusation, alice) = create_invalid_accusation();
 
         let mut carol = TestParsec::from_parsed_contents(parse_dot_file_with_test_name(
             "carol.dot",
-            "parsec_functional_tests_handle_malice_accomplice",
+            "functional_tests_handle_malice_accomplice",
```
fizyk20: I understand you want to reuse the dot graphs from the `accomplice` test. Wouldn't it make more sense to still use `parse_test_dot_file` without supplying the name, and just create a symlink to the folder with `accomplice` graphs? (Not sure how well git handles symlinks, though.) This way, if we wanted to use different graphs for some reason, we would only have to change the folder, not the code as well.

octol: Well I think git handles symlinks fine, but what about the Windows peeps?

fizyk20: Windows has symlinks, too (in NTFS, at least). I'm not sure if Git translates them correctly from the repo, though. I would expect it to, but I never tried.

PR #164, opened 2018-11-05
MAID-3251: feature gate malice detection

octol:
To allow easy integration with routing place malice detection behind a feature gate, that way latest master of parsec can easily be used.

### DIFF
--- a/src/dev_utils/mod.rs
+++ b/src/dev_utils/mod.rs
@@ -15,10 +15,10 @@ mod peer;
 pub mod proptest;
 mod schedule;
 
+#[cfg(all(test, feature = "malice-detection"))]
+pub(crate) use self::dot_parser::parse_dot_file_with_test_name;
 #[cfg(test)]
-pub(crate) use self::dot_parser::{
-    parse_dot_file_with_test_name, parse_test_dot_file, ParsedContents,
-};
+pub(crate) use self::dot_parser::{parse_test_dot_file, ParsedContents};
 pub use self::environment::{Environment, RngChoice};
 pub use self::network::Network;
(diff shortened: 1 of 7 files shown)


src/dev_utils/mod.rs
```diff
@@ -16,9 +16,10 @@ pub mod proptest;
 mod schedule;
 
 #[cfg(test)]
-pub(crate) use self::dot_parser::{
-    parse_dot_file_with_test_name, parse_test_dot_file, ParsedContents,
-};
+#[cfg(feature = "malice-detection")]
```
madadam: I think this should be `#[cfg(all(test, feature = "malice-detection"))]`.

src/gossip/packed_event.rs
```diff
@@ -35,6 +36,7 @@ impl<T: NetworkEvent, P: PublicId> Debug for PackedEvent<T, P> {
 }
 
 #[cfg(test)]
+#[cfg(feature = "malice-detection")]
```
madadam: You can squeeze the two attributes into `#[cfg(all(test, feature = "malice-detection"))]` if you want. I don't have a strong preference either way so I'll leave it up to you.

src/gossip/packed_event.rs
```diff
@@ -8,6 +8,7 @@
 
 use gossip::content::Content;
 #[cfg(test)]
+#[cfg(feature = "malice-detection")]
```
madadam: Very minor: Since you decided to collapse `test` and `feature = "malice-detection" into one attribute in that other place, it would be better to also do it here, for consistency.

PR #166, opened 2018-11-05
Fix invalid InvalidGossipCreator accusations

madadam:
This commit includes multiple changes:

1. Modify the testing framework to fail when a node raises invalid accusation
2. Modify the testing framework so a joining node always votes for all previous `Add`s and `Remove`s, in addition to `OpaquePayload`s
3. In `Parsec::from_existing`, do not initialise membership list of non-genesis peers. Instead, leave the initialization to when we receive their first event with other-parent.
4. When detecting `InvalidGossipCreator`, consider only the direct other-parent, instead of all ancestors.
5. Modify the `add_peers` integration test to start at 2 peers and add 8 more peers (for the total of 10) one by one. Also rename to `add_many_peers`.

### DIFF
--- a/src/dev_utils/network.rs
+++ b/src/dev_utils/network.rs
@@ -13,7 +13,7 @@ use block::Block;
 use error::Error;
 use gossip::{Request, Response};
 use mock::{PeerId, Transaction};
-use observation::Observation as ParsecObservation;
+use observation::{Malice, Observation as ParsecObservation};
 use parsec::is_supermajority;
 use std::collections::{BTreeMap, BTreeSet};
 
@@ -62,6 +62,11 @@ pub enum ConsensusError {
     TooFewSignatures {
         observation: Observation,
         signatures: BTreeSet<PeerId>,
+    },
+    InvalidAccusation {
+        accuser: PeerId,
+        accused: PeerId,
+        malice: Malice<Transaction, PeerId>,
     },
 }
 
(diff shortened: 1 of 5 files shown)


src/dev_utils/network.rs
```diff
@@ -290,6 +295,22 @@ impl Network {
         Ok(())
     }
 
+    /// Check that honest nodes have not been accused by other honest nodes.
```
pierrechevalier83: I don't see any enforcement of honesty on either side, which makes me think that this comment is misleading. Maybe this refers to the call site? (I may also be missing something)

tests/integration_tests.rs
```diff
@@ -233,12 +233,15 @@ fn random_schedule_no_delays() {
...
+fn add_many_peers() {
     let mut env = Environment::new(SEED);
+
     let schedule = Schedule::new(
         &mut env,
         &ScheduleOptions {
-            peers_to_add: 2,
+            genesis_size: 2,
+            peers_to_add: 8,
+            opaque_to_add: 0,
```
pierrechevalier83: I would be for adding a few opaque transactions too to be sure consensus in general still works when adding peers (i.e: it is not somehow specific to peer addition). That being said, I may change my mind if you tell me that it was very slow to run when having a few opaque payloads added to the mix.

madadam: I just wanted to create a minimal scenario where the `InvalidGossipCreator` big occurred and turned out it didn't even require opaque payloads. I'd actually be for keeping this test as is, and instead extending the `add_peers_and_vote` test to use more peers. Or maybe even writing a completely new test for it.

pierrechevalier83: BTW, one reason to ensure this is that opaque payloads may end up treated differently from transparent payloads with upcoming changes to customizable interesting_events.

PR #169, opened 2018-11-07
MAID-3254: Improve benchmarks

madadam:
This PR improves the benchmark so its results are reproducible (it's no longer based on randomness). For that purpose, it introduces `Record` which is a utility struct that can take an input DOT file and produce a sequence of actions (`vote_for`, `handle_request` and `handle_response`) which produced that DOT file. The sequence can then be replayed and it's performance measured.


benches/bench.rs
```diff
@@ -54,41 +55,20 @@
...
-                let mut env = Environment::new(RngChoice::SeededRandom);
-                let schedule = Schedule::new(&mut env, &ScheduleOptions::default());
-
-                (env, schedule)
-            },
-            |(mut env, schedule)| {
-                let _ = env.network.execute_schedule(schedule);
-            },
-        )
+        let record = unwrap!(Record::parse("input_graphs/benches_minimal/alice.dot"));
```
pierrechevalier83: Just a question: does the parsing time end up included in the benchmark? (That wouldn't be desirable)

madadam: No. The stuff outside `iter_with_setup` is executed only once and not included in the measurement. The first closure in `iter_with_setup` is executed every iteration, but not measured. Only the second closure is actually measured.

src/parsec.rs
```diff
@@ -212,6 +212,7 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         self.confirm_self_state(PeerState::VOTE)?;
 
         if self.have_voted_for(&observation) {
+            println!("{:?} duplicate vote {:?}", self.our_pub_id(), observation);
```
octol: Use the log framework stuff instead of printing?

Cargo.toml
```diff
@@ -43,3 +48,4 @@ required-features = ["mock"]
 [[bench]]
 name = "bench"
 harness = false
+required-features = ["testing"]
```
Fraser999: I prefer the approach we previously took since it provides a useful error message if trying to run just `cargo bench`.  In other words, I'd revert this and reapply the `fn main()` for the case where the `testing` feature isn't used.

PR #171, opened 2018-11-08
MAID-3253: update to handle dynamic membership

Fraser999:


examples/basic.rs
```diff
@@ -214,110 +298,346 @@ fn get_params() -> Params {
...
+    }
+
+    // Sets `self.new_peer` to a new `PeerId` with a random chance, if we still haven't created
+    // `add_peers_count`.
+    fn try_new_peer(&mut self) {
+        self.current_new_peer = if self.peers.len() < self.params.max_peer_count
+            && self.params.add_peers_count != self.peers_added_count
+            && self.rng.gen_weighted_bool(2)
+        {
+            Some(PeerId::new_with_random_keypair(
```
madadam: Why not use the `new_peer_id` function?

examples/basic.rs
```diff
@@ -214,110 +298,346 @@ fn get_params() -> Params {
...
                     .handle_response(&receiver_id, response)
             );
 
-            let peer = &mut peers[sender_index];
-            if peer.observations.len() < params.event_count && rng.gen_weighted_bool(3) {
-                rng.shuffle(&mut observations);
-                peer.vote_for_first_not_already_voted_for(&observations);
+            let peer = &mut self.peers[sender_index];
+            if peer.observations.len() < self.params.total_observations()
+                && self.rng.gen_weighted_bool(3)
```
madadam: I think it would be better to extract this magic number into named constant.

examples/basic.rs
```diff
@@ -335,32 +655,53 @@ fn main() {
...
         }
         println!();
+    }
 
-        peers.sort_by_key(|peer| usize::max_value() - peer.blocks.len());
-        let mut payloads = peers[0].blocks_payloads();
-        for peer in peers.iter().skip(1) {
+    fn validate_blocks_order(&mut self) {
+        self.peers
+            .sort_by_key(|peer| usize::max_value() - peer.blocks.len());
```
madadam: Very minor, but you can use `usize::MAX` instead.

examples/basic.rs
```diff
@@ -214,110 +298,346 @@ fn get_params() -> Params {
...
+
+    // Returns a randomly created new `PeerId`.
+    fn new_peer_id(&mut self) -> PeerId {
+        PeerId::new_with_random_keypair(
+            self.rng
+                .gen_ascii_chars()
+                .take(6)
+                .collect::<String>()
+                .as_str(),
+        )
```
madadam: I'd try to use the default `NAMES` (Alice, Bob, ...) first and only resort to random names when we exhaust them. Would make the log output nicer. But feel free to ignore this if it's too much hassle.

Fraser999: I'd prefer to leave that to a separate PR if we feel it's worthwhile.

examples/basic.rs
Fraser999: I forgot about this instance - good catch!

examples/basic.rs
Fraser999: The reason for preferring this is to avoid calling a function?  I.e. optimisation?

examples/basic.rs
```diff
@@ -176,26 +225,61 @@ fn get_params() -> Params {
              http://viz-js.com) to view the gossip graph.",
         ).set_term_width(100)
         .arg(
-            Arg::with_name(EVENTS_ARG_NAME)
-                .short("e")
-                .long(EVENTS_ARG_NAME)
-                .default_value("3")
+            Arg::with_name(MAX_PEER_COUNT_ARG_NAME)
+                .short("m")
```
pierrechevalier83: You may find it silly, but I would prefer to see this as a `const`: `MAX_PEER_COUNT_ARG_SHORTNAME` next to the other constant so if you ever rename it, it's harder to forget to update this to.

Of course, it applies to all the arguments.

If you're feeling fancy, maybe a struct with short_name and long_name would be even better.

examples/basic.rs
```diff
@@ -214,110 +298,346 @@ fn get_params() -> Params {
...
+        params.remove_peers_count = parse_usize(&matches, REMOVE_PEER_EVENTS_ARG_NAME);
+
+        println!("Running example with:");
+        println!("  max peer count: {}", params.max_peer_count);
+        println!("  initial peers: {}", params.genesis_peer_count);
+        println!("  peers to be added: {}", params.add_peers_count);
+        println!("  peers to be removed: {}", params.remove_peers_count);
+        println!("  random network events: {}", params.opaque_event_count);
+
+        if params.genesis_peer_count + params.add_peers_count
```
pierrechevalier83: I haven't read the rest yet so this may be an invalid concern, but depending on the order in which nodes are added and removed, this could not be sufficient to satisfy our invariants.

examples/basic.rs
Fraser999: I've added further `const`s rather than a struct.

examples/basic.rs
Fraser999: I'm aiming to avoid breaking our invariant (in terms of enough peers remaining online to retain consensus) with the `Environment::num_to_drop()` method.

examples/basic.rs
Fraser999: That would be my guess too (about the `const fn`), but I'm not too bothered about which to use.  The `usize::MAX` requires a `use std::usize` statement, which is why I normally go for the function instead.

PR #172, opened 2018-11-09
Fix dot parser: handle empty subgraphs and Add/Remove with related_info

madadam:


src/dev_utils/dot_parser.rs
```diff
@@ -183,14 +189,23 @@ struct ParsedSubgraph {
...
-    let data = id + parse_edge().repeat(0..) - spaces() - sym(b'}') - next_line()
-        + parse_edge().repeat(0..)
-        - next_line();
+    let id = whitespace()
+        * seq(b"style=invis")
+        * whitespace()
+        * seq(b"subgraph cluster_")
+        * parse_peer_id();
+
+    let edges1 = whitespace()
```
pierrechevalier83: can we `edges1` to `self_parents` and `edges2` to `other_parents`?

I know these names were used before within the lambda, but the scope was smaller. I would rename in both places, though if it's not too much of a hassle.

src/gossip/cause.rs
```diff
@@ -48,3 +52,34 @@ impl<T: NetworkEvent, P: PublicId> Display for Cause<T, P> {
...
+
+#[cfg(test)]
+impl Cause<Transaction, PeerId> {
+    pub(crate) fn new_from_dot_input(
+        input: CauseInput,
+        creator: &PeerId,
+        self_parent: Option<Hash>,
+        other_parent: Option<Hash>,
+    ) -> Self {
+        // For the dot file contains only partial graph, we have to manually change the info of
```
pierrechevalier83: Should it say: `When the dot file...` ?

pierrechevalier83: In retrospect, you just moved the code so you don't strictly have to fix this (but feel free to anyway: boy-scout-rule ;) )

PR #180, opened 2018-11-14
MAID-3259: only shorten debug hashes for dump-graphs

octol:
It's useful for clients like routing to have the whole hash printed by the Debug trait. Let's print the whole one except for when dumping graphs.

### DIFF
--- a/src/hash.rs
+++ b/src/hash.rs
@@ -7,12 +7,11 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use std::cmp::Ordering;
+use std::fmt::Display;
 use std::fmt::{self, Debug, Formatter};
 use tiny_keccak;
 
 pub const HASH_LEN: usize = 32;
-#[cfg(any(test, feature = "testing", feature = "dump-graphs"))]
-pub const HEX_DIGITS_PER_BYTE: usize = 2;
 
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
 pub struct Hash([u8; HASH_LEN]);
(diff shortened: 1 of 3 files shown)


src/hash.rs
```diff
@@ -54,11 +56,17 @@ impl<'a> From<&'a [u8]> for Hash {
...
             formatter,
             "{:02x}{:02x}{:02x}{:02x}{:02x}..",
             self.0[0], self.0[1], self.0[2], self.0[3], self.0[4]
         )
     }
+
+    #[cfg(not(feature = "dump-graphs"))]
+    fn fmt(&self, formatter: &mut Formatter) -> fmt::Result {
+        write!(formatter, "{:x}", self.0.iter().format(""))
+    }
```
madadam: I'd consider looking for ways to reuse this code in the `as_full_string` function in some way, to reduce code duplication.

madadam: One way could be something like this:

```rust
struct HashDisplay<'a>(&'a Hash);

impl Display for HashDisplay {
    fn fmt(&self, formatter: &mut Formatter) -> fmt::Result {
        write!(formatter, "{:x}", self.0.0.iter().format(""))
    }
}
```

Then `impl Debug` could just use:
```rust
fn fmt(&self, formatter: &mut Formatter) -> fmt::Result {
    write!(formatter, "{}", HashDisplay(self))
}
```

and `as_full_string`:
```rust
fn as_full_string(&self) -> String {
    format!("{}", HashDisplay(self))
}
```

src/hash.rs
```diff
@@ -54,11 +56,17 @@ impl<'a> From<&'a [u8]> for Hash {
...
         write!(
             formatter,
             "{:02x}{:02x}{:02x}{:02x}{:02x}..",
             self.0[0], self.0[1], self.0[2], self.0[3], self.0[4]
         )
     }
+
+    #[cfg(not(feature = "dump-graphs"))]
+    fn fmt(&self, formatter: &mut Formatter) -> fmt::Result {
+        write!(formatter, "{:x}", self.0.iter().format(""))
```
madadam: Would something like this work:
```rust
for byte in &self.0 {
    write!(formatter, "{:x}", byte)?;
}
```
? If yes, then you could possibly avoid bringing in the `itertools` dependency.

pierrechevalier83: This is a bit of a matter of taste, but I prefer the itertool version: it looks like the kind of code that the compiler could optimize better. (Not making the argument for performance but code that's clear to the compiler has my preference in general for esthetical reasons).

octol: I'll have go at it!

PR #186, opened 2018-11-15
Fix test that relies on hard coded hash

octol:
The test `unpolled_and_unconsensused_observations` reads in the hard coded hash for the first consensused block, which happens to be generated with the mock implementation. Sidestep this by restarting consensus and poll once if we're not using `--feature=mock.`

As a followup, enable `cargo test` without any feature flags on CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ before_script:
 script:
   - set -x;
     cargo fmt -- --check &&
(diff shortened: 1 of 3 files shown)


pierrechevalier83:
We decided to simply avoid running these tests without the mock feature to avoid spending resources on maintenance to the testing framework that won't have a real practical effect on the code.

PR #187, opened 2018-11-16
MAID-3074: add test for event with invalid self-parent

Fraser999:


src/parsec.rs
```diff
@@ -3167,5 +3168,46 @@ mod functional_tests {
...
+            let mut alice = Parsec::from_parsed_contents(parse_dot_file_with_test_name(
+                "alice.dot",
+                test_folder,
+            ));
+
+            // Try to add the event.
+            match alice.add_event(c_4) {
+                Err(Error::InvalidEvent) => (),
+                other => panic!("Expected Err(Error::InvalidEvent), but got {:?}", other),
+            }
```
octol: What about `assert_err!` like in PR 188?

Fraser999: Yup - good point.  Forgot about that macro.

PR #190, opened 2018-11-19
MAID-3098: handle premature gossip malice

Fraser999:


src/parsec.rs
```diff
@@ -511,7 +511,12 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
                     .change_peer_state(&event_creator, PeerState::RECV);
             }
         }
-        Ok(prev_forking_peers)
+
+        if self.confirm_self_state(PeerState::VOTE).is_ok() {
```
pierrechevalier83: I would move this to a function: `handle_premature_gossip`.
Possibly called from a function:
`self.detect_malice_after_processing_events()` (if we think there may end up being more than one malice checks here.

For the function, I think it's important as we could be looking for `handle_` and expect to see all kinds of malice listed.

Note: With this, we need to exclude `PrematureGossip` from the `Accomplice` check. @octol will be interested in reviewing this too.

Fraser999: I added a fixup commit for this.  A couple of points to note:
* I'd forgotten to make this a no-op for the case where the `malice-detection` feature isn't enabled
* I guess you meant "detect" rather than "handle"?
* I don't think there are any other cases which would justify creating `detect_malice_after_processing_events()`
* Your point about the `Accomplice` check doesn't require any further changes to this PR does it?  It relates purely to #163, yes?

src/parsec.rs
```diff
@@ -512,11 +512,10 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
             }
         }
 
-        if self.confirm_self_state(PeerState::VOTE).is_ok() {
-            Ok(prev_forking_peers)
-        } else {
-            Err(Error::PrematureGossip)
-        }
+        #[cfg(feature = "malice-detection")]
```
pierrechevalier83: Good point!

PR #193, opened 2018-11-20
Extract functional tests into a separate file

madadam:
This PR extracts the functional tests into a separate file. Because the tests no longer have access to private members of `Parsec`, a separate, test-only API is added. I decided for this approach over simply putting the test as submodule of `parsec` because it's more explicit and less prone to subtle errors.


src/functional_tests.rs
```diff
@@ -0,0 +1,1254 @@
...
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use block::Block;
+use dev_utils::parse_test_dot_file;
+use error::Error;
+use gossip::{Event, Graph, GraphSnapshot};
+use id::PublicId;
+use meta_voting::MetaElectionsSnapshot;
+use mock::PeerId;
+use mock::{self, Transaction};
```
Fraser999: The two `use mock::...`s could be unified.

src/functional_tests.rs
```diff
@@ -0,0 +1,1254 @@
...
+    assert_eq!(*event.creator(), our_id);
+    assert!(event.is_initial());
+}
+
+// TODO: remove this `cfg` once the `maidsafe_utilities` crate with PR 130 is published.
+#[cfg(feature = "testing")]
+#[test]
+#[should_panic(expected = "Genesis group can't be empty")]
+fn from_existing_requires_non_empty_genesis_group() {
+    use mock;
```
Fraser999: This and three other instances can be removed.

src/parsec.rs
```diff
@@ -1923,6 +1926,35 @@ impl<T: NetworkEvent, S: SecretId> Drop for Parsec<T, S> {
...
+                    created_by_us: false,
+                },
+            )
+        })
+    }
+}
+
+#[cfg(test)]
+impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
+    pub(crate) fn into_test(self) -> TestParsec<T, S> {
```
Fraser999: Implementing `From<Parsec>` for `TestParsec` might be slightly more canonical.  Not a big deal though.

src/parsec.rs
```diff
@@ -1971,1291 +1996,134 @@ impl Parsec<Transaction, PeerId> {
...
-            eric.peer_list.change_peer_state(peer_id, PeerState::RECV);
-        }
-
-        // Eric can no longer gossip to anyone.
-        assert_err!(
-            Error::InvalidSelfState { .. },
-            eric.create_gossip(Some(&PeerId::new("Alice")))
-        );
+#[cfg(test)]
+impl TestParsec<Transaction, PeerId> {
```
pierrechevalier83: I think it would make sense to add a `from_genesis` here, similar to `from_parsed_contents` to avoid the repetition of `from(from_genesis)`

src/functional_tests.rs
```diff
@@ -0,0 +1,1228 @@
...
+
+    // Eric can no longer gossip to anyone.
+    assert_err!(
+            Error::InvalidSelfState { .. },
+            eric.create_gossip(Some(&PeerId::new("Alice")))
+        );
+}
+
+#[test]
+fn unpolled_and_unconsensused_observations() {
```
octol: I sort of think we should feature gate this test behind `mock`, as currently it fails without it. That is, plain `cargo test` fails which I don't think sends a good signal

madadam: That sounds reasonable, but wasn't there a different PR addressing this issue?

octol: Yeah I started working on that last week but got interrupted by holiday :)
Was thinking I'll add it at some point soonish

madadam: I see. Then I would let you raise the fix as it is a bit outside of the scope of this PR.

pierrechevalier83: I agree about the scope.

PR #196, opened 2018-11-21
MAID-3095: Handle malice: Spam

madadam:


src/error.rs
```diff
@@ -7,63 +7,77 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use peer_list::PeerState;
+use std::fmt::{self, Display, Formatter};
 use std::result;
 
-quick_error! {
```
maqi: what's the reason of not using quick-error?

src/observation.rs
```diff
@@ -124,5 +127,101 @@ pub enum Malice<T: NetworkEvent, P: PublicId> {
     /// Event's creator is different to its self_parent's creator. The accusation contains the
     /// original event so other peers can verify the accusation directly.
     SelfParentByDifferentCreator(Box<PackedEvent<T, P>>),
+    /// Detectable but unprovable malice. Relies on consensus.
+    Unprovable(UnprovableMalice),
     // TODO: add other malice variants
 }
+
+impl<T: NetworkEvent, P: PublicId> Malice<T, P> {
+    #[cfg(any(test, feature = "malice-detection", feature = "testing"))]
```
maqi: the feature guard shall be just `malice-detection` ?
and those malice related tests shall have a feature guard of `malice-detection` ?

src/parsec.rs
```diff
@@ -203,8 +210,11 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
             consensused_blocks: VecDeque::new(),
             observations: BTreeMap::new(),
             meta_elections: MetaElections::new(genesis_group.clone()),
-            is_interesting_event,
             pending_accusations: vec![],
+            is_interesting_event,
+
```
maqi: we normally don't have this extra line break in the parameters list?

src/parsec.rs
```diff
@@ -231,7 +241,8 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
             &self.peer_list,
         );
 
-        self.add_event(event)
+        let _ = self.add_event(event)?;
```
maqi: hmm.  why now the result is no longer need to be returned?

src/parsec.rs
```diff
@@ -1898,6 +1922,45 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
             .map_err(|_| Error::PrematureGossip)
     }
 
+    fn detect_spam(&mut self, src: &S::PublicId, known_event_index: EventIndex) {
+        if self.unprovable_offenders.contains(src) {
```
maqi: Once a node accused a node for an `unprovable_offender`, it will never accuse it for again.
If there is many such types of malice, this line will block further accusation?
Also, if B spams event X to A, then event Y to C, then event Z to D ...  in a very long interval, 
to make the spam accusation consensused, the voting needs to track down the graph in a very deep depth ?

src/peer_list.rs
```diff
@@ -278,6 +278,32 @@ impl<S: SecretId> PeerList<S> {
...
     }
 
+    /// Returns the index of the last event gossiped to us by the given peer.
+    pub fn last_gossiped_event_by(&self, peer_id: &S::PublicId) -> Option<EventIndex> {
+        self.peers
+            .get(peer_id)
+            .and_then(|peer| peer.last_gossiped_event)
+    }
+
+    /// Note that the given peer gossiped to us the given event.
```
maqi: the comment is bit confusing to me. do you want to say `record the last event a peer gossiped to us` ?

src/dev_utils/network.rs
```diff
@@ -63,8 +63,10 @@ pub enum ConsensusError {
         observation: Observation,
         signatures: BTreeSet<PeerId>,
     },
-    InvalidAccusation {
-        accuser: PeerId,
+    UnexpectedAccusation {
+        // `Some(peer_id)` means that the accusation has been raised by `peer_id`.
+        // `None` means that the accusation has been raised by multiple peers and consensused.
```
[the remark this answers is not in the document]
madadam: 1. Because we currently don't expect any accusations to be neither raised nor consensused in the integration tests. That is, not until the malice testing task is implemented in which case we would only expect accusation against a node that actually committed malice.

2. The test panics when at least one node raised at least one unexpected accusation, so in case multiple nodes raise it, we panic on the first one.

Also I think it's worth elaborating why I even made these modifications: I didn't want the tests to fail if someone raises `Spam`, because there is a chance of honest false positive. That is, a honest node can send a gossip to another honest node and still be accused of spam. It's very unlikely that this would happen, but not impossible. A consensused accidental `Spam`, on the other hand, shouldn't happen ever, so that's why I panic on it.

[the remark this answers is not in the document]
madadam: But thinking about it, this might be overkill. I think I'm going to revert the part that panics on consensused unexpected accusations.

src/error.rs
madadam: I initially thought I would have to make `Error` generic, which is something `quick_error` doesn't support. In the end I didn't have to do it, but the change was already made and I think it's not that much worse than using `quick_error` and having one less dependency is kind of nice. I don't mind reverting this change though.

src/observation.rs
madadam: `is_provable` is used in `Network` too, so it needs to be `feature = "testing"` too. Regarding those tests - problem is `Malice` is part of `Observation` even if `malice-detection` is not enabled, so those tests are still valid. 

Maybe I would make sense to put `Malice` and the `Observation::Accusation` variant under `malice-detection` too, but that is outside of the scope of this PR.

src/parsec.rs
madadam: It is returned. The `?` operator propagates errors.

src/parsec.rs
madadam: So the thing is that all the types of unprovable malice should be treated as if it were the same type (the `UnprovableMalice` enum is for internal diagnostic purposes only). So allowing a node to raise multiple unprovable malices against the same node would result in the node being accused of duplicate vote. The `unprovable_offenders` is there to prevent that.

Now you have a valid point though and it needs to be addressed, but I think it requires a separate discussion and a separate PR.

src/peer_list.rs
madadam: Yeah, it sounds weird. I'm going to rename it to `record_` instead of `note_`.

src/error.rs
maqi: quick_error could save us some manual work, if later on there is more error type needs to be added.
I don't have a strong preference here, so leave it to others to vote.  :)

pierrechevalier83: I would have add a minor preference for quick_error, but happy to close an eye as both are fine, so the implementer's preference should prevail :smile:

src/parsec.rs
```diff
@@ -1820,19 +1826,16 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-            // Unprovable accusations are never invalid.
-            if let Malice::Unprovable = *malice {
-                return;
-            } else {
+        let their_accusation = match event.vote().map(Vote::payload) {
+            Some(&Observation::Accusation {
+                ref offender,
+                ref malice,
+            })
+                if malice.is_provable() =>
```
pierrechevalier83: Did we mean to change the behaviour here and avoid returning directly when the malice isn't provable?

src/parsec.rs
```diff
@@ -1932,11 +1973,34 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+        event: IndexedEventRef<'a, T, S::PublicId>,
+        creator: &S::PublicId,
+    ) -> Option<IndexedEventRef<'a, T, S::PublicId>> {
+        use gossip::LastAncestor;
+
+        match event.last_ancestor_by(creator) {
+            LastAncestor::Some(index) => self
+                .peer_list
+                .events_by_index(creator, index)
+                .next()
```
pierrechevalier83: I'm not understanding this `next`... I would think `Event::last_ancestor_by` would give you the index you're looking for already... I guess, can you explain the reasoning here? (And maybe in a comment too)

src/gossip/event.rs
```diff
@@ -388,6 +404,23 @@ impl Event<Transaction, PeerId> {
...
+pub(crate) enum UnpackedEvent<T: NetworkEvent, P: PublicId> {
+    // Event is already in our gossip graph
+    Known(EventIndex),
+    // Event is not yet in our gossip graph
+    New(Event<T, P>),
+}
+
+pub(crate) enum LastAncestor {
+    // There are no forks and the ancestor exists.
+    Some(usize),
```
maqi: wondering why it's not
```
enum LastAncestor {
    Option<usize>,
    Fork,
}
```
?

src/parsec.rs
madadam: `Event::last_ancestor_by` gives you the `index-by-creator`, but `PeerList::events_by_index` returns an iterator, because it doesn't know that there is just one event. So, logically, there can be only one event at that index in this case, but we can't prove that to the type system.

I can always use `unwrap` though :smile:

pierrechevalier83: Ah right!. It's almost like we could make an `event_by_index` function that does `events_by_index(...).next()` and that would clarify this. In the flow, it seemed weird to me...

Maybe not worth a new round of CI, though... I'll let you decide of this.

madadam: `event_by_index` which returns `Some` only if there is exactly one event might be useful. Maybe named `unique_event_by_index` for more explicitness.

src/gossip/event.rs
madadam: That's not valid syntax. It would have to be something like:

```rust
enum LastAncestor {
    NoFork(Option<usize>),
    Fork,
}
```

Which would be a valid way to do it, but I don't see any obvious advantage to it.

PR #197, opened 2018-11-22
MAID-3267: Add method to get peers we can gossip to

madadam:


src/peer_list.rs
```diff
@@ -55,6 +55,23 @@ impl<S: SecretId> PeerList<S> {
...
+                .iter()
+                .filter(move |(id, peer)| {
+                    *id != self.our_pub_id() && peer.state.can_vote() && peer.state.can_recv()
+                }).map(|(id, _)| id);
+            Some(ids)
+        } else {
+            None
+        };
+
+        ids.into_iter().flat_map(|ids| ids)
```
pierrechevalier83: Can't we do away with this step by removing the Option, removing the semi-column in the first branch and adding `std::iter::empty()` in the other branch?

My guess is you tried this and had an issue with the types; but can you explain what the pb is?

madadam: The type in both branches needs to be the same if you want to return `impl Trait`, so what you are suggesting wouldn't work. Alternative would be to return `Box<Iterator<...>>` or collect the ids into a collection, both of which would require allocation.

PR #198, opened 2018-11-22
MAID-3265: Use enum instead of function to determine interesting payloads

madadam:


src/parsec.rs
```diff
@@ -911,51 +918,52 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
     fn is_interesting_payload(
         &self,
         builder: &MetaEventBuilder<T, S::PublicId>,
-        valid_voters: &BTreeSet<S::PublicId>,
+        peers_that_can_vote: &BTreeSet<S::PublicId>,
```
octol: I like these small improvements to variable names

src/parsec.rs
```diff
@@ -27,22 +27,29 @@ use std::ops::{Deref, DerefMut};
...
+    /// One vote is enough.
+    Single,
+    /// Supermajority (more than 2/3) is required.
+    Supermajority,
+}
+
+impl ConsensusMode {
+    pub(crate) fn check(self, did_vote: usize, can_vote: usize) -> bool {
+        match self {
+            ConsensusMode::Single => did_vote > 0,
```
pierrechevalier83: Are we sure the nodes that `did_vote` were in a situation where their vote should be considered? I was under the impression that taking the intersection with `can_vote` as we used to was important.

madadam: The intersection was unnecessary, because the `peers_that_did_vote` was already a subset of `peers_that_can_vote`. Look at this code and it should be clear: https://github.com/maidsafe/parsec/blob/9a24c9392976a86da38d71b2158c3c368e71d8f4/src/parsec.rs#L945-L967

PR #199, opened 2018-11-23
MAID-3268 test/parsec: test gossip_recipients

maqi:


src/dev_utils/peer.rs
```diff
@@ -114,6 +114,11 @@ impl Peer {
                 _ => None,
             })
     }
+
+    /// Returns ids of peers who we can send gossips to.
+    pub fn gossip_recipients(&self) -> impl Iterator<Item = &PeerId> {
```
Fraser999: Since `gossip_recipients()` is public anyway, and this is a pure wrapper, I don't think it's useful.  We can just call `peer.parsec.gossip_recipients()`.

src/dev_utils/network.rs
```diff
@@ -403,9 +403,18 @@ impl Network {
...
-                            | Err(Error::InvalidSelfState { .. }) => (),
-                            Err(e) => panic!("{:?}", e),
+                            Err(e) => match e {
+                                Error::InvalidPeerState { .. } | Error::InvalidSelfState { .. } => {
+                                    if self
+                                        .peer(&peer)
+                                        .gossip_recipients()
+                                        .any(|peer_id| peer_id == &req.recipient)
+                                    {
+                                        panic!("{:?}", e);
```
Fraser999: I think it'd help to either have a code comment explaining why we're panicking in this case, or maybe even better to have it in the actual panic message?

Also, rather than two levels of `match`, wouldn't it be cleaner to stick to one level:
```rust
Ok(request) => { ... }
Err(e @ Error::InvalidPeerState { .. }) | Err(e @ Error::InvalidSelfState { .. }) => { ... }
Err(e) => panic!("{:?}", e),
```

PR #200, opened 2018-11-23
MAID-3266: Excess votes

madadam:
WIP because tests are failing.

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -19,24 +19,27 @@ use vote::Vote;
 pub struct Block<T: NetworkEvent, P: PublicId> {
     payload: Observation<T, P>,
     proofs: BTreeSet<Proof<P>>,
+    excess: bool,
 }
 
 impl<T: NetworkEvent, P: PublicId> Block<T, P> {
     /// Creates a `Block` from `payload` and `votes`.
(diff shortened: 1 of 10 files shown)


src/block.rs
```diff
@@ -49,6 +52,14 @@ impl<T: NetworkEvent, P: PublicId> Block<T, P> {
         &self.proofs
     }
 
+    /// Is this an "excess" block?
+    /// Excess block is a block signed by less than the required number of voters.
+    /// Excess block can only be retrieved from `Parsec` if a regular block with the same payload
```
maqi: the task description mentioned:
```
poll should return an Option<ConsensusOutput> where ConsensusOutput is an enum that can represent
* A block made valid
* A vote for a payload reaching consensus for the first time
* A subsequent vote for an already consensused payload
```
Which, I I think it will be more naturally to have the follow-up votes to be separated from the `block` concept ?

madadam: I feel like the difference between what I've implemented and what you are suggesting is mostly cosmetic. I'm not opposed to it though. We should probably implement it in the way most useful for routing integration. That said, I'm currently encountering test failures which seem to point to some deeper issue than this. I'll come back to this once I deal with those failures.

madadam:
Closing. Partially superseded by #205 .

PR #201, opened 2018-11-26
updating cargo commands

u-amine:
tutorial refers to older versions

### DIFF
--- a/tutorial.md
+++ b/tutorial.md
@@ -30,7 +30,7 @@ From the parsec repository, run the tests. Be sure to use the feature: `dump-gra
 
 ```
 cd parsec
-cargo test --release --features=dump-graphs -- --nocapture
+cargo test --release --features=mock -- --nocapture
 ```
 
 # View the graphs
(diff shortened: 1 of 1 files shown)


tutorial.md
```diff
@@ -30,7 +30,7 @@ From the parsec repository, run the tests. Be sure to use the feature: `dump-gra
 
 ```
 cd parsec
-cargo test --release --features=dump-graphs -- --nocapture
```
pierrechevalier83: We should probably still recommend dumping the graphs by default, so `--features=dump-graphs,mock` would make more sense (same for two other places)

tutorial.md
```diff
@@ -89,12 +89,12 @@ Now, `test_minimal_network` is a silly example as reaching consensus on a single
...
 For more control over the scenario, please run the example. To see which options you may use, run
 ```
-cargo run --release --example basic --features=dump-graphs -- --help
+cargo run --release --example basic --features=mock -- --help
 ```
 
 For instance, for an example with 10 peers agreeing on the order of 5 events, run
 ```
-cargo run --release --example basic --features=dump-graphs -- -p10 -e5
+cargo run --release --example basic --features=mock -- -p10 -e5
```
pierrechevalier83: Seing this, I'm realising that the arguments that the example takes changed too.
```
cargo run --release --example basic --features=dump-graphs,mock -- -i10 -o5
```
would be the equivalent to this. Could I ask you to address it in the same PR?

PR #205, opened 2018-11-28
MAID-3266: Support single-vote blocks

madadam:
This PR implements one part of MAID-3266: it makes it so that when we are in the `ConsensusMode::Single`, each voted opaque payload is outputted in its own block signed only by the single voter.

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -23,11 +23,11 @@ pub struct Block<T: NetworkEvent, P: PublicId> {
 
 impl<T: NetworkEvent, P: PublicId> Block<T, P> {
     /// Creates a `Block` from `votes`.
-    pub fn new(votes: &BTreeMap<P, Vote<T, P>>) -> Result<Option<Self>, Error> {
+    pub fn new(votes: &BTreeMap<P, Vote<T, P>>) -> Result<Self, Error> {
         let payload = if let Some(vote) = votes.values().next() {
             vote.payload().clone()
         } else {
-            return Ok(None);
+            return Err(Error::MissingVotes);
         };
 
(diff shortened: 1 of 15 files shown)


src/dev_utils/network.rs
```diff
@@ -221,26 +227,48 @@ impl Network {
         Ok(())
     }
 
+    fn block_key<'a>(
+        &self,
+        block: &'a Block<Transaction, PeerId>,
+    ) -> (&'a Observation, Option<&'a PeerId>) {
+        let peer_id = if block.payload().is_opaque() {
```
maqi:
```
    if block.payload().is_opaque() && self.consensus_mode == ConsensusMode::Single {
        Some(&unwrap!(block.proofs().into_iter().next()).public_id)
    } else {
        None
    };
````

File: src/block.rs [hunk 2]
```diff
@@ -22,21 +22,26 @@ pub struct Block<T: NetworkEvent, P: PublicId> {
 }
 
 impl<T: NetworkEvent, P: PublicId> Block<T, P> {
-    /// Creates a `Block` from `payload` and `votes`.
-    pub fn new(payload: Observation<T, P>, votes: &BTreeMap<P, Vote<T, P>>) -> Result<Self, Error> {
-        let proofs: BTreeSet<Proof<P>> = votes
+    /// Creates a `Block` from `votes`.
```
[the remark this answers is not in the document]
madadam: I'll probably make it an error. It'd be clearer.

src/dev_utils/network.rs
madadam: :+1:

PR #206, opened 2018-11-29
MAID-3264: Use human readable peer names in basic example

octol:
Random names for the peers can cause issues when visualising the generated dot file. Pick human readable names if available to mitigate it.

Also pads `NAMES` with a few more entries. Wasn't sure which names to use, so I just checked a baby name website!

### DIFF
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -434,6 +434,7 @@ struct Environment {
     current_remove_peers: Vec<PeerId>,
     current_new_peer: Option<PeerId>,
     current_round: usize,
+    pretty_id_count: usize,
 }
 
 impl Environment {
@@ -456,6 +457,7 @@ impl Environment {
             current_remove_peers: vec![],
             current_new_peer: None,
             current_round: 0,
+            pretty_id_count: 0,
         };
 
         // Set up the requested number of peers and random network events.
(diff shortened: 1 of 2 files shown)


examples/basic.rs
```diff
@@ -456,6 +457,7 @@ impl Environment {
             current_remove_peers: vec![],
             current_new_peer: None,
             current_round: 0,
+            peers_added: 0,
```
Fraser999: I think it might help to try and name this to indicate that it represents the number of human-readable IDs generated as opposed to actual peers added.  Something like `pretty_id_count` maybe?

examples/basic.rs
```diff
@@ -478,13 +480,19 @@ impl Environment {
...
     // Returns a randomly created new `PeerId`.
     fn new_peer_id(&mut self) -> PeerId {
-        PeerId::new_with_random_keypair(
-            self.rng
-                .gen_ascii_chars()
-                .take(6)
-                .collect::<String>()
-                .as_str(),
-        )
+        let peer = if let Some(peer) = PeerId::from_index(self.peers_added) {
```
Fraser999: This could be just:
```rust
PeerId::from_index(self.peers_added).unwrap_or_else(|| {
    PeerId::new_with_random_keypair(...)
})
```
And if the `peers_added` increment happens first, we don't need the temp variable `peer`.

src/mock.rs
```diff
@@ -78,6 +79,13 @@ impl PeerId {
             initial, NAMES
         );
     }
+
+    pub fn from_index(peer_index: usize) -> Option<Self> {
+        if peer_index >= NAMES.len() {
```
Fraser999: This could be simplified to:
```rust
NAMES.get(peer_index).map(|name| PeerId::new(name))
```

octol: Hey that's a solid improvement!

examples/basic.rs
octol: Note though that would mean indexing using `pretty_id_count - 1`. Like the functional construct though!

PR #208, opened 2018-11-30
MAID-3080/3083: Detect invalid self/other parent

octol:
When the hash of the other_parent doesn’t point to any known gossip event, `handle_request` and `handle_response` should return `Error<UnknownOtherParent>` or `Error<UnknownSelfParent>`.

Note in the original task `InvalidOtherParent`/`InvalidSelfParent` was specified.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -37,8 +37,10 @@ pub enum Error {
     },
     /// The given event is invalid or malformed.
     InvalidEvent,
-    /// This event's self-parent or other-parent is unknown to our node.
-    UnknownParent,
+    /// The event's self-parent is unknwon to our node.
+    UnknownSelfParent,
+    /// The event's other-parent is unknown to our node.
+    UnknownOtherParent,
     /// Our node has already voted for this network event.
     DuplicateVote,
     /// The peer sent a message to us before knowing we could handle it.
@@ -71,10 +73,12 @@ impl Display for Error {
                 required, actual
             ),
             Error::InvalidEvent => write!(f, "The given event is invalid or malformed."),
-            Error::UnknownParent => write!(
-                f,
-                "This event's self-parent or other-parent is unknown to this node."
-            ),
+            Error::UnknownSelfParent => {
+                write!(f, "The event's self-parent is unknown to this node.")
+            }
+            Error::UnknownOtherParent => {
+                write!(f, "The event's other-parent is unknown to this node.")
+            }
             Error::DuplicateVote => write!(f, "Our node has already voted for this network event."),
             Error::PrematureGossip => write!(
                 f,
--- a/src/gossip/event.rs
+++ b/src/gossip/event.rs
@@ -519,7 +519,7 @@ fn get_parent<'a, T: NetworkEvent, P: PublicId>(
                 .and_then(|index| graph.get(index))
                 .ok_or_else(|| {
                     debug!("{:?} missing {} parent for {:?}", our_id, parent, content);
-                    Error::UnknownParent
+                    parent.to_unknown_error()
                 })?,
         ))
     } else {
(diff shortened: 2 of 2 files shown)


src/parsec.rs
```diff
@@ -298,6 +298,8 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
             self.our_pub_id(),
             src
         );
+        #[cfg(feature = "malice-detection")]
+        self.detect_invalid_other_parent(&req.packed_events)?;
```
madadam: You can move this call to `unpack_and_add_events` and avoid having to duplicate it in `handle_response`.

src/parsec.rs
```diff
@@ -1990,6 +1994,22 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
             .map_err(|_| Error::PrematureGossip)
     }
 
+    fn detect_invalid_other_parent(
```
madadam: We already check if parents are valid in `Event::unpack`. Maybe just modify the code there to distinguish between invalid self-parent and invalid other-parent (currently we raise `Error::UnknownParent` for both). Should be simpler and would avoid duplicating the detection logic.

madadam: This is where the detection currently take place: https://github.com/maidsafe/parsec/blob/master/src/gossip/event.rs#L522

src/error.rs
```diff
@@ -41,6 +41,8 @@ pub enum Error {
     DuplicateVote,
     /// The peer sent a message to us before knowing we could handle it.
     PrematureGossip,
+    /// The event received has an unknown other-parent.
+    InvalidOtherParent,
```
madadam: We already have `UnknownParent`. I'd try to use consistent naming (`Unknown` vs `Invalid`).

src/parsec.rs
madadam: But you can ignore this comment and address this https://github.com/maidsafe/parsec/pull/208#discussion_r237908004 instead.

octol: The thought did indeed hit me, but I ended up putting in them both just because the task speciefied that. But then again I probably shouldn't be that rigid :)

src/parsec.rs
octol: Right, that's a good point! I'll have a go at moving it

src/parsec.rs
```diff
@@ -1990,6 +1994,22 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+        &self,
+        packed_events: &[PackedEvent<T, S::PublicId>],
+    ) -> Result<()> {
+        let packed_event_hashes = packed_events.iter().map(|pe| pe.hash()).collect::<Vec<_>>();
+
+        if packed_events
```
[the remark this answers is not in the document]
octol: I'm thinking `packed_events` is probably 10-100 most of the time, and occasionally (when a node joins) a lot larger

src/error.rs
```diff
@@ -37,12 +37,14 @@ pub enum Error {
     InvalidEvent,
     /// This event's self-parent or other-parent is unknown to our node.
     UnknownParent,
```
madadam: I think this variant can be removed now.

octol: You think it's preferable to not `map_err` the `Error`, instead returning the precise one directly? (Currently `UnknownParent` is still used). Would that mean creating two `get_self_parent` `get_other_parent` functions?

madadam: Well, `UnknownParent` isn't needed anymore. It's now used only internally to dispatch to `UnknownSelfParent` or `UnknownOtherParent`, but that can be done by other means. For example the one you suggested.

src/gossip/event.rs
```diff
@@ -519,7 +532,7 @@ fn get_parent<'a, T: NetworkEvent, P: PublicId>(
                 .and_then(|index| graph.get(index))
                 .ok_or_else(|| {
                     debug!("{:?} missing {} parent for {:?}", our_id, parent, content);
-                    Error::UnknownParent
+                    error_type
```
pierrechevalier83: Call it nitpick, but I would prefer to see a function: `unknown_error` that takes a `Parent` enum and maps it to the appropriate error.
It would make for an easier to call `get_parent` as the ownace wouldn't be on the caller to keep the type of `Parent` and the `error_type` in sync...

PR #209, opened 2018-12-03
MAID-3272: Ensure syncs have correct other-parent

Fraser999:


src/gossip/messages.rs
```diff
@@ -39,4 +45,25 @@ impl<T: NetworkEvent, P: PublicId> Response<T, P> {
...
         }
     }
+
+    pub(crate) fn hash_of_last_event_created_by(&self, src: &P) -> Result<Option<EventHash>> {
+        hash_of_last_event_created_by(src, &self.packed_events)
+    }
+}
+
+// Returns `Err(Error::InvalidMessage)` if `packed_events` is non-empty, but doesn't contain an
+// event created by `creator`.
```
maqi: the `creator` here is actually the `src` passed from the req/resp's function call.
Shall we make it consistent above or change the comment/variable name here?

src/parsec.rs
```diff
@@ -1557,6 +1559,7 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         src: &S::PublicId,
         is_request: bool,
         forking_peers: &BTreeSet<S::PublicId>,
+        opt_other_parent: Option<EventHash>,
```
maqi: some comment to explain why this is required?

PR #210, opened 2018-12-03
Fix and optimize the test framework + fix subtle bug in Parsec

madadam:
1. Fix the `Parsec::previous_interesting_content` function to return `None` in case the previous interesting content is empty. In that case we must perform the full computation, because there can be new interesting content that wasn't present in the previous meta-election.
2. Make `PeerRemovalGuard` more robust to handle edge cases.
3. Lower the gossip probability to speed the tests up, but increase the `adjustment_coeff` correspondingly, to make sure they still pass. Overall significant test performance increase.
4. Remove `prob_local_step` as it wasn't really used anymore (local step now happens on every step).
5. Some minor refactoring of the test framework

### DIFF
--- a/src/dev_utils/network.rs
+++ b/src/dev_utils/network.rs
@@ -231,6 +231,49 @@ impl Network {
         }
     }
 
+    fn send_gossip<R: Rng>(
+        &mut self,
+        rng: &mut R,
+        options: &ScheduleOptions,
+        sender: &PeerId,
+        present_peers: &[PeerId],
+        step: usize,
+    ) {
+        let recipient = loop {
+            let recipient = unwrap!(rng.choose(present_peers));
+            if recipient != sender {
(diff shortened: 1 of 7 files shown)


src/dev_utils/network.rs
```diff
@@ -544,34 +534,115 @@ impl Network {
...
+// Helper struct that protects the test network from losing too many nodes.
 #[derive(Default)]
 struct PeerRemovalGuard {
-    voters: BTreeSet<PeerId>,
-    nodes_to_be_removed: BTreeSet<PeerId>,
+    peers: BTreeMap<PeerId, PeerRemovalState>,
+}
+
+enum PeerRemovalState {
+    Adding(usize),
```
maqi: better have comment here to explain the `usize` is for the count of voters

PR #214, opened 2018-12-07
Fix several issues in the test framework

madadam:
1. During schedule generation, votes are now cast by all peers, not just by the active ones. This is because the status of a peer during schedule generation might not necessarily match its status during execution (because of `PeerRemovalGuard`). Removed and failed peers are skipped when actually casting votes durging execution.

2. In single-vote mode, any votes cast by peers that are scheduled for removal or fail anywhere in the schedule are not guaranteed to be consensused. The framework is updated to reflect this.

3. Re-queueing of schedule events that failed to be executed is changed to prevent single failed event to be tried over and over until the schedule is exhausted.

### DIFF
--- a/src/dev_utils/network.rs
+++ b/src/dev_utils/network.rs
@@ -458,28 +458,21 @@ impl Network {
             events,
         } = schedule;
         let mut peer_removal_guard = PeerRemovalGuard::default();
-        let mut pending_events: VecDeque<ScheduleEvent> = VecDeque::new();
+        let mut queue: VecDeque<_> = events.into_iter().collect();
+        let mut retry = Vec::new();
 
-        for event in events {
-            while let Some(pending_event) = pending_events.pop_front() {
-                if !self.execute_event(
-                    rng,
-                    options,
-                    pending_event.clone(),
-                    &mut peer_removal_guard,
-                )? {
-                    pending_events.push_front(pending_event);
+        while let Some(event) = queue.pop_front() {
+            if self.execute_event(rng, options, event.clone(), &mut peer_removal_guard)? {
+                for event in retry.drain(..).rev() {
+                    queue.push_front(event)
+                }
+
+                self.check_consensus_broken()?;
+                if self.consensus_complete(&peers, max_observations) {
                     break;
                 }
-            }
-
-            if !self.execute_event(rng, options, event.clone(), &mut peer_removal_guard)? {
-                pending_events.push_back(event);
-            }
-
-            self.check_consensus_broken()?;
-            if self.consensus_complete(&peers, max_observations) {
-                break;
+            } else {
+                retry.push(event);
             }
         }
 
@@ -573,6 +566,12 @@ impl Network {
                 }
             }
             ScheduleEvent::VoteFor(peer, observation) => {
+                // Skip voting by removed/failed peers.
+                match self.peer(&peer).status {
+                    PeerStatus::Active | PeerStatus::Pending => (),
+                    PeerStatus::Removed | PeerStatus::Failed => return Ok(true),
+                }
+
                 if let ParsecObservation::Remove { ref peer_id, .. } = observation {
                     if !peer_removal_guard.attempt_to_remove_peer(&peer_id) {
                         return Ok(false);
@@ -592,6 +591,7 @@ struct PeerRemovalGuard {
     states: BTreeMap<PeerId, PeerRemovalState>,
 }
 
+#[derive(Debug)]
 enum PeerRemovalState {
     // Peer is being added. The `usize` is the number of peers that consensused the add.
     Adding(usize),
--- a/src/dev_utils/schedule.rs
+++ b/src/dev_utils/schedule.rs
@@ -87,6 +87,7 @@ pub struct PendingObservations {
     max_delay: usize,
     p_delay: f64,
     queues: BTreeMap<PeerId, BTreeMap<usize, Vec<Observation>>>,
+    opaque_vote_counts: BTreeMap<PeerId, usize>,
 }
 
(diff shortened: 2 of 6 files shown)


src/dev_utils/schedule.rs
```diff
@@ -606,14 +599,25 @@ impl Schedule {
         // Gossip should theoretically complete in O(log N) steps
         // The constant (adjustment_coeff) is for making the number big enough.
         let n = peers.present_peers().count() as f64;
-        let adjustment_coeff = 750.0;
+        let adjustment_coeff = 5000.0;
```
octol: Pretty large increase it seems?

madadam: Yes, as per a discussion with @Fraser999 and @pierrechevalier83 . We decided to set this number to high enough so the likelihood of a test failing because there are not enough schedule events is minimal. It should not matter much for well behaving tests, because those finish way before this number if hit. It's only if we have a bug somewhere we might hit this number.

madadam:
What's unreasonable about the commit message line lengths?

octol:
> What's unreasonable about the commit message line lengths?

Well the first line is 346 character long. I guess most tools can soft wrap so it's nothing major. Is there something about this in the git style guide?

src/dev_utils/schedule.rs
octol: Sorry what I meant is that it's probably fine, was just double checking that it was intentional since it was such substantial change. Also is this change related to point 2. in the commit message or an additional tweak?

src/dev_utils/network.rs
```diff
@@ -468,7 +468,7 @@ impl Network {
                     pending_event.clone(),
                     &mut peer_removal_guard,
                 )? {
-                    pending_events.push_front(pending_event);
+                    pending_events.push_back(pending_event);
```
maqi: the `pending_event` that not executed shall be put back to the front to retain the order.

src/dev_utils/network.rs
```diff
@@ -573,6 +573,12 @@ impl Network {
                 }
             }
             ScheduleEvent::VoteFor(peer, observation) => {
+                // Skip voting by removed/failed peers.
+                match self.peer(&peer).status {
+                    PeerStatus::Active | PeerStatus::Pending => (),
+                    PeerStatus::Removed | PeerStatus::Failed => return Ok(true),
```
maqi: I am not sure about this.
Say a peer is removed from the network perspective, will there be chance it still send out vote to others, and the others shall handle this situation?

src/dev_utils/network.rs
madadam: See point 3 in the commit message. With `push_front`, I sometimes got tests failure. For example, try to change it back to `push_front` and use this seed: `[3085689988, 3474909085, 3387560573, 166421199]`.

madadam: And to elaborate a but: what I think might happen with `push_front` is a scenario like this:

Let's say `pending_events` contains `Remove(Gina)` and `Add(Hank)`. We pop `Remove(Gina)` and try to execute it, but it fails due to the `PeerRemovalGuard`, so we push it back to front. We process the next regular event (which is `LocalStep`) and try `Remove(Gina)` again. Problem is it will never succeed, because `PeerRemovalGuard` will keep rejecting it. It can never succeed until `Add(Hank)` is processed first, but because we keep returning `Remove(Gina)` to the front, we never get to that. Eventually we exhaust all `LocalStep`s and the test fails.

src/dev_utils/network.rs
madadam: No, when a peer is removed, it will never send anything or vote for anything anymore.

src/dev_utils/network.rs
maqi: I don't think the case you described and the failure fixed on the seed is a valid reason.
In your example scenario, the `PeerRemovalGuard ` shall `pass` the removal eventually when the on-going removal votes completed.
The failure (with that seed) is fixed after this change is just because the `order of events` is changed.

madadam: Say we have 3 active peers (Alice, Bob, Carol), and schedule events `Remove(Carol)` and `Add(Dave)`. With the `push_front` approach, such scenario would never succeed. `Remove(Carol)` would be forever blocked by `PeerRemovalGuard`, and `Add(Dave)` would never get its turn. 

If we instead allowed `Add(Dave)` to be processed first, the test would pass, because we would get to `{Alice, Bob, Carol, Dave}` first which would unblock removal of Carol.

Note also that the scheduled order does not necessarily guarantee consensus order, so swapping the two event is not an error.

maqi: In your above example , `Remove(Carol)` shall not be scheduled before `Add(Dave)` at beginning.
If it does, it shall mean an error in the testing setup. Not we shall adjust it in the testing execution.

Also, `PeerRemovalGuard` will never hold back a vote on `Add(Dave)`. We only call `attempt_to_remove` on a vote for `Remove`.

I agree that `the scheduled order does not necessarily guarantee consensus order`. However, swapping such order may hide other potential issue, such as the previous removal of nodes got consensused super late for some reason or even hang.

src/dev_utils/network.rs
```diff
@@ -458,28 +458,21 @@ impl Network {
...
 
-            self.check_consensus_broken()?;
-            if self.consensus_complete(&peers, max_observations) {
-                break;
+                self.check_consensus_broken()?;
+                if self.consensus_complete(&peers, max_observations) {
+                    break;
+                }
+            } else {
+                retry.push(event);
```
octol: Must admit I had to reread this a few times before understanding it

PR #218, opened 2018-12-14
MAID-3275: Reduce PublicId lookups

madadam:
This PR replaces `PublicId` with `PeerIndex` in almost all places. As `PeerIndex` is just a newtype over `usize`, this reduces the time spend looking up entries in maps end sets keyed by `PeerIndex`. Further, it replaces all `BTreeMap` / `BTreeSet` of small keys (`EventIndex` or `PeerIndex`) with `FnvHashMap` and `FnvHashSet` from the [fnv crate](https://github.com/servo/rust-fnv), which further speeds the lookups up.

As `PeerIndex` is no longer "objective" (different peers will have different `PeerIndex` identifying the same peer), this PR also modifies the test code to account for that (similarly to how we handled replacing `EventHash` with `EventIndex` in the past).


src/peer_list/peer.rs
```diff
@@ -0,0 +1,167 @@
...
+            .events
+            .indexed()
+            .rev()
+            .next()
+            .map(|(index, _)| index)
+            .unwrap_or(0);
+        self.membership_list_changes.push((index, change));
+    }
+
+    pub(crate) fn membership_list(&self) -> &BTreeSet<P> {
```
jeanphilippeD: Do we need to remove the #[cfg(feature = "dump-graphs")], the member is public for the module that needs it?

src/peer_list/peer.rs
```diff
@@ -85,17 +93,51 @@ impl<P: PublicId> Peer<P> {
...
+impl Peer<PeerId> {
+    pub(super) fn new_from_dot_input(
+        peer_id: &PeerId,
+        state: PeerState,
+        membership_list: BTreeSet<PeerId>,
+        graph: &Graph<Transaction, PeerId>,
+    ) -> Self {
+        let mut events = Events::new();
+        for event in graph {
+            if event.creator() == peer_id {
```
jeanphilippeD: This was dropped, why?

```
} else if !peer_ids.contains(event.creator()) {
                    debug!(
                        "peer_data list doesn't contain the creator of event {:?}",
                        *event
                    );
                }
```

src/peer_list/mod.rs
```diff
@@ -435,35 +433,12 @@ impl PeerList<PeerId> {
...
-                .into_iter()
-                .chain(iter::once(peer_id.clone()))
-                .map(MembershipListChange::Add)
-            {
-                peer.change_membership_list(change)
-            }
-
-            let _ = peers.insert(peer_id, peer);
-        }
+        let peers = peer_data
```
jeanphilippeD: This is so much better.

src/dump_graph.rs
```diff
@@ -116,15 +116,16 @@ mod detail {
     thread_local!(static DUMP_COUNTS: RefCell<BTreeMap<String, usize>> =
         RefCell::new(BTreeMap::new()));
 
-    fn catch_dump<T: NetworkEvent, P: PublicId>(
+    fn catch_dump<T: NetworkEvent, S: SecretId>(
```
jeanphilippeD: So there is a lot of changes in this commit 'chore/peer_list: use PeerIndex instead of PublicId ' that seem to be related to testing and dump graph. I assume our automated tests / unit test will cover the parsing, Does it cover the dumpgraph, what kind of test for that?

src/parsec.rs
```diff
@@ -1588,7 +1584,7 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
 
         // IMPORTANT: We must sort this in consistent order, so when the tie breaking rule kicks in,
         // the outcome is the same for everyone.
-        payloads.sort();
+        payloads.sort_by(|a, b| a.hash().cmp(b.hash()));
```
jeanphilippeD: Not clear to me why this change?

src/dev_utils/dot_parser.rs
```diff
@@ -14,7 +14,7 @@ use meta_voting::{
 };
 use mock::{PeerId, Transaction};
 use observation::{Observation, ObservationHash, ObservationKey};
-use peer_list::{PeerIndex, PeerList, PeerState};
+use peer_list::{PeerIndexMap, PeerIndexSet, PeerList, PeerState};
```
jeanphilippeD: That is great, should really help when trying different maps.

src/dev_utils/dot_parser.rs
```diff
@@ -688,7 +688,7 @@ impl ParsedContents {
     /// Create empty `ParsedContents`.
     pub fn new(our_id: PeerId) -> Self {
         let peer_list = PeerList::new(our_id.clone());
-        let meta_elections = MetaElections::new(BTreeSet::new());
+        let meta_elections = MetaElections::new(PeerIndexSet::default());
```
jeanphilippeD: should this be in previous commit?
I think it most likely should.
Same of similar updates throughout in this commit.

src/functional_tests.rs
```diff
@@ -222,23 +222,29 @@ fn add_peer() {
...
     let event_index = 18;
-    let mut alice_membership_list_for_dave = vec![
+    let mut expected_alice_membership_list_for_dave = vec![
         (event_index, MembershipListChange::Add(alice_index)),
         (event_index, MembershipListChange::Add(bob_index)),
         (event_index, MembershipListChange::Add(carol_index)),
         (event_index, MembershipListChange::Add(dave_index)),
         (event_index, MembershipListChange::Add(eric_index)),
     ];
+    expected_alice_membership_list_for_dave.sort();
```
jeanphilippeD: Why the sort?

src/meta_voting/meta_elections.rs
```diff
@@ -42,7 +43,7 @@ impl Debug for MetaElectionHandle {
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub(crate) struct MetaElection {
-    pub(crate) meta_events: BTreeMap<EventIndex, MetaEvent>,
+    pub(crate) meta_events: FnvHashMap<EventIndex, MetaEvent>,
```
jeanphilippeD: Would it make sense to have the same kind of type alias as for PeerIndexMap, PeerIndexSet

src/peer_list/peer.rs
madadam: If I understand correctly what you are asking, the yes, the attribute had to go because this function is now needed elsewhere, not just in `dump_graph`, because the `membership_list` field is no longer accessible outside of this module.

src/peer_list/peer.rs
madadam: We no longer have access to the necessary data to perform the check here. It was just additional sanity check to make sure all the events are created by someone we know. Don't think the check is essential. I can put it back to some more appropriate place if you thinks it's worth it.

src/dump_graph.rs
madadam: Yes, the `dev_utils::dot_parser::tests::dot_parser` also covers dumping. Also, as mentioned in other comment, the changes were necessary to make everything still compile. Even though they might seem unrelated to the purpose of this commit.

src/parsec.rs
madadam: As the comment says, we need consistent order here (every node should arrive at the same order), because this is the point where we pick the next consensused block. `payloads` contain `ObservationKey` which contains `PeerIndex`, but `PeerIndex` is not consistent (different nodes might use different `PeerIndex` to refer to the same peer), so we need to exclude it from the comparison.

src/dev_utils/dot_parser.rs
madadam: If we want to be pedantic then yes, it should. Problem is the compiler didn't complain that I'm using `BTreeSet` instead of `PeerIndexSet` because it was just a type alias at this point. When I changed it to `FnvHashSet` then it started to complain, but it was already in different commit and I didn't feel like going back and changing the previous commit.

src/functional_tests.rs
madadam: The values now come out of `HashMap` which doesn't guarantee any order. Besides, for the same `event_index`, the order isn't important anyway.

src/meta_voting/meta_elections.rs
madadam: It might, but this isn't used nearly as much as `PeerIndexSet` / `PeerIndexMap` so I didn't feel the need.

src/peer_list/peer.rs
jeanphilippeD: No it's ok. Would have been good to have mentioned it in commit comment as it is lost in the diff.

src/dump_graph.rs
jeanphilippeD: I guess my question was did you test/spot check that the dump graph functionality still work (beside it compiling), or is the automated test that did that for you?

madadam: I relied on the tests only. The way it works is that it generates some parsec graphs using the testing framework, dumps them, then parses them back and compares with the originals. Seems thorough enough so I didn't bother checking it manually.

src/parsec.rs
```diff
@@ -479,9 +487,13 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         }
     }
 
+    fn get_peer_index(&self, peer_id: &S::PublicId) -> Result<PeerIndex> {
```
Fraser999: Getters normally shouldn't be prefixed with `get_`.

src/peer_list/membership_list.rs
```diff
@@ -0,0 +1,43 @@
...
+impl MembershipListChange {
+    pub(super) fn apply(&self, peers: &mut PeerIndexSet) -> bool {
+        match *self {
+            MembershipListChange::Add(index) => peers.insert(index),
+            MembershipListChange::Remove(index) => peers.remove(&index),
+        }
+    }
+
+    #[cfg(feature = "malice-detection")]
+    pub(super) fn unapply(&self, peers: &mut PeerIndexSet) -> bool {
```
Fraser999: "unapply" could be "revert" maybe?

src/parsec.rs
madadam: This isn't a getter though. I think getter would be something that doesn't take any arguments, no?

Fraser999: Not sure.  Maybe that is the strict definition, but I guess the spirit of that convention would still apply here, wouldn't it?  For example, the next function could have been called `get_our_last_event_hash`?

Not a big deal anyway - it's purely a style issue.  I'm fine to approve as is.

PR #219, opened 2018-12-14
Use custom, more efficient data structures for PeerIndexSet and PeerIndexMap

madadam:


src/peer_list/peer_index.rs
```diff
@@ -16,5 +18,337 @@ impl PeerIndex {
...
+    pub fn keys<'a>(&'a self) -> impl Iterator<Item = PeerIndex> + 'a {
+        self.0
+            .iter()
+            .enumerate()
+            .filter(|(_, value)| value.is_some())
+            .map(|(index, _)| PeerIndex(index))
+    }
+
+    pub fn insert(&mut self, key: PeerIndex, value: T) -> Option<T> {
+        self.reserve(key.0 + 1);
```
maqi: seems this container will only keeps growing. will it cause problem in space usage, or performance slow down after growing a while?

src/peer_list/peer_index.rs
```diff
@@ -16,5 +18,337 @@ impl PeerIndex {
...
+}
+
+impl Extend<PeerIndex> for PeerIndexSet {
+    fn extend<I>(&mut self, iter: I)
+    where
+        I: IntoIterator<Item = PeerIndex>,
+    {
+        let iter = iter.into_iter();
+        let (lower, _) = iter.size_hint();
+        self.0.reserve(lower);
```
maqi: we are appending the values in the iter into self, shall we reserve `self.0.len() + lower` ?

src/peer_list/peer_index.rs
```diff
@@ -16,5 +18,337 @@ impl PeerIndex {
...
+}
+
+impl Iterator for SetIntoIter {
+    type Item = PeerIndex;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            let index = self.current;
+            if index >= self.set.0.len() {
+                return None;
```
maqi: shall we have a log_or_pance here?

src/peer_list/peer_index.rs
```diff
@@ -16,5 +18,337 @@ impl PeerIndex {
     pub const OUR: Self = PeerIndex(0);
 }
 
-pub(crate) type PeerIndexSet = FnvHashSet<PeerIndex>;
-pub(crate) type PeerIndexMap<T> = FnvHashMap<PeerIndex, T>;
+/// Map keyed by `PeerIndex`.
```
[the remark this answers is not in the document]
madadam: Same thing.

src/peer_list/peer_index.rs
madadam: Well, I assume the number of peers will remain relatively low during a normal run (say less than 100 or so). Also merged / splits and prunes should keep this list from growing too much.

In any case, this PR hasn't changed anything in this regard. We weren't removing peers even before. That said, if we end up needing to remove peers at some point, then these collections would have to be modified.

src/peer_list/peer_index.rs
madadam: The argument passed to `reserve` is the additional number of items we want to allocate, not the total number. So after `reserve(n)`, we end up with at least previous len + n allocated items: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.reserve

src/peer_list/peer_index.rs
madadam: This is where we reach the end of the iteration, so it's a perfectly normal situation. I see no reason to panic.

src/peer_list/peer_index.rs
maqi: do you think under this new structure, the `remove` feature can be supported meanwhile retaining the performance gain?

madadam: `remove` could be implemented by setting the corresponding element to `None`, but that wouldn't decrease the size of the vector. What we could do is keep calling `pop` until the last element in the vector is not `None`. But that doesn't guarantee the vector won't grow indefinitely in some rare weird cases.

PR #222, opened 2018-12-20
MAID-3276: remove redundant fields from Event

madadam:
This PR removes some fields from `Event` that weren't necessary:

- `self_parent_hash: EventHash` and `other_parent_hash: EventHash` - can be looked up by `self_parent: EventIndex` and `other_parent: EventIndex`.
- `creator_id: P` - can be looked up by `creator: PeerIndex`
- `vote: Vote<T, P>` - replaced by `VoteKey`, which can be used to look up the actual `Vote`.

In order to do this, `Content` and `Cause` are now generic over the type of vote (`Vote` or `VoteKey`), the type of event indentifier (`EventHash` or `EventIndex`) and the type of peer identifier (`PublicId` or `PeerIndex`). `Event` now contains the ones specialized with `VoteKey`, `EventIndex` and `PeerIndex`, whereas `PackedEvent` has `Vote`, `EventHash` and `PublicId`.

`Event` is no longer generic over `T: NetworkEvent` which is the cause of the majority of changes in this PR. Those changes are trivial though.

To simplify the code, two new reference-like types are introduced: `EventContextRef` and `EventContextMut`, which contain references to all the data necessary to construct, pack or unpack `Event`s. I prefer to pass `EventContextRef` and  `EventContextMut` by value, because they are conceptually references, but clippy disagrees with this idea. Thus the `#[allow: clippy::needless_pass_by_value]` attributes in few places.

### DIFF
--- a/src/dev_utils/dot_parser.rs
+++ b/src/dev_utils/dot_parser.rs
@@ -1081,6 +1081,7 @@ mod tests {
 
             let mut dot_file_path = entry.path();
             assert!(dot_file_path.set_extension("dot"));
+
             let parsed = unwrap!(parse_dot_file(&dot_file_path));
             let actual_snapshot = (
                 GraphSnapshot::new(&parsed.graph),
--- a/src/dev_utils/dot_parser.rs
+++ b/src/dev_utils/dot_parser.rs
@@ -7,13 +7,25 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use gossip::{CauseInput, Event, EventIndex, Graph, IndexedEventRef};
+#[cfg(any(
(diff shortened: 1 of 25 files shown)


tests/integration_tests.rs
```diff
@@ -77,7 +77,8 @@ use rand::Rng;
 use std::collections::BTreeMap;
 
 // Alter the seed here to reproduce failures
-static SEED: RngChoice = RngChoice::SeededRandom;
+// static SEED: RngChoice = RngChoice::SeededRandom;
+static SEED: RngChoice = RngChoice::Seeded([1472249392, 3295493059, 2971722000, 2214046456]);
```
madadam: Oops, I accidentally committed this :flushed:

PR #229, opened 2019-01-07
bugfix/tests: fix various issues in tests

Fraser999:
This change avoids the blocks handled by non-active test nodes as counting towards the voting quorum.  It also fixes a minor bug in the test code whereby failed peers were treated as running, and a logging statement which triggers a subtraction with overflow.


src/parsec.rs
```diff
@@ -796,7 +796,7 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         info!(
             "{:?} got consensus on block {} with payload {:?} and payload hash {:?}",
             self.our_pub_id(),
-            self.meta_elections.consensus_history().len() - 1,
+            self.meta_elections.consensus_history().len(),
```
madadam: Good catch!

PR #230, opened 2019-01-08
chore/meta_elections: restrict multi-elections to malice-handling

Fraser999:


src/meta_voting/meta_elections.rs
```diff
@@ -456,7 +461,9 @@ impl MetaElections {
     }
 
     fn not_found(handle: MetaElectionHandle) {
-        log_or_panic!("Meta-election at {:?} not found", handle);
+        if cfg!(feature = "malice-detection") {
+            log_or_panic!("Meta-election at {:?} not found", handle);
+        }
```
octol: So this is the new thing we using? The `if cfg` instead of `#[cfg(..)]` ?

Fraser999: They're slightly different.  The `cfg!` macro can make for cleaner code, since it expands to a boolean expression at compile time.  However, compilation is more likely to fail in this case, since code inside such an expression needs to be compilable whether or not the specified configuration is enabled for that build or not.

So, for example if we have:
```rust
#[cfg(windows)]
fn windows_function() {}
```
then on a unix platform, the following will fail to compile:
```rust
if cfg!(windows) {
    windows_function()
}
```
We'd _have_ to invoke it like so:
```rust
#[cfg(windows)]
windows_function();
```

pierrechevalier83:
As discussed today, we will instead remove multi-elections in all cases for performance/simplicity. Closing this PR as the approach taken will be different.

PR #231, opened 2019-01-17
Test scalability with section size

jeanphilippeD:
Add env var PARSEC_DUMP_GRAPH_SVG & PARSEC_DUMP_GRAPH_PEERS to speed up dot file generation. (This can also be used for large test, i.e use dump-graph in routing tests.)

Add New bench marking tests.
Note: a number of the dot file generated panic when used for benchmark so are disabled until the bug in master is fixed.

With 8 Opaque events and 8, 16, 24 nodes:
a_node8_opaque_evt8     9.4998 ms
a_node16_opaque_evt8    86.623 ms
a_node24_opaque_evt8    426.38 ms

With 16 Opaque events and 4, 8, 12, 16, 24 nodes:
a_node4_opaque_evt16    5.0746 ms
a_node8_opaque_evt16    24.332 ms
a_node12_opaque_evt16   187.82 ms
a_node16_opaque_evt16   442.63 ms
a_node24_opaque_evt16   1.8346 s


dot_gen/src/main.rs
```diff
@@ -41,9 +41,13 @@
 //!
 //! When scenarios are defined, run the tool from within Parsec's root directory:
 //!
-//!     cargo run --manifest-path=dot_gen/Cargo.toml -- ARGS
-//!
+//!     cargo run --release --manifest-path=dot_gen/Cargo.toml -- ARGS
```
madadam: :+1:

src/dump_graph.rs
```diff
@@ -87,6 +87,18 @@ mod detail {
...
+
+        static ref GENERATE_SVG: bool = {
+            // PARSEC_DUMP_GRAPH_SVG=0 to disable svg file generation
+            env::var("PARSEC_DUMP_GRAPH_SVG").ok().map_or(true, |x| x != "0")
+        };
+
+        static ref FILTER_PEERS: Option<Vec<String>> = {
+            // PARSEC_DUMP_GRAPH_PEERS=Alice,Bob to only dump graph for them.
+            env::var("PARSEC_DUMP_GRAPH_PEERS").ok().map(|x| {
+                x.split(',').map(|x | x.to_string()).collect::<Vec<String>>()
```
madadam: I'd `trim` the names before collecting them, just in case someone decides to use spaces: `PARSEC_DUMP_GRAPH_PEERS="Alice, Bob"`

src/mock.rs
```diff
@@ -18,7 +18,7 @@ use std::hash::{Hash, Hasher};
 pub const NAMES: &[&str] = &[
     "Alice", "Bob", "Carol", "Dave", "Eric", "Fred", "Gina", "Hank", "Iris", "Judy", "Kent",
     "Lucy", "Mike", "Nina", "Oran", "Paul", "Quin", "Rose", "Stan", "Tina", "Ulf", "Vera", "Will",
-    "Xaviera", "Yakov", "Zaida",
+    "Xaviera", "Yakov", "Zaida", "0", "1", "2", "3", "4", "5", "6", "7", "8", "9",
```
madadam: I'd prefer the numerical names to be generated instead of hard-coded. That way we won't limit the tests to a fixed number of nodes.

src/dump_graph.rs
```diff
@@ -87,6 +87,18 @@ mod detail {
                 .collect::<String>()
         };
         static ref ROOT_DIR: PathBuf = { ROOT_DIR_PREFIX.join(&*ROOT_DIR_SUFFIX) };
+
+        static ref GENERATE_SVG: bool = {
+            // PARSEC_DUMP_GRAPH_SVG=0 to disable svg file generation
+            env::var("PARSEC_DUMP_GRAPH_SVG").ok().map_or(true, |x| x != "0")
+        };
+
+        static ref FILTER_PEERS: Option<Vec<String>> = {
```
madadam: Minor, but consider changing the type to just `Vec<String>` and treating empty `Vec` same as you treat `None` now. Currently someone could use `PARSEC_DUMP_GRAPH_PEERS=""` which would disable the dumps of all peers which doesn't seem useful.

dot_gen/src/main.rs
```diff
@@ -206,6 +210,43 @@ fn main() {
         })
         .file("Alice", "dynamic.dot");
 
+    let add_bench_scalability = |s: &mut Scenarios, opaque_to_add: usize, genesis_size: usize| {
```
Fraser999: I'm not sure this block should be committed.  I think the main idea here was to provide some example code which showed how to use the generator.

If you feel we'll likely be adding further scalability cases, then perhaps this could even be reduced to show a single file inside being created inside `add_bench_scalability()`, and only having one subsequent call to that function?

(I'm also a bit surprised that rustfmt allowed the spaces after the `opaque_to_add`s.)

dot_gen/src/main.rs
```diff
@@ -41,9 +41,13 @@
...
 //! When scenarios are defined, run the tool from within Parsec's root directory:
 //!
-//!     cargo run --manifest-path=dot_gen/Cargo.toml -- ARGS
-//!
+//!     cargo run --release --manifest-path=dot_gen/Cargo.toml -- ARGS
+//! 
 //! where ARGS are the arguments to the tool. Run with --help for more info.
+//!
+//! To speed up generation of bigger test cases:
+//!     PARSEC_DUMP_GRAPH_SVG=0 PARSEC_DUMP_GRAPH_PEERS=Alice,Bob,Carol cargo run --release --manifest-path=dot_gen/Cargo.toml -- ARGS
```
Fraser999: It's not obvious how `PARSEC_DUMP_GRAPH_PEERS` interacts with the list of files specified via `Scenarios::file` calls.  It's also not obvious why we need both of these things.  But given that this env var isn't likely to be widely used I think that may be ok?

Perhaps worth adding a bit more to the rationale for why it's needed (since the list of files specified in the `Scenario` are actually just copied over from the temp dir dump of all nodes' files) and maybe also link to that explanation from the new comment at line 97 of dump_graphs.rs?

Not a blocker though, as far as I'm concerned.

benches/bench.rs
```diff
@@ -65,15 +65,64 @@ use parsec::dev_utils::Record;
...
 fn bench(c: &mut Criterion) {
-    bench_dot_file(c, "minimal");
-    bench_dot_file(c, "static");
-    bench_dot_file(c, "dynamic");
+    for name in &["minimal", "static", "dynamic"] {
+        bench_dot_file(c, "benches", name);
+    }
+
+    for name in &[
+        //"a_node4_opaque_evt8", // panic
```
Fraser999: I think we should have a slightly longer comment to explain that these instances are disabled because they currently fail, and that they'll be uncommented later.

src/dump_graph.rs
jeanphilippeD: I think I would not really want to do any advance parsing for this.
Because it is so low level I don't know that we should accept that flexibility.

src/mock.rs
jeanphilippeD: No, this is a currently hard limitation of the code.
I would want what you want since I wanted 256 nodes. The problem is we have a short name that is used in dump-graph that use the first character of the Id and capitalize it.

So this is the full list of possible pretty names.

src/dump_graph.rs
jeanphilippeD: I was thinking of this as a potential added benefit.
Also it is semantically strange, it make empty vector a special value, it seems more intuitive to not have special value in vector, the fact that there is a difference is explicit in use of option.

src/mock.rs
madadam: Ah, right. Makes sense then.

dot_gen/src/main.rs
madadam: This is a good point. Actually I'd now prefer to not have the `PARSEC_DUMP_GRAPH_PEERS` env var at all, instead use `Scenarios::file` for that purpose. Would require more changes to the code but I think it would result in cleaner in the end.

dot_gen/src/main.rs
jeanphilippeD: I followed how previous benchmark were added.
I also feel strongly that we want to be able to regenerate .dot files without having to write code again for maintainability of our tests

dot_gen/src/main.rs
jeanphilippeD: This is a much more difficult thing to do.
If you are to restrictive with the env var the dot_gen will fail and tell you it could not find some files.

benches/bench.rs
jeanphilippeD: This information was in the commit comment:
"
Note:
-Many of the new tests are commented out as they panic.
 These new test reveal an existing bug on master that can be fixed at
 a later time.
"

Is that the kind of detail we would put in code comment?

dot_gen/src/main.rs
```diff
@@ -41,9 +41,13 @@
 //!
 //! When scenarios are defined, run the tool from within Parsec's root directory:
 //!
-//!     cargo run --manifest-path=dot_gen/Cargo.toml -- ARGS
-//!
+//!     cargo run --release --manifest-path=dot_gen/Cargo.toml -- ARGS
+//!
```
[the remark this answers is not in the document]
jeanphilippeD: Updated my settings.
I can push a fixup commit for that?

src/dump_graph.rs
Fraser999: I see your point, but don't feel particularly strongly either way.  I think I still find Adam's approach less surprising, but I don't mind too much either way.

dot_gen/src/main.rs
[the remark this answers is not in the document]
Fraser999: As far as I'm concerned, only if there are other changes being made too.

dot_gen/src/main.rs
Fraser999: I think this might make the production code a bit worse though, so I'd prefer if we kept the env var.  If anything, modifying that to allow it to replace the `Scenarios::file()` would seem preferable.  But at the very least, I think the behaviour should be documented in code comments in a bit more detail.

benches/bench.rs
Fraser999: I would find it more useful as a code comment than a commit one.  I don't see any reason to not copy the commit comment exactly, but I'd also think something less verbose would do as a code comment.

jeanphilippeD: I think this is a lot less relevant now that the last commit leave no panic tests.

dot_gen/src/main.rs
jeanphilippeD: Addressed.

dot_gen/src/main.rs
jeanphilippeD: ok. leaving it then.

src/dump_graph.rs
jeanphilippeD: ok. leaving it then.

src/dump_graph.rs
jeanphilippeD: ok. leaving it then.

PR #235, opened 2019-01-22
Less reprocessing

Fraser999:
This is simply #233 rebased on current master.

1. Try optimize the computation of observer. Before we calculated the set of "observees" for every event and then use it to determine if the event is observer. But we don't need to keep the observees around, we only need to know if the event is observer.
So we now compute the observer status right away and store the observees only if the event IS observer. This saves memory and might also improve performance a bit. I was also hoping to reuse the observer status in subsequent meta-elections, but that didn't work out.

2. Mostly refactoring to simplify how we calculate interesting content. This also improves the performance: Only calculate interesting content when creating new meta-event, not when reprocessing them. On membership-change, clear all meta-events first so we falls-back to recalculating everything.
Note: recalculate everything in this case also fixed a bug where we would sometimes invalidly reuse some meta-events after a membership change.

3. Tried to reuse the observer status from previous meta-election, but I think this change is reverted in later  commits because it didn't really work and/or make a difference.

4. Better graph dumps (not really related to the rest of the PR, but it simplified debugging some issues).

5. Main part of the PR: Now reprocess only from the first observer instead of from the first unconsensused event.

### DIFF
--- a/src/dev_utils/dot_parser.rs
+++ b/src/dev_utils/dot_parser.rs
@@ -376,7 +376,7 @@ fn parse_hash() -> Parser<u8, Hash> {
 struct ParsedMetaElection {
     round_hashes: BTreeMap<PeerId, Vec<RoundHash>>,
     interesting_events: BTreeMap<PeerId, Vec<String>>,
-    all_voters: BTreeSet<PeerId>,
+    voters: BTreeSet<PeerId>,
     payload: Option<Observation<Transaction, PeerId>>,
     unconsensused_events: BTreeSet<String>,
     observation_map: BTreeMap<ObservationKey, Observation<Transaction, PeerId>>,
@@ -397,17 +397,14 @@ fn parse_meta_election(ctx: &Rc<ParserCtx>) -> Parser<u8, ParsedMetaElection> {
         * (parse_consensus_history() - next_line()
             + parse_round_hashes()
             + parse_interesting_events()
-            + parse_all_voters()
+            + parse_voters()
             + parse_payload().opt()
             + parse_unconsensused_events()
             + parse_meta_events(ctx))
         .map(
             |(
                 (
-                    (
-                        (((consensus_history, round_hashes), interesting_events), all_voters),
-                        payload,
-                    ),
+                    ((((consensus_history, round_hashes), interesting_events), voters), payload),
                     unconsensused_events,
                 ),
                 observation_map_and_meta_events,
@@ -422,7 +419,7 @@ fn parse_meta_election(ctx: &Rc<ParserCtx>) -> Parser<u8, ParsedMetaElection> {
                 ParsedMetaElection {
                     round_hashes,
                     interesting_events,
-                    all_voters,
+                    voters,
                     payload,
                     unconsensused_events,
                     observation_map,
@@ -490,7 +487,7 @@ fn parse_interesting_events_for_peer() -> Parser<u8, (PeerId, Vec<String>)> {
         - next_line()
 }
 
-fn parse_all_voters() -> Parser<u8, BTreeSet<PeerId>> {
+fn parse_voters() -> Parser<u8, BTreeSet<PeerId>> {
     comment_prefix() * seq(b"all_voters: ") * parse_peers() - next_line()
 }
 
@@ -872,7 +869,7 @@ fn convert_to_meta_election(
             })
             .collect(),
         round_hashes: convert_peer_id_map(meta_election.round_hashes, peer_list),
-        all_voters: convert_peer_id_set(meta_election.all_voters, peer_list),
+        voters: convert_peer_id_set(meta_election.voters, peer_list),
         interesting_events: meta_election
             .interesting_events
             .into_iter()
@@ -899,6 +896,8 @@ fn convert_to_meta_election(
             .filter_map(|id| event_indices.get(&id))
             .cloned()
             .collect(),
+        // TODO: parse these too:
+        observers: BTreeSet::new(),
         consensus_history: meta_election.consensus_history,
     }
 }
--- a/src/dev_utils/dot_parser.rs
+++ b/src/dev_utils/dot_parser.rs
@@ -717,15 +717,15 @@ impl ParsedContents {
     }
 
     #[cfg(all(feature = "malice-detection", feature = "mock"))]
-    /// Insert event into the `ParsedContents`. Note this does not perform any validations whatsoever,
-    /// so this is useful for simulating all kinds of invalid or malicious situations.
+    /// Insert event into the `ParsedContents`. Note this does not perform any validations
+    /// whatsoever, so this is useful for simulating all kinds of invalid or malicious situations.
     pub fn add_event(&mut self, event: Event<PeerId>) -> EventIndex {
         let indexed_event = self.graph.insert(event);
         self.peer_list.add_event(indexed_event);
 
         let start_index = indexed_event.event_index().topological_index() + 1;
-        self.meta_election.lower_start_index = start_index;
-        self.meta_election.upper_start_index = start_index;
+        self.meta_election.new_consensus_start_index = start_index;
+        self.meta_election.continue_consensus_start_index = start_index;
 
         indexed_event.event_index()
     }
@@ -902,8 +902,8 @@ fn convert_to_meta_election(
             .cloned()
             .collect(),
         consensus_history: meta_election.consensus_history,
-        upper_start_index: 0,
-        lower_start_index: 0,
+        continue_consensus_start_index: 0,
+        new_consensus_start_index: 0,
     }
 }
 
@@ -911,11 +911,7 @@ fn convert_to_meta_event(meta_event: ParsedMetaEvent, peer_list: &PeerList<PeerI
     let observees = convert_peer_id_set(meta_event.observees, peer_list);
 
     MetaEvent {
-        observer: if observees.is_empty() {
-            Observer::None
-        } else {
-            Observer::This(observees)
-        },
+        observer: Observer::new(observees),
         interesting_content: meta_event.interesting_content,
         meta_votes: convert_peer_id_map(meta_event.meta_votes, peer_list),
     }
--- a/src/dev_utils/dot_parser.rs
+++ b/src/dev_utils/dot_parser.rs
@@ -722,6 +722,11 @@ impl ParsedContents {
     pub fn add_event(&mut self, event: Event<PeerId>) -> EventIndex {
         let indexed_event = self.graph.insert(event);
         self.peer_list.add_event(indexed_event);
+
+        let start_index = indexed_event.event_index().topological_index() + 1;
+        self.meta_election.lower_start_index = start_index;
+        self.meta_election.upper_start_index = start_index;
+
         indexed_event.event_index()
     }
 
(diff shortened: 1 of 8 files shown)


src/meta_voting/meta_election.rs
```diff
@@ -203,6 +200,117 @@ impl MetaElection {
...
+        decided_key: &ObservationKey,
+        peer_list_changed: bool,
+    ) {
+        self.interesting_events.clear();
+
+        if peer_list_changed {
+            return;
+        }
+
+        for meta_event in self.meta_events.values_mut() {
```
jeanphilippeD: I think this loop should be in update_meta_events

src/meta_voting/meta_election.rs
```diff
@@ -203,6 +200,117 @@ impl MetaElection {
...
+    fn update_lower_start_index<P: PublicId>(&mut self, graph: &Graph<P>) {
+        self.lower_start_index = self
+            .unconsensused_events
+            .iter()
+            .next()
+            .map(|event_index| event_index.topological_index())
+            .unwrap_or_else(|| graph.len());
+    }
+
+    fn update_upper_start_index(&mut self, peer_list_changed: bool) {
```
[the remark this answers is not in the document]
Fraser999: I don't agree that this is a problem.  These are private methods and `update_lower_start_index()` has a similar dependency on `update_unconsensused_events()` having been called first.  If we link all of these which have an implicit dependency of ordering, we more or less end up with `new_election()`.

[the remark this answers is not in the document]
jeanphilippeD: Yes, this is probably reasonable.

src/meta_voting/meta_election.rs
```diff
@@ -234,22 +237,26 @@ impl MetaElection {
...
 
-    fn update_lower_start_index<P: PublicId>(&mut self, graph: &Graph<P>) {
+    fn update_lower_start_index(&mut self) {
         self.lower_start_index = self
             .unconsensused_events
             .iter()
             .next()
             .map(|event_index| event_index.topological_index())
-            .unwrap_or_else(|| graph.len());
+            .unwrap_or(usize::MAX);
```
jeanphilippeD: This feels quite weird to me (but i guess in principle it is the same as the original change, at least it is less subtle).
If possible I'd rather avoid a sentinel value: Should it be an option or would it make using it much harder?

src/meta_voting/meta_election.rs
```diff
@@ -268,34 +275,31 @@ impl MetaElection {
...
+                    if event_index.topological_index() >= lower_start_index {
+                        meta_event
+                            .interesting_content
+                            .retain(|payload_key| payload_key != decided_key);
+                        Some((event_index, meta_event))
+                    } else {
+                        None
+                    }
+                })
+                .collect();
```
jeanphilippeD: I am not sure that it is better than bringing the loop withing this function i.e:

```
      self.meta_events
                .retain(|event_index, _| event_index.topological_index() >= lower_start_index);

       for meta_event in self.meta_events.values_mut() {
            meta_event
                .interesting_content
                .retain(|payload_key| payload_key != decided_key);
        }
```

I am having quite a struggle identifying if it does the same thing.
It is not clear to me that it would be more efficient either

src/meta_voting/meta_election.rs
```diff
@@ -203,6 +203,118 @@ impl MetaElection {
...
+        decided_key: &ObservationKey,
+    ) {
+        let remove = self
+            .unconsensused_events
+            .iter()
+            .filter(|event_index| {
+                graph
+                    .get(**event_index)
+                    .map(|event| event.payload_key() == Some(decided_key))
+                    .unwrap_or(false)
```
jeanphilippeD: Would that be clearer?

```
            .filter(|event_index| {
                graph
                    .get(**event_index)
                    .and_then(|event| event.payload_key()) == Some(decided_key)
```

jeanphilippeD:
Would it be worth moving in its own PR the commit `chore/parsec: improve graph dumps`
This commit will modify the output of dump graph and so require regenerating them.

I do not believe the other commits intend to change dump graphs? so it would also be a great sanity check that things have not changed.

src/meta_voting/meta_election.rs
```diff
@@ -268,34 +275,31 @@ impl MetaElection {
...
+    fn update_meta_events(&mut self, decided_key: &ObservationKey, peer_list_changed: bool) {
         if peer_list_changed {
             self.meta_events.clear();
         } else {
             let lower_start_index = self.lower_start_index;
-            self.meta_events
-                .retain(|event_index, _| event_index.topological_index() >= lower_start_index);
+            let mut meta_events = mem::replace(&mut self.meta_events, FnvHashMap::default());
+            self.meta_events = meta_events
+                .drain()
```
madadam: It doesn't seem `meta_events` is used after this expression, so the `drain` can be replaced by `into_iter` and then `meta_events` won't have to be `mut`.

src/meta_voting/meta_election.rs
madadam: Apart from what @jeanphilippeD  said, I think this is actually a minor performance pessimisation: we are throwing away the memory allocated in `self.meta_events` and creating new collection every time. Previous code would reuse the memory (I think. Depends on how `HashMap::retain` is implemented, but I assume it doesn't reallocate).

src/meta_voting/meta_election.rs
madadam: `usize::MAX` can be considered infinity for all practical purposes* and it's use simplifies the code. `Option` would make things messy I think.

*) If we want to be paranoid and defensive, we can insert an assertion info `Graph::insert` which makes sure the number of events is always less than `usize::MAX`.

src/meta_voting/meta_election.rs
Fraser999: I can make the change as requested.  I'm not sure about pessimisation or clarity; with the proposed change we're iterating twice, but I'll change to it since both reviewers are in agreement.

src/dev_utils/dot_parser.rs
```diff
@@ -668,6 +665,11 @@ impl ParsedContents {
     pub fn add_event(&mut self, event: Event<PeerId>) -> EventIndex {
         let indexed_event = self.graph.insert(event);
         self.peer_list.add_event(indexed_event);
+
+        let start_index = indexed_event.event_index().topological_index() + 1;
+        self.meta_election.lower_start_index = start_index;
+        self.meta_election.upper_start_index = start_index;
```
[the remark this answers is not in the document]
madadam: The honest answer would be "to make the existing tests pass" :). Without this change, if you constructed a `Parsec` instance from `ParsedContents` and then called anything that triggers event processing (`handle_request`, `handle_response`, ...), then it would reprocess also the events added by this function, which might not be desirable. I think the idea is that `ParsedContents` is a sort of snapshot of `Parsec` state, so the events in it are considered as already processed. 

That said, this feels bit dirty and it might be worth refactoring at some point. Probably low priority for now though.

src/dump_graph.rs
```diff
@@ -548,6 +553,8 @@ mod detail {
                 .all_ids()
                 .map(|(_, id)| id)
                 .collect::<Vec<_>>();
+            peer_ids.sort();
```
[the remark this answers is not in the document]
Fraser999: Well, we still need it to be mutable since we pop from it later, but I'll change to use `sorted()` regardless.

src/meta_voting/meta_election.rs
Fraser999: It would be marginally clearer maybe, but the compiler isn't keen on returning a reference to `event`'s member from `|event| event.payload_key()`, so I'd have to clone the contained key.

src/meta_voting/meta_event.rs
```diff
@@ -121,13 +133,17 @@ pub(crate) mod snapshot {
         where
             S: SecretId<PublicId = P>,
         {
-            Self {
-                observees: meta_event
-                    .observees
+            let observees = match meta_event.observer {
+                Observer::This(ref observees) => observees
```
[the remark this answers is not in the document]
Fraser999: I've had problems before with returning iterators like this, since in one case we'd be returning an iterator proper and in the other cases we'd want something like `std::iter::Empty`.

Would you be happy with a method `Observation::has_observee(&self, peer_index: PeerIndex) -> bool`?

src/parsec.rs
```diff
@@ -750,13 +692,13 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
     }
 
     /// Handles consensus reached by us.
-    fn handle_consensus(&mut self, payload_key: &ObservationKey) {
+    fn handle_consensus(&mut self, payload_key: &ObservationKey) -> Option<PeerListChange> {
         match self
             .observations
             .get(payload_key)
             .map(|info| info.observation.clone())
```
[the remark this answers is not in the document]
Fraser999: Yes.

src/parsec.rs
```diff
@@ -1038,7 +950,45 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+
+        // If self-parent is initial, we don't have to check it's meta-event, as we already know it
+        // can not have any observees.
+        if self_parent.is_initial() {
+            return false;
+        }
+
+        // If self-parent is earlier in history than the start of the meta-election, it won't have
+        // a meta-event; but it also means that it wasn't an observer.
+        if self.meta_election.upper_start_index() > self_parent.topological_index() {
```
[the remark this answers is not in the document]
Fraser999: I agree, but I can't think of anything better which is not overly verbose.  Is it enough that these fields have had doc comments added in MetaElection?

src/parsec.rs
```diff
@@ -946,36 +944,16 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-            false
-        }
+        self.graph
+            .self_parent(event)
+            .and_then(|self_parent| {
+                self.meta_election
+                    .meta_event(self_parent.event_index())
+                    .map(|meta_parent| {
+                        meta_parent.is_observer() || meta_parent.has_ancestor_observer()
+                    })
```
jeanphilippeD: I like it.
Would this work, and be clearer because less nesting like this?

```
        self.graph
            .self_parent(event)
            .and_then(|self_parent| self.meta_election.meta_event(self_parent.event_index()))
            .and_then(|meta_parent| meta_parent.is_observer() || meta_parent.has_ancestor_observer())
            .unwrap_or(false)
```

src/dump_graph.rs
```diff
@@ -290,10 +291,12 @@ mod detail {
             );
         }
 
-        let meta_votes: BTreeMap<_, _> = meta_votes
+        let meta_votes = meta_votes
             .iter()
             .map(|(peer_index, meta_votes)| (unwrap!(peer_list.get(peer_index)).id(), meta_votes))
-            .collect();
+            .sorted_by(|(lhs_peer_index, _), (rhs_peer_index, _)| {
```
jeanphilippeD: I got confused here, for a second I thought the behaviour changed.
I think it should be `lhs_peer_id` since it is the `id` not the `index` that we use (distinction between `id` and `index` is its own confusing thing)

Fraser999: Yeah - good catch.  It should have been `_id` indeed.

PR #237, opened 2019-01-24
fix/tests: fix an issue whereby peers' states were not updated

Fraser999:
Upon removing a peer in the test network, we failed to recheck existing peers which were transitioning to Added or Removed to see if the new lower number of members meant that a supermajority was valid now.

Rather than persist with the `PeerRemovalGuard` which effectively represented a "God view" of the network in terms of the joining and leaving states of the peers, there is now a network view held as a
member of the `Peer` struct.

This avoids synchronisation problems and allows better encapsulation of `Peer` by tying the transition between states into `Peer` methods.

This also moves the `PeerStatuses` struct unchanged into its own file to avoid cluttering peer.rs.

### DIFF
--- a/src/dev_utils/network.rs
+++ b/src/dev_utils/network.rs
@@ -76,7 +76,7 @@ pub enum ConsensusError {
 }
 
 impl Network {
-    /// Create an empty test network
+    /// Create an empty test network.
     pub fn new(consensus_mode: ConsensusMode) -> Self {
         Network {
             peers: BTreeMap::new(),
@@ -86,7 +86,7 @@ impl Network {
         }
     }
 
-    /// Create a test network with initial peers constructed from the given IDs
+    /// Create a test network with initial peers constructed from the given IDs.
     pub fn with_peers<I: IntoIterator<Item = PeerId>>(
         all_ids: I,
         consensus_mode: ConsensusMode,
@@ -151,7 +151,7 @@ impl Network {
     }
 
     /// Returns true if all peers hold the same sequence of stable blocks.
-    fn blocks_all_in_sequence(&self) -> Result<(), ConsensusError> {
+    fn check_blocks_all_in_sequence(&self) -> Result<(), ConsensusError> {
         let first_peer = unwrap!(self.active_peers().next());
         let payloads = first_peer.blocks_payloads();
         if let Some(peer) = self
@@ -197,7 +197,7 @@ impl Network {
             });
     }
 
-    /// Handles incoming requests and responses
+    /// Handles incoming requests and responses.
     fn handle_messages(&mut self, peer: &PeerId, step: usize) {
         if let Some(msgs) = self.msg_queue.remove(peer) {
             let (to_handle, rest) = msgs
@@ -330,7 +330,7 @@ impl Network {
         .is_ok()
     }
 
-    /// Checks whether there is a right number of blocks and the blocks are in an agreeing order
+    /// Checks whether there is a right number of blocks and the blocks are in an agreeing order.
     fn check_consensus(
         &self,
         expected_peers: &BTreeMap<PeerId, PeerStatus>,
@@ -361,7 +361,7 @@ impl Network {
         }
 
         // Check everybody has the same blocks in the same order.
-        self.blocks_all_in_sequence()
+        self.check_blocks_all_in_sequence()
     }
 
     fn check_block_signatories(
@@ -401,15 +401,15 @@ impl Network {
         Ok(())
     }
 
-    /// Checks if the blocks are only signed by valid voters
+    /// Checks if the blocks are only signed by valid voters.
     fn check_blocks_signatories(&self) -> Result<(), ConsensusError> {
-        let blocks = self.active_peers().next().unwrap().blocks();
+        let blocks = unwrap!(self.active_peers().next()).blocks();
         let mut valid_voters = BTreeSet::new();
         for block in blocks {
             match *block.payload() {
                 ParsecObservation::Genesis(ref g) => {
-                    // explicitly don't check signatories - the list of valid voters
-                    // should be empty at this point
+                    // Explicitly don't check signatories - the list of valid voters should be empty
+                    // at this point.
                     valid_voters = g.clone();
                 }
                 ParsecObservation::Add { ref peer_id, .. } => {
@@ -445,7 +445,7 @@ impl Network {
         }
     }
 
-    /// Simulates the network according to the given schedule
+    /// Simulates the network according to the given schedule.
     pub fn execute_schedule<R: Rng>(
         &mut self,
         rng: &mut R,
@@ -493,7 +493,7 @@ impl Network {
         match event {
             ScheduleEvent::Genesis(genesis_group) => {
                 if !self.peers.is_empty() {
-                    // if the peers are already initialised, we won't initialise them again
+                    // If the peers are already initialised, we won't initialise them again.
                     return Ok(true);
                 }
                 let peers = genesis_group
@@ -510,22 +510,22 @@ impl Network {
                 }
                 self.peers = peers;
                 self.genesis = genesis_group;
-                // do a full reset while we're at it
+                // Do a full reset while we're at it.
                 self.msg_queue.clear();
             }
-            ScheduleEvent::AddPeer(peer) => {
+            ScheduleEvent::AddPeer(peer_id) => {
                 let current_peers = self.active_peers().map(|peer| peer.id.clone()).collect();
                 let _ = self.peers.insert(
-                    peer.clone(),
+                    peer_id.clone(),
                     Peer::from_existing(
-                        peer.clone(),
+                        peer_id.clone(),
                         &self.genesis,
                         &current_peers,
                         self.consensus_mode,
                     ),
                 );
 
-                peer_removal_guard.add_peer(peer);
+                peer_removal_guard.add_peer(peer_id);
             }
             ScheduleEvent::RemovePeer(peer) => {
                 if peer_removal_guard.attempt_to_remove_peer(&peer) {
--- a/src/dev_utils/network.rs
+++ b/src/dev_utils/network.rs
@@ -8,7 +8,7 @@
 
 #[cfg(feature = "testing")]
 use super::parse_test_dot_file;
-use super::peer::{Peer, PeerStatus};
+use super::peer::{NetworkView, Peer, PeerStatus};
 use super::schedule::{Schedule, ScheduleEvent, ScheduleOptions};
 use super::Observation;
 use crate::block::Block;
@@ -134,20 +134,33 @@ impl Network {
         self.consensus_mode
     }
 
-    fn peers_with_status(&self, status: PeerStatus) -> impl Iterator<Item = &Peer> {
+    fn active_peers(&self) -> impl Iterator<Item = &Peer> {
         self.peers
             .values()
-            .filter(move |&peer| peer.status == status)
+            .filter(|peer| peer.status() == PeerStatus::Active)
     }
 
-    fn active_peers(&self) -> impl Iterator<Item = &Peer> {
-        self.peers_with_status(PeerStatus::Active)
+    /// Returns the IDs of peers which consider themselves to be still running correctly, i.e. those
+    /// for which `is_running()` is true.
+    fn running_peers_ids(&self) -> Vec<PeerId> {
+        self.peers
+            .values()
+            .filter_map(|peer| {
+                if peer.is_running() {
+                    Some(peer.id().clone())
+                } else {
+                    None
+                }
+            })
+            .collect()
     }
 
-    fn present_peers(&self) -> impl Iterator<Item = &Peer> {
-        self.peers.values().filter(move |&peer| {
-            peer.status == PeerStatus::Active || peer.status == PeerStatus::Pending
-        })
+    /// Returns the number of peers for which the network has the given view of their state.
+    fn num_with_network_view(&self, network_view: NetworkView) -> usize {
+        self.peers
+            .values()
+            .filter(|peer| peer.network_view() == network_view)
+            .count()
     }
 
     /// Returns true if all peers hold the same sequence of stable blocks.
@@ -160,11 +173,11 @@ impl Network {
         {
             Err(ConsensusError::DifferingBlocksOrder {
                 order_1: BlocksOrder {
-                    peer: first_peer.id.clone(),
+                    peer: first_peer.id().clone(),
                     order: payloads.into_iter().cloned().collect(),
                 },
                 order_2: BlocksOrder {
-                    peer: peer.id.clone(),
+                    peer: peer.id().clone(),
                     order: peer.blocks_payloads().into_iter().cloned().collect(),
                 },
             })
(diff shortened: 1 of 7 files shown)


src/dev_utils/network.rs
```diff
@@ -401,15 +412,15 @@ impl Network {
         Ok(())
     }
 
-    /// Checks if the blocks are only signed by valid voters
+    /// Checks if the blocks are only signed by valid voters.
     fn check_blocks_signatories(&self) -> Result<(), ConsensusError> {
-        let blocks = self.active_peers().next().unwrap().blocks();
+        let blocks = unwrap!(self.active_peers().next()).blocks();
```
jeanphilippeD: Why the change? does it do the same thing?

src/dev_utils/network.rs
```diff
@@ -505,216 +514,82 @@ impl Network {
...
-                }
                 self.peers = peers;
                 self.genesis = genesis_group;
-                // do a full reset while we're at it
+                // Do a full reset while we're at it.
                 self.msg_queue.clear();
             }
-            ScheduleEvent::AddPeer(peer) => {
-                let current_peers = self.active_peers().map(|peer| peer.id.clone()).collect();
+            ScheduleEvent::AddPeer(peer_id) => {
```
jeanphilippeD: There are a lot nice small refactoring like this.
It would be good if they were all in a refactoring commit because it adds a lot of noise to the diff.

src/dev_utils/network.rs
```diff
@@ -505,216 +514,82 @@ impl Network {
...
-                        }
-                    }
-
+                for peer_id in self.running_peers_ids() {
+                    self.peer_mut(&peer_id).make_votes();
+                    self.handle_messages(&peer_id, step);
+                    self.peer_mut(&peer_id).poll_all();
+                    self.check_unexpected_accusations(&peer_id)?;
+                }
+                Peer::update_network_views(&mut self.peers);
```
jeanphilippeD: Same comment, but for bigger refactoring it is even more important: Quite difficult to see what changed in the functionality of update_network_views, but would be trivial if we split the change in 2 steps.

src/dev_utils/peer.rs
```diff
@@ -11,27 +11,58 @@ use super::Observation;
...
+    is_more_than_two_thirds, ConsensusMode, Malice, Observation as ParsecObservation,
+};
 use crate::parsec::Parsec;
-use rand::Rng;
 use std::collections::{BTreeMap, BTreeSet};
 use std::fmt::{self, Debug, Formatter};
 
+/// This represents the peer's own view of its current status.
+///
+/// A new peer will start as `Pending` and transition to `Active` once `Peer::poll()` yields a block
```
jeanphilippeD: I'm happy with this comment, it is really helpful.
I wonder if it would be clearer to describe this small state machine with ASCII art?

src/dev_utils/peer.rs
```diff
@@ -11,27 +11,58 @@ use super::Observation;
...
-    Active,
     Pending,
+    Active,
     Removed,
     Failed,
 }
 
+/// This represents the network's view of a peer's state.
+///
+/// A new peer will start as `Joining`.  While in this state, its own status may or may not
```
jeanphilippeD: Same

src/dev_utils/peer.rs
```diff
@@ -97,175 +137,152 @@ impl Peer {
...
+        let mut removed_counts = BTreeMap::new();
+        let mut running_peers_count = 0;
+        for peer in all_peers
+            .values()
+            .filter(|peer| peer.network_view == NetworkView::Joined)
+        {
+            for added_peer_id in &peer.added_peers_ids {
+                let count = added_counts.entry(added_peer_id.clone()).or_insert(0);
+                *count += 1;
+            }
```
jeanphilippeD: I'm not super keen on for within for.

Maybe:
```
let add_count = |new_ids, counts| {
            for id in new_ids {
                let count = counts.entry(id.clone()).or_insert(0);
                *count += 1;
            }
};
...
   add_count( &peer.added_peers_ids, &mut added_counts);
   add_count( &peer.removed_peers_ids, &mut removed_counts);
```


Also maybe in addition can avoid loop for updating ids with something like:
```
let add_count = |new_ids, counts| {
    new_ids.map( |id| {
        counts.entry(id.clone())
           .and_modify(|c| { *c += 1 })
           .or_insert(1);
    } );
};
```

src/dev_utils/peer_statuses.rs
```diff
@@ -0,0 +1,140 @@
...
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use super::PeerStatus;
+use crate::mock::PeerId;
+use rand::Rng;
+use std::collections::{BTreeMap, BTreeSet};
+
+pub struct PeerStatuses {
```
jeanphilippeD: Same, this move/refactor should probably have been done in a separate commit.
I cannot really see what changed.

src/dev_utils/network.rs
Fraser999: The macro provides a much more useful error message.  It's been a policy to prefer the macro for quite a while now: https://github.com/maidsafe/QA/blob/master/Documentation/Rust%20Style.md#unwrap

src/dev_utils/network.rs
Fraser999: I agree for larger changes (e.g. if rustfmt changes a bunch of files), but for very few, trivial changes like this, I feel it slows down the author more than the reviewers to burst a PR into more commits.

src/dev_utils/network.rs
Fraser999: I don't feel this falls into the same category as the previous comment.  It's not a trivial renaming of a variable or function; rather it's a key element of the refactoring effort.

src/dev_utils/peer.rs
Fraser999: I'd be happy to add that, but not quite sure what you had in mind.  The better we can document these states, the easier future debugging will be!

src/dev_utils/network.rs
jeanphilippeD: I generally think this is a tooling issue.
Splitting a commit in multiple parts should be a trivial operation (5/15 minutes for a change like this + run minimal testing), so the author's time should not be much of a consideration here.

I would agree if the rest of the code change was small, but since the code change big as well, removing anything not relevant is really important so the review can be effective (not only faster).

Also, help keep the history clean so when we go back we can understand what it does.

src/dev_utils/peer.rs
Fraser999: Well, I like the deduplication aspect of the former option you provided (I considered something like this but rejected the idea as it's a fairly basic two-line function).  I guess I'd still view that as a `for` within `for`, but then I don't have an issue with nesting `for`s, unless the total nested depth gets big (but that applies to any code, not just `for`s).

I can't say I'm a fan of the second option; I think it's mildly harder to read.

src/dev_utils/peer_statuses.rs
Fraser999: I mentioned in the PR comment that nothing changed.

src/dev_utils/peer.rs
jeanphilippeD: I had something like that in mind. But mostly putting it to show the important info and transition.

```
/// Peer Start
/// |-> `Pending`
///      |
///    Peer::poll()` yields a block)
///      |->`Active`
///          |
///        Need to kill peer
///          |-> `Removed`
///          |-> `Failed`
```

src/dev_utils/peer.rs
```diff
@@ -11,27 +11,58 @@ use super::Observation;
...
     Removed,
     Failed,
 }
 
+/// This represents the network's view of a peer's state.
+///
+/// A new peer will start as `Joining`.  While in this state, its own status may or may not
+/// transition from `PeerStatus::Pending` to `PeerStatus::Active`.  Once a supermajority of
+/// currently `Joined` peers have reached consensus on this peer being added, its network view will
+/// change to `Joined`.
```
maqi: It might be worth to mention that while a peer is in NetworkView::Joining, it won't be allowed to be removed.

src/dev_utils/network.rs
Fraser999: I disagree in this particular case as the cleanups are so small and unobtrusive.  However, I can accept that I'm in the minority here.

src/dev_utils/peer.rs
Fraser999: Added in a fixup commit.

src/dev_utils/peer.rs
Fraser999: That's a good point.  I think we can actually enforce that in the code too by checking the peer's network state when calling the `mark_as_...` methods.

src/dev_utils/peer.rs
```diff
@@ -40,9 +50,22 @@ pub enum PeerStatus {
...
+/// will change to `Left`.
+///
+/// Peer start
+///     |-> `Joining`
+///     |
+///     Supermajority of `Joined` peers see `Add` for this one
+///         |-> `Joined`
+///         |
+///         Any peer votes to remove this one OR this peer is killed
+///             |-> `Leaving`
```
maqi: In case we will need to similuate the scenario that a malicious node voted to `remove` a peer, will this `any peer votes to remove` triggers an incorrect updated in the network view?

Fraser999: Well, it should trigger a panic hopefully.  But I don't think we should be catering to that yet, since all our malice scenarios so far are being created by parsing dot files and/or passing malicious messages rather than running actual malicious test nodes.

src/dev_utils/peer.rs
```diff
@@ -97,175 +164,154 @@ impl Peer {
...
     }
 
-    /// Randomly chooses a peer to remove.
-    pub fn remove_random_peer<R: Rng>(&mut self, rng: &mut R, min_active: usize) -> Option<PeerId> {
-        let name = self.choose_name_to_remove(rng);
-
-        let mut active_peers = self.num_active_peers();
-        let mut failed_peers = self.num_failed_peers();
+    // Consider only the peers which the network views as being `Joined`, and check to see if a
+    // supermajority of these have polled blocks which allow the network to update its view of any
```
jeanphilippeD: Why supermajority?
Why not all, or 50%?

src/dev_utils/peer.rs
```diff
@@ -97,175 +164,154 @@ impl Peer {
...
+    // `Joining` or `Leaving` peers, transitioning these to `Joined` or `Left` respectively.
+    pub fn update_network_views(all_peers: &mut BTreeMap<PeerId, Peer>) {
+        let mut added_counts = BTreeMap::new();
+        let mut removed_counts = BTreeMap::new();
+        let mut running_peers_count = 0;
```
jeanphilippeD: Should these declaration be moved just above where they are used?

src/dev_utils/peer.rs
```diff
@@ -97,175 +164,154 @@ impl Peer {
...
     }
 
-    pub fn fail_peer(&mut self, peer: &PeerId) {
-        let status = unwrap!(self.statuses.get_mut(peer));
-        *status = PeerStatus::Failed;
+    pub fn is_active_and_has_block(&self, payload: &Observation) -> bool {
+        if self.status != PeerStatus::Active {
+            return false;
+        }
+        self.blocks.iter().any(|block| block.payload() == payload)
```
jeanphilippeD: self.status == PeerStatus::Active &&
   self.blocks.iter().any(|block| block.payload() == payload)

src/dev_utils/network.rs
```diff
@@ -134,20 +134,33 @@ impl Network {
...
             .values()
-            .filter(move |&peer| peer.status == status)
+            .filter(|peer| peer.status() == PeerStatus::Active)
     }
 
-    fn active_peers(&self) -> impl Iterator<Item = &Peer> {
-        self.peers_with_status(PeerStatus::Active)
+    /// Returns the IDs of peers which consider themselves to be still running correctly, i.e. those
+    /// with `PeerStatus::Pending/Active`.
+    fn running_peers_ids(&self) -> Vec<PeerId> {
```
jeanphilippeD: /// Returns IDs of peer thinking they are running.

Not sure the comment really help though.
Also PeerStatus::Pending/Active`. is an implementation detail of is_running so it is liable to becoming out of date if is_running is changed and should not be mentioned here.

Fraser999: I take your point about about `PeerStatus::Pending/Active` being an implementation detail, but disagree with the rest of your comment.

I'll update the code comment to refer to `is_running()`.

src/dev_utils/peer.rs
```diff
@@ -97,175 +164,154 @@ impl Peer {
...
-        self.peers_by_status(|s| *s == PeerStatus::Active || *s == PeerStatus::Pending)
-            .map(|(id, _)| id)
-    }
-
-    /// Returns an iterator through the list of inactive peers (removed and failed).
-    pub fn inactive_peers(&self) -> impl Iterator<Item = &PeerId> {
-        self.peers_by_status(|s| *s == PeerStatus::Removed || *s == PeerStatus::Failed)
-            .map(|(id, _)| id)
+    /// Sets the node's own status to `Removed` and the network's view of it to `Leaving`.  Panics
+    /// if the node in question isn't yet viewed by the network as being `Joined`.
```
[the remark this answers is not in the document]
Fraser999: I can revise the comment to something similar to what you suggest:
```
/// Update status and network view when peer is removed from the the network.
```
I don't see that as much of an improvement however, since I'd still keep the part about panicking if not joined.

Another option is to remove the comments for `mark_as_resolved()` and `mark_as_failed()` if others feel they're unhelpful.

src/dev_utils/peer.rs
```diff
@@ -97,175 +164,154 @@ impl Peer {
...
+    pub fn mark_as_removed(&mut self) {
+        if self.status == PeerStatus::Failed {
```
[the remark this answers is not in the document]
Fraser999: This behaviour is part of `self.mark_as_leaving();`

src/dev_utils/peer.rs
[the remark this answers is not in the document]
jeanphilippeD: Maybe the panic part could be more concise and says:
`/// Panic if self network_view is Joining.`

Or:
`/// Expect self network_view is not 'Joining'.`

src/dev_utils/peer.rs
```diff
@@ -97,175 +164,154 @@ impl Peer {
...
-    /// Randomly chooses a peer to remove.
-    pub fn remove_random_peer<R: Rng>(&mut self, rng: &mut R, min_active: usize) -> Option<PeerId> {
-        let name = self.choose_name_to_remove(rng);
-
-        let mut active_peers = self.num_active_peers();
-        let mut failed_peers = self.num_failed_peers();
+    // Consider only the peers which the network views as being `Joined`, and check to see if a
+    // supermajority of these have polled blocks which allow the network to update its view of any
+    // `Joining` or `Leaving` peers, transitioning these to `Joined` or `Left` respectively.
+    pub fn update_network_views(all_peers: &mut BTreeMap<PeerId, Peer>) {
```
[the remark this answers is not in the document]
Fraser999: I prefer the second suggestion; I'll use that.

src/dev_utils/peer.rs
Fraser999: They are already, aren't they?

jeanphilippeD: I guess I was thinking above the loop, but that make sense here.

PR #239, opened 2019-01-31
refactor/parsec remove EventContextMut

jeanphilippeD:
Refactor to simplify future work, particularly with performance.

In order to dump all events received on dump-graph, we may want to
unpack all events up-front. Using EventContextMut made it unclear whether it
was safe.

Also, observation is populated on unpack, but we may not add the event
to the graph in case of malice. This maybe something we want to change.

Test:
cargo test --features=testing,malice-detection

### DIFF
--- a/src/dev_utils/dot_parser.rs
+++ b/src/dev_utils/dot_parser.rs
@@ -6,12 +6,17 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use crate::gossip::{CauseInput, Event, EventIndex, Graph, IndexedEventRef};
 #[cfg(any(
     all(test, feature = "malice-detection", feature = "mock"),
     feature = "testing"
 ))]
-use crate::gossip::{EventContextMut, EventContextRef};
+use crate::error::Error;
+#[cfg(any(
+    all(test, feature = "malice-detection", feature = "mock"),
+    feature = "testing"
+))]
+use crate::gossip::EventContextRef;
+use crate::gossip::{CauseInput, Event, EventIndex, Graph, IndexedEventRef};
 use crate::hash::Hash;
 use crate::hash::HASH_LEN;
 use crate::meta_voting::{BoolSet, MetaElection, MetaEvent, MetaVote, Step};
@@ -680,6 +685,7 @@ impl ParsedContents {
             graph: &self.graph,
             peer_list: &self.peer_list,
             observations: &self.observations,
+            consensus_mode: ConsensusMode::Supermajority,
         }
     }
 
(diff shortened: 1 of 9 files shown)


src/parsec.rs
```diff
@@ -612,6 +607,43 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+    ) -> Result<UnpackedEvent<T, S::PublicId>> {
+        match Event::unpack(packed_event, &forking_peers, self.event_context())? {
+            UnpackedEvent::New(event, observation) => {
+                if let Some((payload_key, observation_info)) = observation {
+                    let _ = self
+                        .observations
+                        .entry(payload_key)
+                        .or_insert_with(|| observation_info);
+                }
+                Ok(UnpackedEvent::New(event, None))
```
Fraser999: This seemed a little surprising to me at first reading.  Possibly things could be more explicit if we add the following impl to `UnpackedEvent`:
```rust
impl<T: NetworkEvent, P: PublicId> UnpackedEvent<T, P> {
    pub fn take_observation(&mut self) -> ObservationForStore<T, P> {
        match self {
            UnpackedEvent::New(_, observation) => observation.take(),
            UnpackedEvent::Known(_) => None,
        }
    }
}
```
Then we can use it here:
```rust
let mut unpacked = Event::unpack(packed_event, &forking_peers, self.event_context())?;
if let Some((payload_key, observation_info)) = unpacked.take_observation() {
...
}
Ok(unpacked)
```

src/parsec.rs
```diff
@@ -612,6 +607,43 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+            UnpackedEvent::Known(index) => Ok(UnpackedEvent::Known(index)),
+        }
+    }
+
+    fn new_event_from_observation(
+        &mut self,
+        self_parent: EventIndex,
+        observation: Observation<T, S::PublicId>,
+    ) -> Result<Event<S::PublicId>> {
+        let (event, observation) =
```
Fraser999: Again here I think using `observation` as the name for the `ObservationForStore` variable is confusing given that we do have an actual `observation` of type `Observation` already present.

src/dev_utils/dot_parser.rs
```diff
@@ -5,13 +5,17 @@
 // under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
-
```
Fraser999: All our other source files have a blank line here I believe.

src/parsec.rs
```diff
@@ -612,29 +612,27 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-                    let _ = self
-                        .observations
-                        .entry(payload_key)
-                        .or_insert_with(|| observation_info);
-                }
-                Ok(UnpackedEvent::New(event, None))
-            }
-            UnpackedEvent::Known(index) => Ok(UnpackedEvent::Known(index)),
+        let unpacked_event = Event::unpack(packed_event, &forking_peers, &self.event_context())?;
+        let (unpacked_event, observation_for_store) = unpacked_event.split_off_observation();
```
Fraser999: I still prefer my original suggestion.  It seems cleaner to me than using shadowing of `unpacked_event` to give the appearance of immutability.

PR #242, opened 2019-02-04
Perf fix vote collection

jeanphilippeD:
Address performance issue related to set_meta_votes.

The first 2 commits apply smaller updates close to refactoring, but they do change behavior:
-First use references to vector instead of cloning the contents.
-Then: like for parent_meta_votes ignore empty meta_votes everywhere meta_votes was called: rename to populated_meta_votes

Finally Update code to avoid duplicate lookup highlighted by 'perf' and 'callgrind'
```
RUSTFLAGS=-g RUST_BACKTRACE=1 cargo bench --features=testing -- --baseline=master-da5b291

minimal                 time:   [726.91 us 744.71 us 764.01 us]
                        change: [-17.884% -14.460% -10.985%] (p = 0.00 < 0.05)
                        Performance has improved.

static                  time:   [3.7173 ms 3.7292 ms 3.7455 ms]
                        change: [-12.445% -12.002% -11.570%] (p = 0.00 < 0.05)
                        Performance has improved.

dynamic                 time:   [2.4579 ms 2.4616 ms 2.4669 ms]
                        change: [-9.6652% -8.6950% -7.4378%] (p = 0.00 < 0.05)
                        Performance has improved.

a_node4_opaque_evt8     time:   [1.4799 ms 1.4824 ms 1.4860 ms]
                        change: [-9.7467% -8.8776% -7.9721%] (p = 0.00 < 0.05)
                        Performance has improved.

a_node8_opaque_evt8     time:   [8.3695 ms 8.4027 ms 8.4609 ms]
                        change: [-25.324% -24.511% -23.396%] (p = 0.00 < 0.05)
                        Performance has improved.

a_node12_opaque_evt8    time:   [31.629 ms 31.786 ms 31.915 ms]
                        change: [-32.982% -32.524% -32.065%] (p = 0.00 < 0.05)
                        Performance has improved.

a_node16_opaque_evt8    time:   [59.365 ms 59.632 ms 59.896 ms]
                        change: [-38.843% -38.333% -37.779%] (p = 0.00 < 0.05)
                        Performance has improved.

a_node24_opaque_evt8    time:   [207.65 ms 209.20 ms 210.63 ms]
                        change: [-44.275% -43.856% -43.450%] (p = 0.00 < 0.05)
                        Performance has improved.

a_node32_opaque_evt8    time:   [683.51 ms 707.38 ms 732.72 ms]
                        change: [-53.254% -52.011% -50.657%] (p = 0.00 < 0.05)
                        Performance has improved.

a_node4_opaque_evt16    time:   [2.9375 ms 3.0356 ms 3.1498 ms]
                        change: [-13.487% -8.0745% -2.7411%] (p = 0.02 < 0.05)
                        Performance has improved.

a_node8_opaque_evt16    time:   [21.749 ms 21.877 ms 22.117 ms]
                        change: [-26.768% -24.192% -21.836%] (p = 0.00 < 0.05)
                        Performance has improved.

a_node12_opaque_evt16   time:   [70.469 ms 70.733 ms 71.084 ms]
                        change: [-33.807% -30.776% -28.163%] (p = 0.00 < 0.05)
                        Performance has improved.

a_node16_opaque_evt16   time:   [166.09 ms 166.51 ms 166.89 ms]
                        change: [-35.255% -34.316% -33.516%] (p = 0.00 < 0.05)
                        Performance has improved.

a_node24_opaque_evt16   time:   [540.04 ms 544.97 ms 552.49 ms]
                        change: [-36.192% -35.299% -34.419%] (p = 0.00 < 0.05)
                        Performance has improved.

a_node32_opaque_evt16   time:   [1.3376 s 1.3637 s 1.3879 s]
                        change: [-39.308% -38.306% -37.251%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/src/meta_voting/meta_election.rs
+++ b/src/meta_voting/meta_election.rs
@@ -88,10 +88,15 @@ impl MetaElection {
         self.meta_events.get(&event_index)
     }
 
-    pub fn meta_votes(&self, event_index: EventIndex) -> Option<&PeerIndexMap<Vec<MetaVote>>> {
+    /// The event meta votes if available
+    pub fn populated_meta_votes(
+        &self,
+        event_index: EventIndex,
+    ) -> Option<&PeerIndexMap<Vec<MetaVote>>> {
         self.meta_events
             .get(&event_index)
             .map(|meta_event| &meta_event.meta_votes)
+            .filter(|meta_votes| !meta_votes.is_empty())
     }
 
     pub fn round_hashes(&self, peer_index: PeerIndex) -> Option<&Vec<RoundHash>> {
--- a/src/meta_voting/meta_vote.rs
+++ b/src/meta_voting/meta_vote.rs
@@ -74,7 +74,7 @@ impl Debug for MetaVote {
 impl MetaVote {
     pub fn new(
         initial_estimate: bool,
-        others: &[Vec<MetaVote>],
+        others: &[&[MetaVote]],
         total_peers: usize,
         is_voter: bool,
     ) -> Vec<Self> {
@@ -87,7 +87,7 @@ impl MetaVote {
 
     pub fn next(
         parent: &[MetaVote],
-        others: &[Vec<MetaVote>],
+        others: &[&[MetaVote]],
         coin_tosses: &BTreeMap<usize, bool>,
         total_peers: usize,
         is_voter: bool,
@@ -109,6 +109,10 @@ impl MetaVote {
             next.push(next_meta_vote);
         }
         next
+    }
+
+    pub fn round_and_step(&self) -> (usize, Step) {
+        (self.round, self.step)
     }
 
     fn update_meta_vote(
@@ -138,7 +142,7 @@ impl MetaVote {
 
     fn next_meta_vote(
         parent: Option<&MetaVote>,
-        others: &[Vec<MetaVote>],
+        others: &[&[MetaVote]],
         coin_tosses: &BTreeMap<usize, bool>,
         total_peers: usize,
         is_voter: bool,
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -31,7 +31,7 @@ use crate::observation::{
     ObservationStore,
 };
 use crate::parsec_helpers::find_interesting_content_for_event;
-use crate::peer_list::{PeerIndex, PeerIndexSet, PeerList, PeerState};
+use crate::peer_list::{PeerIndex, PeerIndexMap, PeerIndexSet, PeerList, PeerState};
 use std::collections::{BTreeMap, BTreeSet, VecDeque};
 use std::mem;
 #[cfg(all(test, feature = "mock"))]
@@ -1214,15 +1214,14 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         peer_index: PeerIndex,
         round: usize,
     ) -> Option<bool> {
-        self.meta_votes_since_round_and_step(
-            creator,
-            creator_event_index,
-            peer_index,
-            round,
-            Step::GenuineFlip,
-        )
-        .next()
-        .and_then(|meta_vote| meta_vote.aux_value)
+        self.meta_votes_by_creator(creator, creator_event_index)
+            .and_then(|meta_votes| meta_votes.get(peer_index))
+            .and_then(|votes| {
+                votes
+                    .iter()
+                    .find(|meta_vote| meta_vote.round_and_step() >= (round, Step::GenuineFlip))
+            })
+            .and_then(|meta_vote| meta_vote.aux_value)
     }
 
     // Skips back through events created by the peer until passed `responsiveness_threshold`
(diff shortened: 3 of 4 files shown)


src/meta_voting/meta_election.rs
```diff
@@ -88,10 +88,15 @@ impl MetaElection {
         self.meta_events.get(&event_index)
     }
 
-    pub fn meta_votes(&self, event_index: EventIndex) -> Option<&PeerIndexMap<Vec<MetaVote>>> {
+    /// The event meta votes if available
```
maqi: returns the meta votes of the event if available in the formant of PeerIndexMap?
just wondering, does PeerIndexMap mean it will be per-peer's perspective?

src/parsec.rs
```diff
@@ -1029,57 +1030,52 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-                if !parent_meta_votes.is_empty() {
-                    Some(parent_meta_votes)
-                } else {
-                    None
-                }
-            });
+            .and_then(|parent_hash| self.meta_election.populated_meta_votes(parent_hash));
+
+        if parent_meta_votes.is_none() && !self.is_observer(builder) {
+            // No meta votes to set for this event
```
maqi: better to mention the reason for not setting for this event is that we are not observer && parent doesn't have meta_votes.

src/parsec.rs
```diff
@@ -1029,57 +1030,52 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+            return Ok(());
+        }
+
+        let voters = self.voters();
+        let is_voter = voters.contains(builder.event().creator());
```
maqi: shall the comment of `For the case that event's creator is not a voter, the initial estimation shall not becreated.` be moved to here?

src/parsec.rs
```diff
@@ -1234,64 +1229,55 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+                return self.meta_election.populated_meta_votes(event);
+            }
+        }
+        None
     }
 
-    // Returns the set of meta votes held by all peers other than the creator of `event` which are
-    // votes by the peer at `peer_index`.
-    fn collect_other_meta_votes(
+    // All the event's voting ancestors meta votes excluding self_parent.
```
maqi: hmm...  the comment seems doesn't match the code?

src/meta_voting/meta_election.rs
jeanphilippeD: Note that this changeset does not change that.

Yes, it does contain meta votes for each peers.
Hence the usage to get meta votes for a specific event:
```
self.meta_election
            .meta_votes(event_index)
            .populated_meta_votes(event_index)
            .and_then(|meta_votes| meta_votes.get(event.creator()))
```

src/parsec.rs
jeanphilippeD: Is that not repeating 'if' condition that is already clear?

src/parsec.rs
jeanphilippeD: I do not think this comment is accurate for both branch (`MetaVote::next` & `MetaVote::new`), so it would probably need a caveat that it is true only for new observer, and then may not be a very helpful comment.

In the end I thought the callees (`MetaVote::next` & `MetaVote::new`) should document why they need that flag if it is important. (But I don't know exactly why they do, this PR should not change the parameter these function are called with, so it is seemed ok)

src/parsec.rs
jeanphilippeD: As discussed, will fix wtih:
```
// Returns all the meta votes from the event's voting ancestors except the event's creator.
```

src/meta_voting/meta_vote.rs
```diff
@@ -74,7 +74,7 @@ impl Debug for MetaVote {
 impl MetaVote {
     pub fn new(
         initial_estimate: bool,
-        others: &[Vec<MetaVote>],
+        others: &[&Vec<MetaVote>],
```
pierrechevalier83: Looks a bit unidiomatic to take a `&Vec`. Wouldn't a slice to the trick? (Applies to rest of the commit)

src/parsec.rs
```diff
@@ -1293,9 +1294,19 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
                     .and_then(|creator_event_index| {
                         self.meta_votes_by_creator(creator, *creator_event_index)
                     })
-                    .and_then(|meta_votes| meta_votes.get(peer_index))
             })
-            .collect()
+            .collect_vec()
```
pierrechevalier83: I don't see the point in using `collect_vec` instead of `collect` since the return type already makes it obvious to humans and compilers

src/parsec.rs
```diff
@@ -1293,9 +1294,19 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+
+    // Collect the vectors of meta votes for the peer
+    fn peer_meta_votes<'a>(
+        meta_votes_maps: &'a [&PeerIndexMap<Vec<MetaVote>>],
+        peer_index: PeerIndex,
+    ) -> Vec<&'a Vec<MetaVote>> {
+        meta_votes_maps
+            .iter()
+            .filter_map(|meta_votes| meta_votes.get(peer_index))
+            .collect_vec()
```
pierrechevalier83: same  point about `collect_vec`.

PR #245, opened 2019-02-09
Dot files with generic peer ids

jeanphilippeD:
Allow replay dump graphs that contain peer Ids not in our list of hard coded names.
It is the initial step in replaying dump graph from routing tests, but does not handle peer ids with the same initial (multi-char short name), or non alpha numerical characters yet(id sanitizing).

Add smoke test to support this current and future development.
Note: .dot file does not contain everything, so we need to ignore some things in our comparison:
- consensus history: the .dot file does not contain the last consensused block as the dump occurs before it is added to the history
- The graph of events may contain a fake request that send us the remaining gossip events if the dump graph was produced on a gossip event that was not ours (consensus when processing another peer gossip event).

### DIFF
--- a/src/dev_utils/dot_parser.rs
+++ b/src/dev_utils/dot_parser.rs
@@ -35,6 +35,7 @@ use pom::char_class::{alphanum, digit, hex_digit, multispace, space};
 use pom::parser::*;
 use pom::Result as PomResult;
 use pom::{DataInput, Parser};
+use std::cell::RefCell;
 use std::collections::{BTreeMap, BTreeSet};
 use std::fs::File;
 use std::io::{self, Read};
@@ -76,16 +77,45 @@ struct ParsedFile {
     meta_election: ParsedMetaElection,
 }
 
-type ParserCtxRc = Rc<ParserCtx>;
-
-struct ParserCtx {}
+struct ParserCtx {
+    peer_ids: RefCell<BTreeMap<String, PeerId>>,
+}
 
 impl ParserCtx {
     fn new_rc() -> Rc<Self> {
-        Rc::new(ParserCtx {})
+        Rc::new(ParserCtx {
+            peer_ids: RefCell::new(BTreeMap::new()),
+        })
+    }
+
+    fn populate_peer_ids(&self, peer_list: &ParsedPeerList) {
+        *self.peer_ids.borrow_mut() = peer_list
+            .0
+            .keys()
+            .map(|id| (id.id().to_string(), id.clone()))
+            .collect();
+    }
+
+    fn peer_id_from_short_name(&self, short_name: char) -> PeerId {
+        use std::ops::Bound::{Included, Unbounded};
+
+        let peer_ids = self.peer_ids.borrow();
+        let search_short_name = short_name.to_string();
+
+        // The full name will be the first greater or equal full name
+        let mut range = peer_ids.range((Included(search_short_name), Unbounded));
+        if let Some((_name, peer_id)) = range.next() {
+            return peer_id.clone();
+        }
+
+        panic!(
+            "cannot find a name starts with {:?} within {:?}",
+            short_name, peer_ids
+        )
     }
 }
-fn parse_file(ctx: &ParserCtxRc) -> Parser<u8, ParsedFile> {
+
+fn parse_file(ctx: &Rc<ParserCtx>) -> Parser<u8, ParsedFile> {
     (parse_our_id() + parse_peer_list(ctx) + parse_graph() + parse_meta_election(ctx) - parse_end())
         .map(|(((our_id, peer_list), graph), meta_election)| ParsedFile {
             our_id,
@@ -110,12 +140,18 @@ fn parse_our_id() -> Parser<u8, PeerId> {
 #[derive(Debug)]
 struct ParsedPeerList(BTreeMap<PeerId, PeerState>);
 
-fn parse_peer_list(_ctx: &ParserCtxRc) -> Parser<u8, ParsedPeerList> {
+fn parse_peer_list(ctx: &Rc<ParserCtx>) -> Parser<u8, ParsedPeerList> {
     let list_defs =
         comment_prefix() * seq(b"peer_list: {") * next_line() * parse_peer().repeat(0..)
             - comment_prefix()
             - sym(b'}') * next_line();
-    list_defs.map(|defs| ParsedPeerList(defs.into_iter().collect()))
+
+    let ctx = Rc::clone(ctx);
+    list_defs.map(move |defs| {
+        let peer_list = ParsedPeerList(defs.into_iter().collect());
+        ctx.populate_peer_ids(&peer_list);
+        peer_list
+    })
 }
 
 fn parse_peer() -> Parser<u8, (PeerId, PeerState)> {
@@ -355,7 +391,7 @@ struct ParsedMetaEvent {
     meta_votes: BTreeMap<PeerId, Vec<MetaVote>>,
 }
 
-fn parse_meta_election(ctx: &ParserCtxRc) -> Parser<u8, ParsedMetaElection> {
+fn parse_meta_election(ctx: &Rc<ParserCtx>) -> Parser<u8, ParsedMetaElection> {
     seq(b"/// ===== meta-elections =====")
         * next_line()
         * (parse_consensus_history() - next_line()
@@ -526,7 +562,7 @@ fn parse_transaction() -> Parser<u8, String> {
 }
 
 fn parse_meta_events(
-    ctx: &ParserCtxRc,
+    ctx: &Rc<ParserCtx>,
 ) -> Parser<u8, BTreeMap<String, (ObservationMap, ParsedMetaEvent)>> {
     (comment_prefix()
         * seq(b"meta_events: {")
@@ -539,7 +575,7 @@ fn parse_meta_events(
 }
 
 fn parse_single_meta_event(
-    ctx: &ParserCtxRc,
+    ctx: &Rc<ParserCtx>,
 ) -> Parser<u8, (String, (ObservationMap, ParsedMetaEvent))> {
     comment_prefix() * parse_event_id() - seq(b" -> {") - next_line()
         + parse_meta_event_content(ctx)
@@ -548,7 +584,7 @@ fn parse_single_meta_event(
         - next_line()
 }
 
-fn parse_meta_event_content(ctx: &ParserCtxRc) -> Parser<u8, (ObservationMap, ParsedMetaEvent)> {
+fn parse_meta_event_content(ctx: &Rc<ParserCtx>) -> Parser<u8, (ObservationMap, ParsedMetaEvent)> {
     (parse_observees() + parse_interesting_content() + parse_meta_votes(ctx).opt()).map(
         |((observees, observation_map), meta_votes)| {
             let interesting_content = observation_map.iter().map(|(key, _)| *key).collect();
@@ -584,7 +620,7 @@ fn parse_interesting_content() -> Parser<u8, ObservationMap> {
     })
 }
 
-fn parse_meta_votes(ctx: &ParserCtxRc) -> Parser<u8, BTreeMap<PeerId, Vec<MetaVote>>> {
+fn parse_meta_votes(ctx: &Rc<ParserCtx>) -> Parser<u8, BTreeMap<PeerId, Vec<MetaVote>>> {
     (comment_prefix()
         * seq(b"meta_votes: {")
         * next_line()
@@ -596,15 +632,17 @@ fn parse_meta_votes(ctx: &ParserCtxRc) -> Parser<u8, BTreeMap<PeerId, Vec<MetaVo
     .map(|v| v.into_iter().collect())
 }
 
-fn parse_peer_meta_votes(_ctx: &ParserCtxRc) -> Parser<u8, (PeerId, Vec<MetaVote>)> {
+fn parse_peer_meta_votes(ctx: &Rc<ParserCtx>) -> Parser<u8, (PeerId, Vec<MetaVote>)> {
     let peer_line = comment_prefix() * is_a(alphanum).map(char::from) - seq(b": ")
         + parse_meta_vote()
         - next_line();
     let next_line = comment_prefix() * parse_meta_vote() - next_line();
-    (peer_line + next_line.repeat(0..)).map(|((peer_initial, first_mv), other_mvs)| {
+    let ctx = Rc::clone(ctx);
+
+    (peer_line + next_line.repeat(0..)).map(move |((peer_short_name, first_mv), other_mvs)| {
         let mut mvs = vec![first_mv];
         mvs.extend(other_mvs);
-        (PeerId::from_initial(peer_initial), mvs)
+        (ctx.peer_id_from_short_name(peer_short_name), mvs)
     })
 }
 
--- a/src/dev_utils/dot_parser.rs
+++ b/src/dev_utils/dot_parser.rs
@@ -39,6 +39,7 @@ use std::collections::{BTreeMap, BTreeSet};
 use std::fs::File;
 use std::io::{self, Read};
 use std::path::Path;
+use std::rc::Rc;
 use std::str::FromStr;
 
 pub const HEX_DIGITS_PER_BYTE: usize = 2;
@@ -75,15 +76,23 @@ struct ParsedFile {
     meta_election: ParsedMetaElection,
 }
 
-fn parse_file() -> Parser<u8, ParsedFile> {
-    (parse_our_id() + parse_peer_list() + parse_graph() + parse_meta_election() - parse_end()).map(
-        |(((our_id, peer_list), graph), meta_election)| ParsedFile {
+type ParserCtxRc = Rc<ParserCtx>;
+
+struct ParserCtx {}
+
+impl ParserCtx {
+    fn new_rc() -> Rc<Self> {
+        Rc::new(ParserCtx {})
+    }
+}
+fn parse_file(ctx: &ParserCtxRc) -> Parser<u8, ParsedFile> {
+    (parse_our_id() + parse_peer_list(ctx) + parse_graph() + parse_meta_election(ctx) - parse_end())
+        .map(|(((our_id, peer_list), graph), meta_election)| ParsedFile {
             our_id,
             peer_list,
             graph,
             meta_election,
-        },
-    )
+        })
 }
 
 fn parse_peer_id() -> Parser<u8, PeerId> {
@@ -101,7 +110,7 @@ fn parse_our_id() -> Parser<u8, PeerId> {
 #[derive(Debug)]
 struct ParsedPeerList(BTreeMap<PeerId, PeerState>);
 
-fn parse_peer_list() -> Parser<u8, ParsedPeerList> {
+fn parse_peer_list(_ctx: &ParserCtxRc) -> Parser<u8, ParsedPeerList> {
     let list_defs =
         comment_prefix() * seq(b"peer_list: {") * next_line() * parse_peer().repeat(0..)
             - comment_prefix()
@@ -346,7 +355,7 @@ struct ParsedMetaEvent {
     meta_votes: BTreeMap<PeerId, Vec<MetaVote>>,
 }
 
-fn parse_meta_election() -> Parser<u8, ParsedMetaElection> {
+fn parse_meta_election(ctx: &ParserCtxRc) -> Parser<u8, ParsedMetaElection> {
     seq(b"/// ===== meta-elections =====")
         * next_line()
         * (parse_consensus_history() - next_line()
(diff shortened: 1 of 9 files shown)


src/dev_utils/dot_parser.rs
```diff
@@ -75,61 +77,107 @@ struct ParsedFile {
...
+        - parse_end(ctx))
+    .map(|(((our_id, peer_list), graph), meta_election)| ParsedFile {
+        our_id,
+        peer_list,
+        graph,
+        meta_election,
+    })
+}
+
+fn parse_peer_id(_ctx: &ParserCtxRc) -> Parser<u8, PeerId> {
```
maqi: seems `_ctx` doesn't need to be passed down?

maqi: seems `_ctx` doesn't need to be passed down?

src/dev_utils/dot_parser.rs
```diff
@@ -75,61 +77,107 @@ struct ParsedFile {
...
         .repeat(1..)
         .collect()
         .convert(String::from_utf8)
         .map(|s| PeerId::new(&s))
 }
 
-fn parse_our_id() -> Parser<u8, PeerId> {
-    comment_prefix() * seq(b"our_id: ") * parse_peer_id() - next_line()
+fn parse_our_id(ctx: &ParserCtxRc) -> Parser<u8, PeerId> {
+    comment_prefix() * seq(b"our_id: ") * parse_peer_id(ctx) - next_line()
```
maqi: doesn't need to pass `ctx` down?

src/dev_utils/dot_parser.rs
```diff
@@ -75,61 +77,107 @@ struct ParsedFile {
...
+    let state = seq(b"PeerState(") * list(parse_single_state(ctx), sym(b'|')) - sym(b')');
     state.map(|states| {
         states
             .into_iter()
             .fold(PeerState::inactive(), |s1, s2| s1 | s2)
     })
 }
 
-fn parse_single_state() -> Parser<u8, PeerState> {
+fn parse_single_state(_ctx: &ParserCtxRc) -> Parser<u8, PeerState> {
```
maqi: seems `_ctx` doesn't need to be passed down?

src/dev_utils/dot_parser.rs
```diff
@@ -75,61 +77,107 @@ struct ParsedFile {
...
 
-fn parse_peer() -> Parser<u8, (PeerId, PeerState)> {
-    comment_prefix() * parse_peer_id() - seq(b": ") + parse_peer_state() - next_line()
+fn parse_peer(ctx: &ParserCtxRc) -> Parser<u8, (PeerId, PeerState)> {
+    comment_prefix() * parse_peer_id(ctx) - seq(b": ") + parse_peer_state(ctx) - next_line()
 }
 
-fn parse_peer_state() -> Parser<u8, PeerState> {
-    let state = seq(b"PeerState(") * list(parse_single_state(), sym(b'|')) - sym(b')');
+fn parse_peer_state(ctx: &ParserCtxRc) -> Parser<u8, PeerState> {
```
maqi: doesn't need to pass `ctx` down?

src/dev_utils/dot_parser.rs
```diff
@@ -75,61 +77,107 @@ struct ParsedFile {
...
+    list_defs.map(move |defs| {
+        let peer_list = ParsedPeerList(defs.into_iter().collect());
+        ctx.populate_peer_ids(&peer_list);
+        peer_list
+    })
 }
 
-fn parse_peer() -> Parser<u8, (PeerId, PeerState)> {
-    comment_prefix() * parse_peer_id() - seq(b": ") + parse_peer_state() - next_line()
+fn parse_peer(ctx: &ParserCtxRc) -> Parser<u8, (PeerId, PeerState)> {
```
maqi: doesn't need to pass `ctx` down?

src/dev_utils/dot_parser.rs
jeanphilippeD: The functions that needs it is pretty deep, so we would not save that many context passing.
The reason I passed it to all parse function is that it seemed a lot clearer to have it consistently instead of on some and not others. If they were member functions (as I originally wrote it), they would all have &self (and it would not mention that it is unused).

I had a note in the commit comment:
```
We could make all the parse function member functions, however, in order to
use a field within a closure used for parsing, Rc need to be used, so it seem
cleaner to pass this Rc in all parse functions. (Also it seem more readable
and avoid unecessary changes).

The context is given to all the 'parse_*' functions for consistency, as it they
all were member functions with ctx as self.
```

maqi: hmm, but it seems causing many un-necessary changes to the file.
Also, I think we are supposed to avoid using `_` parameter, unless there is a very strong reason that particular parameter shall be retained for later on usage. However `consistency` is not a strong reason?

src/dev_utils/dot_parser.rs
```diff
@@ -75,15 +77,54 @@ struct ParsedFile {
     meta_election: ParsedMetaElection,
 }
 
-fn parse_file() -> Parser<u8, ParsedFile> {
-    (parse_our_id() + parse_peer_list() + parse_graph() + parse_meta_election() - parse_end()).map(
-        |(((our_id, peer_list), graph), meta_election)| ParsedFile {
+type ParserCtxRc = Rc<ParserCtx>;
+
+struct ParserCtx {
```
maqi: I am not sure, however if there is only one member, normally we prefer `ParserCtx()` instead of `ParserCtx{}`

src/dev_utils/dot_parser.rs
```diff
@@ -75,15 +77,54 @@ struct ParsedFile {
...
+}
+
+impl ParserCtx {
+    fn new_rc() -> Rc<Self> {
+        Rc::new(ParserCtx {
+            peer_ids: RefCell::new(BTreeMap::new()),
+        })
+    }
+
+    fn populate_peer_ids(&self, peer_list: &ParsedPeerList) {
```
maqi: maybe `peer_ids_from_peer_list` to be consistenet with the other function?

src/dev_utils/dot_parser.rs
```diff
@@ -714,7 +768,8 @@ impl ParsedContents {
 
 /// Read a dumped dot file and return with parsed event graph and associated info.
 pub(crate) fn parse_dot_file<P: AsRef<Path>>(full_path: P) -> io::Result<ParsedContents> {
-    let result = unwrap!(read(File::open(full_path)?));
+    let name: Option<String> = full_path.as_ref().to_str().map(|s| s.to_string());
+    let result = unwrap!(read(File::open(full_path)?), "read: {:?}", name);
```
maqi: this `name` is only for the unwrap print out? maybe it's un-necessary?

src/dev_utils/record.rs
```diff
@@ -241,19 +241,36 @@ fn extract_genesis_group<'a>(
...
     use super::*;
-    use crate::parsec::assert_same_events;
-
-    #[test]
-    fn smoke() {
-        let path = "input_graphs/benches/minimal.dot";
+    use crate::parsec::get_graph_snapshot;
 
+    /// Run smoke test using given dot file
+    /// ignore_last_event = true if replay needed to generate a fake request to pass the remaining gossip events
```
maqi: there is only `ignore_last_events`, and it is usize ?

src/dev_utils/record.rs
```diff
@@ -241,19 +241,36 @@ fn extract_genesis_group<'a>(
...
+    fn smoke(path: &str, ignore_last_events: usize) {
         let contents = unwrap!(parse_dot_file(path));
         let expected = Parsec::from_parsed_contents(contents);
+        let expected_events = get_graph_snapshot(&expected, 0 /* No events to skip */);
```
maqi: better put comment separately above the line?
same to the other places.

src/gossip/graph/mod.rs
```diff
@@ -222,11 +222,24 @@ pub(crate) mod snapshot {
...
-    pub(crate) struct GraphSnapshot(BTreeSet<EventHash>);
+    pub(crate) struct GraphSnapshot(pub BTreeSet<EventHash>);
 
     impl GraphSnapshot {
         pub fn new<P: PublicId>(graph: &Graph<P>) -> Self {
-            GraphSnapshot(graph.iter().map(|event| *event.hash()).collect())
+            Self::new_with_ignore(graph, 0)
+        }
+
+        pub fn new_with_ignore<P: PublicId>(graph: &Graph<P>, ignore_last_events: usize) -> Self {
```
maqi: better have a comment saying : the number of last events will be ignored?

src/parsec.rs
```diff
@@ -2373,13 +2375,11 @@ impl<T: NetworkEvent, S: SecretId> DerefMut for TestParsec<T, S> {
     }
 }
 
-/// Assert that the two parsec instances have the same events modulo their insertion order.
+/// Get the parsec instances inserted events out of order.
```
maqi: hmmm....    `Get the parsec instances with inserted events out of order.` ?

src/dev_utils/dot_parser.rs
jeanphilippeD: This is a generic structure meant to pass anything (i.e it stand in for &self). With `ParserCtx()`  the member would not have a name so it would not be clear what it is for. I'd rather avoid that, and quite keen to have to rename ParserCtx to `ParserCtxPeerIds` which i think would not reflect the design intent.

src/dev_utils/dot_parser.rs
jeanphilippeD: one is an accessor, this one modify the list.

it could be called `set_peer_ids`, but I would think a `set` function would take the `BTreeMap<String, PeerId>` as argument, that is why I prefixed it with populate.

src/dev_utils/dot_parser.rs
jeanphilippeD: I added it because it made it so much simpler to read the error message. This is an error message that come up every time there is a parsing error, so pretty much every time work is done on the dot_parser.  The next step is to open this file to look where this error occurs.

src/dev_utils/record.rs
jeanphilippeD: Good catch, need to update that comment

src/parsec.rs
jeanphilippeD: Thanks, yes that comment update wasn't well thought trough. will use your version.

src/dev_utils/dot_parser.rs
maqi: if so, could you make the error message more explicit, say: failed to read {:?} ?

src/dev_utils/record.rs
jeanphilippeD: Used a named constant (same for the others).

src/parsec.rs
jeanphilippeD: No, that is not right, we are not getting a Parsec instances we are getting events. Will try to reformulate.

benches/bench.rs
```diff
@@ -69,7 +69,7 @@ use std::cmp;
 
 #[cfg(feature = "testing")]
 fn bench(c: &mut Criterion) {
-    for name in &["minimal", "static", "dynamic"] {
+    for name in &["minimal", "minimal_annie", "static", "dynamic"] {
```
madadam: I might be missing something, but why are you adding this to benchmarks? It looks to me the purpose is to test that non-standard peer names don't break the dot parser. Why do we also need to benchmark it?

jeanphilippeD: A couple of reasons, but I hesitated a lot for that.

The main one, most sensible one, is that I wanted to ensure the benchmark are working, in particular, they compare the consensus history at the end, and I don't think the smoke test does 

Also, I felt it made sense to make the 2(and then 3) smoke test use files from the same directory, and it would have felt weird to not have it in benchmark.

madadam: I feel this makes things confusing. We are overloading the meaning of benchmarks. I'd maybe expand the smoke test to check the consensus history too, or write a separate tests for that instead. The last point feels weak. I'm OK with reusing dot files from other tests/benchmarks, but putting a unrelated dot file among the existing benchmarks just to have them all in the same directory causes more harm than good in my opinion.

jeanphilippeD: Good point, I'll try to update the smoke test.

jeanphilippeD:
The last 2 fixup commits address Adams comment.

src/dev_utils/record.rs
```diff
@@ -241,19 +241,88 @@ fn extract_genesis_group<'a>(
...
+        fn new_with_one_trucated_element(hashes: &'th [Hash]) -> Self {
+            Self {
+                actual_len: hashes.len(),
+                hashes: &hashes[0..cmp::max(hashes.len(), 1) - 1],
+            }
+        }
+    }
+
+    /// Run smoke test using given dot file
+    /// Use ignore_last_event to skip fake request that replay generates to send remaining gossip events.
```
maqi: `ignore_last_event` or `ignore_last_events` ?

src/dev_utils/dot_parser.rs
```diff
@@ -75,15 +76,23 @@ struct ParsedFile {
     meta_election: ParsedMetaElection,
 }
 
-fn parse_file() -> Parser<u8, ParsedFile> {
-    (parse_our_id() + parse_peer_list() + parse_graph() + parse_meta_election() - parse_end()).map(
-        |(((our_id, peer_list), graph), meta_election)| ParsedFile {
+type ParserCtxRc = Rc<ParserCtx>;
```
pierrechevalier83: I don't really see the value in this type alias. Given it doesn't shorten the type much or provide an abstraction that's particularly easy to reason about, I see it as just adding an unnecessary level of indirection.

src/dev_utils/dot_parser.rs
```diff
@@ -101,7 +110,7 @@ fn parse_our_id() -> Parser<u8, PeerId> {
 #[derive(Debug)]
 struct ParsedPeerList(BTreeMap<PeerId, PeerState>);
 
-fn parse_peer_list() -> Parser<u8, ParsedPeerList> {
+fn parse_peer_list(_ctx: &ParserCtxRc) -> Parser<u8, ParsedPeerList> {
```
pierrechevalier83: Nitpick: (Feel free to ignore but maybe keep in mind for next time)
Just a comment in general: a commit should do one thing. I think this commit does less than one thing: without reading further commits, it is not obvious that this should be done. I would probably have squashed this with a subsequent commit to have a commit that presents one conceptual unit.

src/dev_utils/dot_parser.rs
```diff
@@ -78,13 +79,44 @@ struct ParsedFile {
...
+
+        let peer_ids = self.peer_ids.borrow();
+        let search_short_name = short_name.to_string();
+
+        // The full name will be the first greater or equal full name
+        let mut range = peer_ids.range((Included(search_short_name), Unbounded));
+        if let Some((_name, peer_id)) = range.next() {
+            return peer_id.clone();
+        }
+
```
pierrechevalier83: I think it would make sense to also panic if there is more than one match for that initial. It would make the contract more explicit/catch possible misuses.

src/mock.rs
```diff
@@ -66,18 +66,39 @@ impl PeerId {
         }
     }
 
+    #[cfg(not(feature = "mock"))]
+    fn new_with_keypair(id: &str) -> Self {
+        Self::new_with_random_keypair(id)
+    }
+
+    #[cfg(feature = "mock")]
```
pierrechevalier83: Should it be for the `mock` feature or for the `testing` feature? Does this mean that some of the tests won't work correctly with real crypto?

jeanphilippeD:
squashed.

src/mock.rs
```diff
@@ -66,18 +66,39 @@ impl PeerId {
...
+    fn new_with_deterministic_keypair(id: &str) -> Self {
+        use crate::hash::Hash;
+        use rand::{SeedableRng, XorShiftRng};
+        use safe_crypto::init_with_rng;
+
+        let name_hash = Hash::from(id.as_bytes());
+        let seed = name_hash
+            .as_bytes()
+            .chunks_exact(4)
+            .map(|bytes| u32::from_le_bytes([bytes[0], bytes[1], bytes[2], bytes[3]]))
```
pierrechevalier83: since this 0, 1, 2, 3, pattern is done twice in the same function; maybe it's worth extracting into a
`to_array_of_len_4` function. I don't think there is a really elegant way to do this but at least you can de-duplicate it and separate it from the actual app logic to avoid distracting the reviewer ;)

src/dev_utils/record.rs
```diff
@@ -241,19 +241,88 @@ fn extract_genesis_group<'a>(
...
+    #[test]
+    fn smoke_parsec() {
+        let ignore_last_events = 0;
+        smoke("input_graphs/benches/minimal.dot", ignore_last_events)
+    }
 
-        assert_same_events(&actual, &expected);
+    #[test]
+    fn smoke_other_peer_names() {
+        let ignore_last_events = 1;
```
pierrechevalier83: I didn't get your commit message explanation for this. Can you explain again?

src/dev_utils/record.rs
```diff
@@ -241,19 +241,88 @@ fn extract_genesis_group<'a>(
...
+    #[derive(PartialEq, Eq, Debug)]
+    struct TrucatedHashes<'th> {
+        actual_len: usize,
+        hashes: &'th [Hash],
+    }
+
+    impl<'th> TrucatedHashes<'th> {
+        fn new_with_one_missing_element(hashes: &'th [Hash]) -> Self {
+            Self {
+                actual_len: hashes.len() + 1,
```
pierrechevalier83: seems like an off-by-one error to me... Are you keeping the one-based length for some strange reason? (same at line 263)

src/dev_utils/dot_parser.rs
jeanphilippeD: yes, this commit was a lot bigger originally. maybe I'll squash it now if it does not add too much noise to the following commit.

src/dev_utils/dot_parser.rs
jeanphilippeD: This is the same contract that existed before, I do not see a reason to change it particularly considering that it is changed in the next PR, where this new contract would be incorrect:

The short name 'Al' Will match both the name 'Al' and 'Alice' and should be understood to be for 'Al', the first greater or equal.

src/dev_utils/record.rs
jeanphilippeD: It is of sort of: The .dot file is missing the last consensus event but for a mostly sensible reason (though we may change it in the future).

We generate the dumps when we reach consensus, that is before the new block is added.
As a result in the dump, the last block is missing because we are still in the middle of handling the event, but when running parsec, the event handling complete, the block is added to the history.

When initially adding the test to the benchmark we decided to avoid changing the .dot files, and it would not have been easy to do. So now, I'm importing this test as is from bench.rs (though I refactored to make it clearer).

src/dev_utils/record.rs
jeanphilippeD: I'll update the PR comment.
If the last consensus is reached as we process an event that is not ours, then the dup-graph does not contain the last request we receive (just the event needed to reach the consensus). To replay we need to send these events to parsec so it can process them and reach the same consensus. It was decided that the most pragmatic approach at the time was to create a fake request will all the remaining gossip and send it during replay.

As a result, when replay complete, the parsec graph contain this fake request, however, the parsec from the parsed .dot file does not contain it. So we need to ignore it.

src/mock.rs
jeanphilippeD: Because the required  key generation interface is only available if mock feature is present for safe_crypto, it seems reasonable to check for mock.

Because testing imply mock, no test will have problems.

src/mock.rs
jeanphilippeD: We need to create exactly an array of 4 type.
I had trouble when I was trying to do that originally, especially since it needs to be a generic type as it apply to u8 in the first case and u32 in the second.

Do you have a good idea on how to do that in a clean way?

src/dev_utils/record.rs
jeanphilippeD: updated PR comment

src/dev_utils/dot_parser.rs
pierrechevalier83: I expected something like this but didn't find it in this PR, hence the comment. For next time, I would definitely make this entire thing a single PR so the motivation for each change can be obvious to the reviewer. Happy for this not too change though given this explanation.

src/mock.rs
pierrechevalier83: not very clean but you could take an iterator over a generic type and do this `[x[0], x[1], x[2], x[3]]` pattern. At least, that would be a very clear function even though not pretty impl. A bit of googling didn't convince me there was a clean way to do it.

PR #246, opened 2019-02-10
Benchmark with 48 peers: robust short peer name

jeanphilippeD:
The first commit handle gracefully peers with same initial: Compute how long the short name needs to be to avoid any collision with given peers and use it for all peers. (This scheme does not modify existing test cases that only use peer with names from NAMES.)

The second commit extend our benchmark generating code to handle any number of peers. Add a 48 peers benchmark for 8 events only as 16 is still a bit slow.

### DIFF
--- a/src/dev_utils/dot_parser.rs
+++ b/src/dev_utils/dot_parser.rs
@@ -96,14 +96,13 @@ impl ParserCtx {
             .collect();
     }
 
-    fn peer_id_from_short_name(&self, short_name: char) -> PeerId {
+    fn peer_id_from_short_name(&self, short_name: &str) -> PeerId {
         use std::ops::Bound::{Included, Unbounded};
 
         let peer_ids = self.peer_ids.borrow();
-        let search_short_name = short_name.to_string();
 
         // The full name will be the first greater or equal full name
-        let mut range = peer_ids.range((Included(search_short_name), Unbounded));
+        let mut range = peer_ids.range::<str, _>((Included(short_name), Unbounded));
         if let Some((_name, peer_id)) = range.next() {
             return peer_id.clone();
         }
@@ -630,7 +629,8 @@ fn parse_meta_votes(ctx: &Rc<ParserCtx>) -> Parser<u8, BTreeMap<PeerId, Vec<Meta
 }
 
 fn parse_peer_meta_votes(ctx: &Rc<ParserCtx>) -> Parser<u8, (PeerId, Vec<MetaVote>)> {
-    let peer_line = comment_prefix() * is_a(alphanum).map(char::from) - seq(b": ")
+    let peer_line = comment_prefix() * is_a(alphanum).repeat(1..).convert(String::from_utf8)
+        - seq(b": ")
         + parse_meta_vote()
         - next_line();
     let next_line = comment_prefix() * parse_meta_vote() - next_line();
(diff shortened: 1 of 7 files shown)


src/dev_utils/schedule.rs
```diff
@@ -296,7 +296,11 @@ pub struct ObservationSchedule {
 impl ObservationSchedule {
     fn gen<R: Rng>(rng: &mut R, options: &ScheduleOptions) -> ObservationSchedule {
         let mut schedule = vec![];
-        let mut names_iter = NAMES.iter();
+        let mut names_iter = NAMES
+            .iter()
+            .map(|name| name.to_string())
+            // Generate numbered names skipping the ones in NAMES.
+            .chain((10..).map(|num| num.to_string()));
```
octol: Does this mean we can just remove the numbers in NAMES?

jeanphilippeD: We can do that, but that would require regenerating the .dot files that were using a number (a few of the benchmarks).
I think it would add noise to this PR, so I think it would be better as a different PR, I can do that if you want?

octol: It's a pretty minor thing so I'll leave that to your discretion :)
But if you do it, yes then a separate PR is probably better

jeanphilippeD:
Sorry, my local branch had progressed with some work for another PR, and pushed it by accident.
pushed force to restore to the original state that was reviewed (i.e remove spurious commits).

src/dev_utils/dot_parser.rs
```diff
@@ -98,14 +98,13 @@ impl ParserCtx {
...
-    fn peer_id_from_short_name(&self, short_name: char) -> PeerId {
+    fn peer_id_from_short_name(&self, short_name: &str) -> PeerId {
         use std::ops::Bound::{Included, Unbounded};
 
         let peer_ids = self.peer_ids.borrow();
-        let search_short_name = short_name.to_string();
 
         // The full name will be the first greater or equal full name
-        let mut range = peer_ids.range((Included(search_short_name), Unbounded));
+        let mut range = peer_ids.range((Included(short_name.to_string()), Unbounded));
```
madadam: You don't need `to_string` here, but it requires explicit type annotation for some reason:
```rust
peer_ids.range::<str, _>((Included(short_name), Unbounded))
```

jeanphilippeD: Nice one. fixed

jeanphilippeD:
Rebased on top of master that now includes dependent branch.

src/dump_graph.rs
```diff
@@ -918,7 +918,8 @@ mod detail {
             .map(|adjacent: &[&str]| find_mismatch(adjacent[0].as_bytes(), adjacent[1].as_bytes()))
             .max()
             .map(|len| len + 1)
-            .unwrap_or(1 /* 1 char enough if only one element */);
+            // 1 char enough if only one element
+            .unwrap_or(1);
```
octol: Quick question: does this prevent `rustfmt` from autoformatting?

jeanphilippeD: No what do you mean?
The reason I changed this comment is because Qi pointed out we don't do this kind of comments, and there is no other in the code. Do you see a formatting issue?

octol: There is no formatting issue, it looks fine. I might be mistaken but I think `rustfmt` will not try to reformat that whole chain of function calls because of the comment in the middle. Was thinking about the case if let say that code needs reformatting because surrounding code changes (indentation for example)?

jeanphilippeD:
Rebased ontop of master, regenerate the files and merge conflicts with PR #235

src/dump_graph.rs
octol: For the above reason I prefer to not have comments inside expressions, but it's a minor thing so I'm fine with being overruled ;)

jeanphilippeD: No, I just misunderstood that, indeed it does not do well. I'll fix it.

jeanphilippeD: BTW it was not an issue with the `/* */` version.

PR #247, opened 2019-02-11
Add routing generated benchmarks

jeanphilippeD:
The first commit move dump-graph code to dump_graph.rs so it can be modified cleanly in the next one.

The second commit handle gracefully peer ids with special character, and opaque payloads with special character and/or too long like they are when using dump-graphs from within routing tests.

The third commit add the generated tests and use the same consensus mode as routing to replay them correctly.

Benchmark results:
PublicIdname754598-054  57.224 ms
PublicIdname754598-110  93.113 ms
PublicIdname754598-111  4.9948 ms
PublicIdname93b63e-054  63.197 ms
PublicIdname93b63e-113  93.925 ms
PublicIdname93b63e-194  4.8683 ms


src/dump_graph.rs
```diff
@@ -895,6 +915,52 @@ mod detail {
             .collect()
     }
 
+    type DotObservationStore = BTreeMap<ObservationKey, DotObservation>;
```
madadam: Is the purpose of this type just to have pretty debug output of `ObservationStore`? If so, I think a better way would to create a borowing wrapper over `ObservationStore` (or just `Observation`) and just implement `Debug` for it. Something like:

```rust
struct ObservationDebug<'a>(&'a Observation);

impl<'a> Debug for ObservationDebug<'a> {
    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
        match self.0 {
            // ...
            Observation::OpaquePayload(payload) => {
                // ...
            }
        }
    }
}
```

This would avoid unnecessary cloning and allocations. A precedent for this pattern is the `display` method on `std::path::Path`.

src/dump_graph.rs
```diff
@@ -895,6 +915,52 @@ mod detail {
...
+                Observation::Remove { peer_id, .. } => format!("Remove({:?})", peer_id),
+                Observation::Accusation { offender, malice } => {
+                    format!("Accusation {{ {:?}, {:?} }}", offender, malice)
+                }
+                Observation::OpaquePayload(payload) => {
+                    let max_length = 16;
+                    let mut payload_str = format!("{:?}", payload);
+                    if payload_str.len() > max_length {
+                        payload_str.truncate(max_length - 2);
+                        payload_str.push('.');
```
madadam: you can use `push_str("..")` instead of two `push`es.

src/dump_graph.rs
```diff
@@ -961,22 +992,29 @@ mod detail {
...
     }
 
-    fn short_peer_id_names(peer_ids: &PeerIndexMap<String>) -> PeerIndexMap<String> {
+    fn sanitize_peer_id<P: PublicId>(peer_id: &P) -> DotPeerId {
+        let value = sanitize_string(format!("{:?}", peer_id));
+        DotPeerId { value }
+    }
+
+    fn sanitize_string(mut value: String) -> String {
+        value.retain(|c| c.is_ascii() && c.is_alphanumeric());
```
madadam: Consider replacing non-ascii chars with a placeholder, say `?`.

src/dump_graph.rs
jeanphilippeD: I'm not sure it would be a performance win since we end up doing the formatting job again and again as we are displaying the observation again and again.

One of the benefit of storing the String directly is no template parameters, so it cleans up the code. Also it is not performance sensitive code since it is only in dump-graph code.

The reason we have Debug, is that debug was used before, and in particular we are writing out collection of stuff using the debug output (something that we may want to clean in the future).
Otherwise I could just use string, though there is a type safety benefit in using a wrapping object.

src/dump_graph.rs
jeanphilippeD: That would require the parser to handle non alpha numeric char, which I do not think it is worth it at this point.

src/dump_graph.rs
madadam: Although the performance argument might not be important here, I think what I'm suggesting would result in simpler code and fewer modification to the existing codebase (only the part where we actually print the observation would have to change). This isn't critical though, so feel free to leave it as is.

jeanphilippeD: Thanks, I think I will keep it then.

I started with something that looked a lot like what you proposed because it was closer to the original code, but as I was developing this version made it much easier to pass in additional arguments like the peer list of the observation key (that we still need here).

So while this flexibility is not as needed now because I'm not doing everything I originally did, I think this will be easier to maintain and adjust to our needs.

src/dump_graph.rs
jeanphilippeD: This was a copy paste from the original code.
I then modify it in latter commits and not use .. anymore since it is not alphanumeric.

jeanphilippeD:
rebased ontop of current state of #246.
Add a fixup comment with the change needed to the smoke_routing test.
Note: cannot run the test for consensus history on the new test case, because as for routing dump graph, the sanitized public ID map to a different Public/Private key than the original ones.

jeanphilippeD:
Rebased on top of master that in particular contains PR #235.
Resolve conflicts and regenerate .dot files.
Squashed the fixups.

PR #251, opened 2019-02-15
Remove unecessary benchmarks, add missing 48 peers and clean up

jeanphilippeD:
Benchmarks now consistent between different peers: Remove Bob and Carol.
Benchmarks now have 48 peers: Remove peer number 12 and 24.
Auto generated code is not example, but maintained code: move to sub function.

This will also support future benchmarking work like adding ConsensusMode::Single or composite test with many more events, but also simplify any regeneration of dot files when our code change.


jeanphilippeD:
Fixed the commit comment.

PR #252, opened 2019-02-15
Compute consensus of multiple payloads at the same time

madadam:
This PR modified Parsec to compute multiple stable blocks at the same time, if there are multiple interesting payloads in a single meta-election. This results is 30 - 40% speed boost (depending on benchmark).


jeanphilippeD:
For reference: Ran on the new test (not published, so just a FYI) with 1024 events with about 8 opaque events for each gossip event.

What is good is that the consensus history now match, and it definitely improve the performance more with higher section size (better scalability).

```
            ScheduleOptions {
                genesis_size: *genesis_size,
                opaque_to_add,
                // Once every 50 steps in one peer in the network
                prob_gossip: 0.02 / *genesis_size as f64,
                // 4 opaque event for each gossip event over the network
                prob_opaque: 0.02 * 8.0,
                // Events will be seen within 2 gossip events
                max_observation_delay: 100,
                ..Default::default()
            }

jp@jp-maidsafe:~/Git/maidsafe/hgproj/parsec$ time RUSTFLAGS=-g cargo bench --features=testing -- --baseline=master-69dbaf 1024
bench_section_size_evt1024_interleave::a_node4_opaque_evt1024                                                                          
                        time:   [263.67 ms 270.95 ms 276.08 ms]
                        change: [-23.064% -21.954% -20.517%] (p = 0.00 < 0.05)
                        Performance has improved.

bench_section_size_evt1024_interleave::a_node8_opaque_evt1024                                                                        
                        time:   [929.99 ms 934.75 ms 937.33 ms]
                        change: [-46.566% -46.318% -46.047%] (p = 0.02 < 0.05)
                        Performance has improved.

bench_section_size_evt1024_interleave::a_node16_opaque_evt1024                                                                        
                        time:   [2.7814 s 2.8019 s 2.8330 s]
                        change: [-56.230% -55.881% -55.411%] (p = 0.02 < 0.05)
                        Performance has improved.

bench_section_size_evt1024_interleave::a_node32_opaque_evt1024                                                                        
                        time:   [10.833 s 10.847 s 10.849 s]
                        change: [-70.947% -70.579% -70.100%] (p = 0.02 < 0.05)
                        Performance has improved.
```

src/meta_voting/meta_election.rs
```diff
@@ -220,15 +232,16 @@ impl MetaElection {
...
     ) {
         let remove = self
             .unconsensused_events
             .iter()
             .filter(|event_index| {
                 graph
                     .get(**event_index)
-                    .map(|event| event.payload_key() == Some(decided_key))
+                    .and_then(|event| event.inner().payload_key())
+                    .map(|payload_key| decided_keys.contains(payload_key))
```
jeanphilippeD: Can use `any` here.

I'm a bit wary of the complexity here (going through all the unconsensused events and for each doing a search on all the decided keys (Which could be many), maybe worth passing HashSet to this function and the other?
(Could be premature but it seems a clearer way of expressing that we are using decided_keys for lookup).

src/observation.rs
```diff
@@ -271,6 +271,23 @@ impl ObservationKey {
             ObservationKey::Supermajority(..) => ConsensusMode::Supermajority,
         }
     }
+
+    /// Compare `ObservationKey`s to achieve ordering that is consistent among different nodes.
+    pub fn consistent_cmp<S: SecretId>(&self, other: &Self, peer_list: &PeerList<S>) -> Ordering {
+        if let (
```
jeanphilippeD: Does this assume that either they are both single, or hashes are different?
Would this be simpler and provide a total order?
```
fn peer_index() -> Option<PeerIndex>;

pub fn consistent_cmp<S: SecretId>(&self, other: &Self, peer_list: &PeerList<S>) -> Ordering {
    self.hash().cmp(other.hash()).then_with(|| {
        let lhs_peer_id = self.peer_index().map(|peer| peer.id());
        let rhs_peer_id = other.peer_index().map(|peer| peer.id());
        lhs_peer_id.cmp(&rhs_peer_id)
    }
}
```

src/parsec.rs
```diff
@@ -662,19 +669,17 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
             }
```
jeanphilippeD: We return early here, but some create_block calls may have succeeded. That does not seem great.
This would not be an issue if we created a single block with multiple payloads.
(This could also be addressed by changing it to `create_blocks` and not looping on payload_keys.)

src/block.rs
```diff
@@ -19,11 +19,12 @@ use std::collections::{BTreeMap, BTreeSet};
 pub struct Block<T: NetworkEvent, P: PublicId> {
     payload: Observation<T, P>,
     proofs: BTreeSet<Proof<P>>,
+    election: u64,
```
jeanphilippeD: So this is a bit strange for me.
It seems it would be more natural and efficient to have `payload` be `payloads` a vector of Observation.

input_graphs/bench_section_size_evt16/a_node16_opaque_evt16.dot
```diff
@@ -7161,20 +7161,20 @@ digraph GossipGraph {
 /// consensus_history:
```
jeanphilippeD: I would be clearer in the PR comment about the performance benefit. (Also maybe comment on the improvement for routing tests)

src/block.rs
madadam: Agree, but the discussion about your proposed alternative is ongoing. When resolved, I'll implement the change.

src/meta_voting/meta_election.rs
madadam: > Can use `any` here.

`any` isn't defined on `Option`. Or I'm missing what you mean.

Re: complexity: failr point, but `decided_keys` would typically be small and the overhead of `HashSet` might be greater than the gain from the `O(1)` lookup. I'd have to be convinced by a profiler before changing it. Valid point about the intent though.

src/observation.rs
madadam: :+1: Yeah, seems clearer.

src/parsec.rs
madadam: Agree. Will address this when the discussion about multiple payloads per block is resolved. That said, the error from `create_block` is pretty much impossible, because it can only happen when the votes passed to it have different payloads, which cannot happen. Would be great to have a way to express that statically, but I'm not sure how and whether it's even worth spending time on.

src/dev_utils/network.rs
```diff
@@ -414,25 +414,32 @@ impl Network {
 
     /// Checks if the blocks are only signed by valid voters.
     fn check_blocks_signatories(&self) -> Result<(), ConsensusError> {
-        let blocks = unwrap!(self.active_peers().next()).blocks();
+        let block_groups = unwrap!(self.active_peers().next()).grouped_blocks();
```
jeanphilippeD: I had to do something for:
I think you'll need the same.

```
    fn check_consensus(
        &self,
        expected_peers: &BTreeMap<PeerId, PeerStatus>,
        min_expected_observations: usize,
        max_expected_observations: usize,
        // Check the number of consensused blocks.
        let got = unwrap!(self.active_peers().next())
            .blocks_payloads()
            .iter()
            .map(|observation| observation.payload_len())
            .sum();
```

madadam: Sure, but multiple payloads per observation is not implemented in this PR, so this is outside of its scope?

jeanphilippeD: I got that wrong, blocks are flattened, so this PR does not need to change this.

src/parsec.rs
```diff
@@ -1271,26 +1287,20 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+            .fold(BTreeMap::new(), |mut map, payload_key| {
+                *map.entry(payload_key).or_insert(0) += 1;
+                map
+            });
 
-        // IMPORTANT: We must sort this in consistent order, so when the tie breaking rule kicks in,
-        // the outcome is the same for everyone.
-        payloads.sort_by(|a, b| a.hash().cmp(b.hash()));
+        let mut payloads: Vec<_> = payloads.into_iter().collect();
+        payloads.sort_by(|(lhs_key, lhs_count), (rhs_key, rhs_count)| {
```
jeanphilippeD: could use `sorted_by()` from iter_tools.

src/parsec.rs
jeanphilippeD: I did not see that this was still calling create_block instead of updating it to a create_blocks.
Did you intend to leave it this way.

input_graphs/bench_section_size_evt16/a_node16_opaque_evt16.dot
jeanphilippeD: Also, Are we dumping the same graph multiple times?

src/parsec.rs
jeanphilippeD: Also, Are we dumping the same graph multiple times?

madadam: Good catch. I'm going to rewrite this part to process the consensused payloads in bulk.

src/dev_utils/network.rs
```diff
@@ -414,25 +414,32 @@ impl Network {
...
-                }
-                ParsecObservation::Remove { ref peer_id, .. } => {
-                    self.check_block_signatories(block, &valid_voters)?;
-                    let _ = valid_voters.remove(peer_id);
+                    continue;
                 }
-                _ => {
-                    self.check_block_signatories(block, &valid_voters)?;
+
+                self.check_block_signatories(block, &valid_voters)?;
```
Fraser999: Couldn't this be problematic if we have multiple `Add`s and/or `Remove`s within a single block group?  We used to update the valid voters after each block and on the next iteration check the signatures.

src/parsec.rs
```diff
@@ -1250,17 +1262,19 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
         if decided_meta_votes.clone().count() < self.voter_count() {
-            return None;
+            return Vec::new();
         }
 
         self.compute_payload_for_consensus(decided_meta_votes)
     }
 
-    fn compute_payload_for_consensus<I>(&self, decided_meta_votes: I) -> Option<ObservationKey>
+    fn compute_payload_for_consensus<I>(&self, decided_meta_votes: I) -> Vec<ObservationKey>
```
Fraser999: I wouldn't block approval for this, but if there are other changes going in anyway, this might be better named as `compute_payloads_for_consensus`?

src/dev_utils/network.rs
madadam: One of the main reasons why block grouping was introduced is to deal with this. We now update `valid_voters` only after the whole group, not after each individual block.

src/parsec.rs
```diff
@@ -658,26 +659,19 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
 
+        let blocks = self.create_blocks(&payload_keys)?;
         if !blocks.is_empty() {
-            self.consensused_blocks.push_back(BlockGroup(blocks));
+            self.consensused_blocks.push_back(blocks);
         }
 
+        let peer_list_changes = payload_keys
+            .iter()
+            .filter_map(|payload_key| self.handle_consensus(payload_key))
```
jeanphilippeD: Is there a problem there with handle_consensus if there are multiple Add/Remove?

src/parsec.rs
```diff
@@ -658,26 +659,19 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-                Ok(block) => blocks.push(block),
-                Err(Error::MissingVotes) => (),
-                Err(error) => return Err(error),
-            }
-
-            self.mark_observation_as_consensused(&payload_key);
-            peer_list_changes.extend(self.handle_consensus(payload_key));
-        }
+        self.output_consensus_info(&payload_keys);
+        self.mark_observations_as_consensused(&payload_keys);
```
jeanphilippeD: Why do we mark it before when we use to mark it after?

madadam: The order doesn't matter.

src/parsec.rs
madadam: I don't think so. But could you elaborate what your concern is?

src/parsec.rs
jeanphilippeD: Should this be handled as:
```
let peer_list_changes = payload_keys
            .iter()
            .filter_map(|payload_key| {
                 self.mark_observation_as_consensused(payload_key);
                 self.handle_consensus(payload_key)
             })
```

madadam: We can do that, but there is no functional difference.

src/parsec.rs
jeanphilippeD: I am wondering whether `handle_add_peer` would get confused if multiple peers are added at the same time on what should be the state of the new peer (especially if we are the new peer).

It may be nothing, but I'm unsure that there is no impact.

src/parsec.rs
jeanphilippeD: The order matter in case of error in create_block.
This could well be that we want to mark observation as consensused even if create_blocks error or filter them out, but it may warrant a comment.

src/parsec.rs
madadam: I think it should work as expected. Say we have `Add(A)` and `Add(B)` in a single consensus. Then A is added first (it's state updated to become voter), then B is added second while treating A already as voter. I don't see any issue with it, but maybe I'm missing something?

jeanphilippeD: ok.

src/parsec.rs
madadam: I see what you mean now. I guess I can move `mark_observations_as_consensused` after the blocks creation. That should be cleaner, no? I'll also adress @Fraser999 's nit while I'm at it.

jeanphilippeD: The remaining question, is what happen if some were filtered?
Before they would not be marked as consensused or call to 
handle_consensus.

I guess it may not happen since missing vote is not possible?
If that is the case, then should we not let the error up the stack instead of swallowing it, and no consensusing anything of these blocks?

madadam: If I remember correctly, the situation can happen when we are in `ConsensusMode::Single` and there is a opaque vote by someone who has subsequently been removed. So we would reach consensus on that vote, but there is no one to sign it anymore. We don't treat this situation as error, because it can legally happen. We still want to mark the vote as consensued, so the algorithm is not slowed down by having to constantly reprocess it.

jeanphilippeD: Ok that make sense to me. (maybe a 1/2 line comment in the filter part to say that would make things clearer).

I guess this comes from:
`.filter(|event| voters.contains(event.creator()))`

(Not super clear why we should filter these out though, were still valid voters at the time they voted for the event (maybe) So we may drop valid event on the floor.)

madadam: I guess because we currently can't tell whether someone is not a voter because they haven't been promoted to one yet or because they have already been removed. Something we might want to change at some point, but I'd say it has low priority. Losing votes of people who are already gone is something we can live with for now.

PR #254, opened 2019-02-19
test/parsec: replay final parsec state for benchmark: no fake request

jeanphilippeD:
We now generate the benchmarks dump-graphs when parsec is dropped.
We no longer need to create a fake request to replay all needed events.
We now have all the consensus history.

Ensure that we can regenerate replayable test cases even with bigger
changes to the algorithm.
Simplify generating routing test case as it is now obvious which ones to
pick (add missing 2 test cases).

Performance results:
minimal                0.6421 ms
static                 2.4734 ms
dynamic                1.1264 ms
a_node4_opaque_evt8    1.3140 ms
a_node8_opaque_evt8    7.1394 ms
a_node16_opaque_evt8   45.464 ms
a_node32_opaque_evt8   481.09 ms
a_node48_opaque_evt8   1256.7 ms
a_node4_opaque_evt16   2.6595 ms
a_node8_opaque_evt16   20.478 ms
a_node16_opaque_evt16  121.96 ms
a_node32_opaque_evt16  987.09 ms
a_node48_opaque_evt16  4741.7 ms

PublicIdname754598-001 66.294 ms
PublicIdname754598-002 101.34 ms
PublicIdname754598-003 5.6049 ms
PublicIdname93b63e-001 74.325 ms
PublicIdname93b63e-002 112.70 ms
PublicIdname93b63e-003 21.538 ms
PublicIdname93b63e-004 44.776 ms
PublicIdname93b63e-005 5.5226 ms


src/dev_utils/record.rs
```diff
@@ -330,13 +335,16 @@ mod tests {
 
     #[test]
     fn smoke_consensus_history_parsec() {
-        smoke_consensus_history("input_graphs/benches/minimal.dot")
+        let missing_one_consensus = false;
```
octol: I like that you're giving a name to the bool flag for readability

Might even make it into an enum, but that might be overkill?

src/dump_graph.rs
```diff
@@ -158,7 +190,18 @@ mod detail {
         meta_election: &MetaElection,
         peer_list: &PeerList<S>,
         observations: &ObservationStore<T, S::PublicId>,
+        info: &DumpGraphInfo,
     ) {
+        let need_process = DUMP_MODE.with(|mode| match (info, &*mode.borrow_mut()) {
+            (DumpGraphInfo::OnParsecDrop, DumpGraphMode::OnParsecDrop) => true,
+            (DumpGraphInfo::OnConsensus, DumpGraphMode::OnConsensus) => true,
```
octol: Picky (style): maybe merge these cases into one with | ?

src/dev_utils/record.rs
jeanphilippeD: Maybe would be a bit too much for something that is very local :-)

src/dump_graph.rs
```diff
@@ -124,11 +144,23 @@ mod detail {
...
 
     thread_local!(static DUMP_COUNTS: RefCell<BTreeMap<String, usize>> =
         RefCell::new(BTreeMap::new()));
 
+    /// To control the dump graph behaviour
+    #[derive(Clone)]
+    pub enum DumpGraphMode {
+        /// Only dump on consensus
+        OnConsensus,
+        /// Only dump when parsec is droped
```
Fraser999: Typo: "dropped".  Also, the comments are slightly misleading in that the graphs will be dumped when parsec is dropped if the current thread is panicking, regardless of which mode is chosen.

src/dump_graph.rs
```diff
@@ -22,6 +22,12 @@ pub(crate) fn init() {
     detail::init()
 }
 
+#[derive(Clone)]
+pub enum DumpGraphInfo {
+    OnConsensus,
```
Fraser999: Maybe these could be renamed to better indicate the difference between `DumpGraphMode` and its variants.  I'm thinking of something more like:
```rust
pub enum DumpGraphContext {
    ConsensusReached,
    DroppingParsec,
}
```
or something like that?

PR #255, opened 2019-02-25
fix/parsec: avoid accusing same accomplice twice

maqi:


src/parsec.rs
```diff
@@ -1905,7 +1905,14 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
 
         let starting_index = self.peer_list.accomplice_event_checkpoint_by(creator);
         for (_, malice) in self.detect_accomplice_for_our_accusations(event, starting_index)? {
-            self.accuse(creator, Malice::Accomplice(event_hash, Box::new(malice)));
+            if !self.pending_accusations.iter().any(|(off, mal)| match mal {
+                Malice::Accomplice(_, ori_mal) => {
+                    off == &creator && ori_mal == &Box::new(malice.clone())
```
Fraser999: `off == &creator && **ori_mal == malice,` would avoid cloning.

src/parsec.rs
```diff
@@ -1905,7 +1905,14 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
 
         let starting_index = self.peer_list.accomplice_event_checkpoint_by(creator);
         for (_, malice) in self.detect_accomplice_for_our_accusations(event, starting_index)? {
-            self.accuse(creator, Malice::Accomplice(event_hash, Box::new(malice)));
+            if !self.pending_accusations.iter().any(|(off, mal)| match mal {
```
jeanphilippeD: Do we have a test to update for this? Otherwise should we have an integration or unit test for this?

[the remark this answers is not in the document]
maqi: This was following the code in `detect_accomplice_for_our_accusations`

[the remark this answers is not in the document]
maqi: will have a try

maqi: there is a test `consensus_with_fork` which will use this. However the test is currently `ignored` and it won't `expose` the change directly.

src/parsec.rs
```diff
@@ -1905,7 +1905,14 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+                Malice::Accomplice(_, ori_mal) => {
```
jeanphilippeD: So does this code not need to go now?
Am I missing something?

Fraser999: I think you're right - I missed that.

src/parsec.rs
```diff
@@ -1951,6 +1951,12 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
                     .iter()
                     .any(|(off, mal)| (off, mal) == (offender, &malice))
             })
+            .filter(|(_, malice)| {
+                !self.pending_accusations.iter().any(|(off, mal)| match mal {
+                    Malice::Accomplice(_, ori_mal) => off == &creator && &**ori_mal == malice,
```
jeanphilippeD: Do we prefer comparing object or reference?
This looks a bit strange with `&**`:
`&**ori_mal == malice` ->is it better as  `**ori_mal == *malice`?

`off == &creator` ->is it better as   `*off == creator`?

src/parsec.rs
jeanphilippeD: Since we are adding this code, should we add the test that cover it then?

src/parsec.rs
maqi: me and Fraser both prefer comparing by reference, though it's not a strong preference.

PR #256, opened 2019-02-25
fix/parsec: do not fail immediately on first failed gossip event

madadam:
On handling gossip (`handle_request` and `handle_response`), do not fail the whole handling on the first failed event. Instead, try to process as many events as possible, and only return the error (if any) afterwards. This fixes an issue where if a newly joined node recieves an gossip which contains an event from someone who is no longer in the section, the node would never process the remaining events which might block consensus and break the section.

WIP, because discussion is required.


madadam:
Closing in favor of https://github.com/maidsafe/parsec/pull/257

PR #257, opened 2019-02-26
fix/parsec: create own events only when we become voter

madadam:
This PR fixes the following issue:

1. Node (A) is removed from the section
2. New node (B) joins and recieves a gossip message from someone
3. The events in the gossip message might be ordered so that an event from A is before an event which triggers the consensus to add A (*)
4. But B doesn't yet know A, so it reject that event and return error, never processing the rest of the gossip message
5. This might sometimes prevent consensus to be reached and make the section disfunctional

The fix is to make sure that any event from A always comes after the event which triggers the consensus on add A. This is achieved by postponing inserting our events into the graph until we become voter - that is - we are either genesis node or we have observed consensus on adding ourselves.

(*) This is because the initial event and the event which triggered the consensus are neither ancestor nor descendant of each other, so they can be ordered in arbitrarily.

### DIFF
--- a/src/peer_list/peer.rs
+++ b/src/peer_list/peer.rs
@@ -11,7 +11,11 @@ use crate::gossip::{EventIndex, IndexedEventRef};
 use crate::hash::Hash;
 use crate::id::PublicId;
 use crate::serialise;
-use std::iter::{self, FromIterator};
+use itertools::Itertools;
+use std::{
+    fmt::{self, Debug, Formatter},
+    iter::{self, FromIterator},
+};
 
 #[derive(Debug)]
 pub(crate) struct Peer<P: PublicId> {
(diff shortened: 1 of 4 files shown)


src/peer_list/peer.rs
```diff
@@ -174,3 +177,9 @@ impl Slot {
         iter::once(self.first).chain(self.rest.iter().cloned())
     }
 }
+
+impl Debug for Slot {
```
madadam: Note to reviewers: this change is not strictly necessary for the rest of the PR, but it makes the debug output of `PeerList` more readable which simplified debugging when working on the rest of this PR.

maqi: I am ok with having this part included in this PR.
However, maybe it will be better to have this part in a separate commit?

madadam: :+1: Moved to a separate commit.

PR #259, opened 2019-02-26
MAID-3282: add Requesting sync event variant

Fraser999:
All functional tests have been updated except `accomplice_framing_attempt` which was removed since it tests a scenario which cannot happen any more, and `invalid_gossip_creator` which is currently ignored as the corresponding production code has been removed and is unlikely to be replaced.

The task specifies handling malice, but we have since agreed to deal with that in a separate PR.


src/gossip/cause.rs
```diff
@@ -44,13 +46,30 @@ pub(crate) enum Cause<V, E> {
...
+        packed_cause: Cause<Vote<T, P>, EventHash, P>,
         creator: PeerIndex,
         ctx: &EventContextRef<T, S>,
     ) -> Result<(Self, ObservationForStore<T, P>), Error> {
         let cause = match packed_cause {
+            Cause::Requesting {
+                ref self_parent,
+                ref recipient,
+            } => {
+                let recipient_index = if let Some(index) = ctx.peer_list.get_index(recipient) {
```
jeanphilippeD: Maybe create a new function like to avoid adding implementation detail to this function.

```
fn self_parent_index<P: PublicId>(graph: &Graph<P>, hash: &EventHash) -> Result<EventIndex, Error> {
    graph.get_index(hash).ok_or_else(|| {
        debug!("unknown self-parent with hash {:?}", hash);
        Error::UnknownSelfParent
    })
}
```

src/gossip/cause.rs
```diff
@@ -94,8 +113,15 @@ impl<P: PublicId> Cause<VoteKey<P>, EventIndex> {
...
         ctx: EventContextRef<T, S>,
-    ) -> Result<Cause<Vote<T, P>, EventHash>, Error> {
+    ) -> Result<Cause<Vote<T, P>, EventHash, P>, Error> {
         let cause = match *self {
+            Cause::Requesting {
+                self_parent,
+                recipient,
+            } => Cause::Requesting {
+                self_parent: self_parent_hash(ctx.graph, self_parent)?,
+                recipient: ctx.peer_list.get_known(recipient)?.id().clone(),
```
jeanphilippeD: Same, maybe add a named function like `self_parent_hash` for this

src/gossip/cause.rs
```diff
@@ -167,7 +198,10 @@ impl Cause<VoteKey<PeerId>, EventIndex> {
         let other_parent = other_parent.unwrap_or(EventIndex::PHONY);
 
         match packed_cause {
-            Cause::Initial => Cause::Initial,
+            Cause::Requesting { .. } => Cause::Requesting {
+                self_parent,
+                recipient: unwrap!(recipient),
```
jeanphilippeD: should we use unwrap_op?

src/gossip/event.rs
```diff
@@ -69,6 +69,44 @@ pub(crate) struct Event<P: PublicId> {
...
+    ) -> Result<Self, Error> {
+        let content: Content<Vote<T, _>, _, _> = Content {
+            creator: ctx.peer_list.our_pub_id().clone(),
+            cause: Cause::Requesting {
+                self_parent: self_parent_hash(ctx.graph, self_parent)?,
+                recipient: recipient.clone(),
+            },
+        };
+        let (hash, signature) = compute_event_hash_and_signature(&content, ctx.peer_list.our_id());
+        let recipient = if let Some(index) = ctx.peer_list.get_index(recipient) {
```
jeanphilippeD: can re-use proposed sub function here

src/gossip/event.rs
```diff
@@ -69,6 +69,44 @@ pub(crate) struct Event<P: PublicId> {
...
+        self_parent: EventIndex,
+        recipient: &P,
+        forking_peers: &PeerIndexSet,
+        ctx: EventContextRef<T, S>,
+    ) -> Result<Self, Error> {
+        let content: Content<Vote<T, _>, _, _> = Content {
+            creator: ctx.peer_list.our_pub_id().clone(),
+            cause: Cause::Requesting {
+                self_parent: self_parent_hash(ctx.graph, self_parent)?,
+                recipient: recipient.clone(),
```
jeanphilippeD: can reuse proposed sub function here

src/dump_graph.rs
```diff
@@ -669,6 +669,15 @@ mod detail {
         ) -> io::Result<()> {
             let mut buffer;
             let cause = match event.cause() {
+                Cause::Requesting { recipient, .. } => {
+                    let recipient_id = self
```
jeanphilippeD: use peer_ids
```
Cause::Requesting { recipient, .. } => {
   buffer = format!("Requesting({})", self.peer_ids.get(*recipient).unwrap_or(&DotPeerId::unknown());
                     buffer.as_str() 
}

File: src/dev_utils/record.rs [hunk 7]
```diff
@@ -218,6 +229,9 @@ impl Action {
     fn run(self, parsec: &mut Parsec<Transaction, PeerId>) {
         match self {
             Action::Vote(observation) => unwrap!(parsec.vote_for(observation)),
+            Action::Requesting(recipient) => {
+                let _ = unwrap!(parsec.create_gossip(&recipient));
```
maqi: we don't need to concern about any error in case of requesting?

src/gossip/cause.rs
```diff
@@ -29,11 +29,13 @@ use crate::vote::{Vote, VoteKey};
...
 #[serde(bound(
-    serialize = "V: Serialize, E: Serialize",
-    deserialize = "V: Deserialize<'de>, E: Deserialize<'de>"
+    serialize = "V: Serialize, E: Serialize, P: Serialize",
+    deserialize = "V: Deserialize<'de>, E: Deserialize<'de>, P: Deserialize<'de>"
 ))]
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Debug, Serialize, Deserialize)]
-pub(crate) enum Cause<V, E> {
+pub(crate) enum Cause<V, E, P> {
+    // Identifier of the latest `Event` of the peer which sent the request.
```
maqi: this comment is not in the same format as for the `Request` ?

src/parsec.rs
```diff
@@ -280,43 +280,41 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
             .map(|(_, peer)| peer.id())
     }
 
-    /// Creates a new message to be gossipped to a peer, containing all gossip events this peer
-    /// thinks that peer needs. If `peer_id` is `None`, a message containing all known gossip
-    /// events is returned. If `peer_id` is `Some` and the given peer is not an active node, an
-    /// error is returned.
+    /// Creates a new message to be gossiped to a peer, containing all gossip events this peer
```
maqi: gossiped or gossipped ?

src/parsec.rs
```diff
@@ -280,43 +280,41 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-        self.pack_events(self.graph.iter().map(|e| e.inner()))
-            .map(Request::new)
+        let self_parent = self.peer_list.last_event(PeerIndex::OUR).ok_or_else(|| {
+            log_or_panic!("{:?} missing our own last event hash.", self.our_pub_id());
+            Error::Logic
+        })?;
+        let forking_peers = PeerIndexSet::default();
+        let sync_event =
+            Event::new_from_requesting(self_parent, peer_id, &forking_peers, self.event_context())?;
+        let _ = self.add_event(sync_event)?;
```
maqi: we don't need to care about the returned error from add_event in this case?

src/parsec.rs
```diff
@@ -2141,6 +2139,10 @@ impl<T: NetworkEvent, S: SecretId> TestParsec<T, S> {
     #[cfg(feature = "malice-detection")]
     pub fn remove_last_event(&mut self) -> Option<(EventIndex, Event<S::PublicId>)> {
         let (event_index, event) = self.graph.remove_last()?;
+        assert_eq!(
```
maqi: I think we normally avoid having assertion in the production code?

dot_gen/src/main.rs
```diff
@@ -128,64 +128,157 @@ fn main() {
...
+            "functional_tests::handle_malice_genesis_event_not_after_initial",
+            |env| {
+                let obs = ObservationSchedule {
+                    genesis: peer_ids!("Alice", "Bob", "Carol", "Dave"),
+                    schedule: vec![],
+                };
+                Schedule::from_observation_schedule(env, &ScheduleOptions::default(), obs)
+            },
+        )
+        .seed([926181213, 2524489310, 392196615, 406869071]);
```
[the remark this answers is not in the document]
jeanphilippeD: so we can regenerate it. (otherwise each generation is different).

dot_gen/src/main.rs
```diff
@@ -128,64 +128,165 @@ fn main() {
...
 
+    let _ = scenarios
+        .add("functional_tests::from_parsed_contents", |env| {
+            let obs = ObservationSchedule {
+                genesis: peer_ids!("Alice", "Bob", "Carol"),
+                schedule: vec![(1, AddPeer(PeerId::new("Dave")))],
+            };
+            Schedule::from_observation_schedule(env, &ScheduleOptions::default(), obs)
+        })
+        .file("Alice", "0.dot")
```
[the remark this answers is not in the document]
Fraser999: We don't really need a seed for this test.  I'd argue the test is fairly pointless actually and would prefer if it were removed, but that's outwith the scope of this PR I feel.

dot_gen/src/main.rs
```diff
@@ -128,64 +128,165 @@ fn main() {
...
+
+    let _ = scenarios
+        .add("functional_tests::add_peer", |env| {
+            let obs = ObservationSchedule {
+                genesis: peer_ids!("Alice", "Bob", "Carol", "Dave", "Eric"),
+                schedule: vec![(1, AddPeer(PeerId::new("Fred")))],
+            };
+            Schedule::from_observation_schedule(env, &ScheduleOptions::default(), obs)
+        })
+        .seed([411278735, 3293288956, 208850454, 2872654992])
```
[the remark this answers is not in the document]
Fraser999: The seed pre-existed as a comment in the test case.  It would indeed be easier to regenerate all dot files if they all used the same seed, but initially we were aiming to use seeds which resulted in smaller dot files.

dot_gen/src/main.rs
```diff
@@ -128,64 +128,165 @@ fn main() {
...
     let _ = scenarios
         .add("functional_tests::remove_peer", |env| {
             let obs = ObservationSchedule {
                 genesis: peer_ids!("Alice", "Bob", "Carol", "Dave", "Eric"),
                 schedule: vec![(1, RemovePeer(PeerId::new("Eric")))],
             };
-
             Schedule::from_observation_schedule(env, &ScheduleOptions::default(), obs)
         })
         .seed([1048220270, 1673192006, 3171321266, 2580820785]);
```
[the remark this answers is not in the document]
Fraser999: Looks like it just overwrites what it finds there.  I'll add the file for consistency with the other scenarios though.

dot_gen/src/main.rs
```diff
@@ -128,64 +128,157 @@ fn main() {
...
+
+    let _ = scenarios
+        .add("functional_tests::handle_malice::accomplice", |env| {
+            let obs = ObservationSchedule {
+                genesis: peer_ids!("Alice", "Bob", "Carol", "Dave"),
+                schedule: vec![(0, Opaque(Transaction::new("EFGH")))],
+            };
+            Schedule::from_observation_schedule(env, &ScheduleOptions::default(), obs)
+        })
+        .seed([3932887254, 691979681, 2029125979, 3359276664])
```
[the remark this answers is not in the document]
Fraser999: Just that first test which I already comment on.

src/dev_utils/record.rs
```diff
@@ -210,6 +220,7 @@ impl From<ParsedContents> for Record {
 #[derive(Clone)]
 enum Action {
     Vote(Observation<Transaction, PeerId>),
+    Requesting(PeerId),
```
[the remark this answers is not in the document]
Fraser999: I don't think that's required.  These are now fundamental event variants in Parsec, and even if they were to be documented via code comments, I think that'd be more appropriate in `Cause` rather than this test helper struct.

src/dev_utils/record.rs
Fraser999: I don't believe so here - this is a test method, and it's already doing similar for the other `Action` variants in this `match`.

src/functional_tests.rs
```diff
@@ -229,27 +229,14 @@ fn add_peer() {
...
-    let mut fred = TestParsec::from_existing(fred_id, &genesis_group, &genesis_group);
-
-    // Create a "naughty Carol" instance where the graph only shows four peers existing before
-    // adding Fred.
-    #[cfg(feature = "malice-detection")]
-    {
-        parsed_contents = parse_test_dot_file("carol.dot");
-        let naughty_carol = TestParsec::from_parsed_contents(parsed_contents);
-        let malicious_message = unwrap!(naughty_carol.create_gossip(None));
-        assert_err!(
```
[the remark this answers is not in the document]
Fraser999: This is another scenario which can't happen any more, since when Carol creates gossip, her message must have a `Requesting` event and that must specify a recipient peer.  She can't specify Fred since he's not added yet.

src/functional_tests.rs
```diff
@@ -881,23 +868,23 @@ mod handle_malice {
         let mut alice_contents =
             parse_dot_file_with_test_name("alice.dot", "functional_tests_handle_malice_accomplice");
 
-        let a_26_index =
-            unwrap!(find_event_by_short_name(&alice_contents.graph, "A_26")).event_index();
+        let a_18_index =
+            unwrap!(find_event_by_short_name(&alice_contents.graph, "A_18")).event_index();
```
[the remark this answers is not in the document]
Fraser999: I'm not quite clear what you're asking.  In this particular test case, the dot file changed from having Alice's last event as `A_26` to `A_18`, hence the need for this change.

src/gossip/cause.rs
Fraser999: I don't see the difference?

src/gossip/cause.rs
Fraser999: I'd rather not write a separate function to replace something as trivial as this.

jeanphilippeD: It would follow the pattern set by self_parent_hash & co.
But I let you decide.

src/gossip/cause.rs
Fraser999: Not sure what you mean here?

jeanphilippeD: like `let other_parent = other_parent.unwrap_or(EventIndex::PHONY);`

Typo, I meant `unwrap_or`

src/gossip/event.rs
Fraser999: Not sure what you're proposing here.

src/parsec.rs
Fraser999: I agree it looks weird, but it's "gossiped" :)

src/gossip/event.rs
jeanphilippeD: Nothing.
Seem I put comment on the wrong line and then added another one for the correct line bellow.

src/parsec.rs
Fraser999: We are caring - we're using the `?` operator.  It's the successful result (the newly-added event's index) we don't need.

src/parsec.rs
Fraser999: `self` here is `TestParsec`, so it's not production code.

src/gossip/cause.rs
Fraser999: Since it would be adding a few lines of code to only be used once, I'll skip this suggestion thanks :)

src/gossip/cause.rs
Fraser999: I think since this is only for `#[cfg(any(test, feature = "testing"))]` we're better to fail here.

src/gossip/cause.rs
maqi: The comment of `Requesting` says `Identifier of the latest `Event` of the peer which sent the request`, which only explains one, and no explaination to `recipient` ?
meanwhile the comment of `Request` says `Identifiers of the latest `Event`s of own and the peer which sent the request`, which explains both.

Fraser999: Ah, right - I see.  I'll update the comment thanks.

octol:
Sorry I probably missed it, about `accomplice_framing_attempt`, what's the reason it can't happen anymore?

PR #260, opened 2019-02-28
Add 1024 and single consensus mode bench

jeanphilippeD:
Routing is using ConsensusMode::Single: Add initial benchmarks for this mode.
Parsec overhead decrease as the number of event increase. Add initial steady state test with 1024 events (low event rate).

Reduce the sample size to 3 to handle benchmarking the bigger test without taking too long.
Increase the threshold to 10% to avoid noise on small regression/improvements.
Fix inefficiency in memory and computation that were blocking tests with large number of events based on profiling dot_gen for generating graph with 65000 events, and slowing down a lot the 1024 event generation.


dot_gen/src/main.rs
```diff
@@ -287,45 +287,89 @@ fn add_benches(scenarios: &mut Scenarios) {
...
+
+            add_bench_scalability_common(
+                scenarios,
+                single_batch_options.clone(),
+                ConsensusMode::Supermajority,
+                &format!("{}", opaque_to_add),
+            );
+            add_bench_scalability_common(
+                scenarios,
+                single_batch_options.clone(),
```
madadam: You don't need to `clone` here as you are not using the value later.

src/dev_utils/network.rs
```diff
@@ -349,12 +350,18 @@ impl Network {
...
     ) -> Result<(), ConsensusError> {
         // Check the number of consensused blocks.
-        let got = unwrap!(self.active_peers().next()).blocks_payloads().len();
-        if got < min_expected_observations || got > max_expected_observations {
+        let lens: Vec<_> = self
+            .active_peers()
+            .map(|peer| peer.blocks_payloads().len())
+            .collect();
+        let got_min = *unwrap!(lens.iter().min());
+        let got_max = *unwrap!(lens.iter().min());
```
madadam: Shouldn't this be `lens.iter().max()` ?

madadam: Also, `itertools` has `minmax`: https://docs.rs/itertools/0.8.0/itertools/trait.Itertools.html#method.minmax, so you can compute both min and max in the same step and don't need to `collect` to `Vec` first.

madadam: That said, I'm confused by this piece of code. Everybody should have the same number of consensused payloads, so `got_min` and `got_max` will always be equal, unless consensus is broken, which is checked elsewhere.

src/dev_utils/network.rs
```diff
@@ -467,18 +474,24 @@ impl Network {
             min_observations,
             max_observations,
             events,
+            additional_steps,
             options,
         } = schedule;
         let mut queue: VecDeque<_> = events.into_iter().collect();
         let mut retry = Vec::new();
+        let mut additional_steps = additional_steps;
+        let mut additional_step = || additional_steps.next().map(ScheduleEvent::LocalStep);
```
madadam: I see you are doing this a lot - why not put the lambda inline to where it's used?

src/dev_utils/schedule.rs
```diff
@@ -405,13 +399,27 @@ impl ObservationSchedule {
...
+}
+
 /// Stores the list of network events to be simulated.
 #[derive(Clone)]
 pub struct Schedule {
     pub peers: BTreeMap<PeerId, PeerStatus>,
     pub min_observations: usize,
     pub max_observations: usize,
     pub events: Vec<ScheduleEvent>,
+    pub additional_steps: std::ops::Range<usize>,
```
madadam: Why is this `Range` instead of just a number?

src/dev_utils/schedule.rs
```diff
@@ -405,13 +399,27 @@ impl ObservationSchedule {
     }
 }
 
+pub struct StepObservationSchedule {
```
madadam: What is the purpose of this (and the corresponding changes in `Schedule::from_observation_schedule`)? Just refactoring or something deeper?

src/dev_utils/network.rs
jeanphilippeD: This is called at at every point in time so each peer is different.
I looked at min_max, I don't remember why I did not use it, I'll use min_max.

src/dev_utils/network.rs
jeanphilippeD: It was bigger when I originally wrote it.
I also prefer giving names to my closures (like I would with variable and intermediate expressions) except for the simplest of cases.

So in this case I'm quite happy with it having a name, I think it clarify the loop.

src/dev_utils/schedule.rs
jeanphilippeD: The range does not start at 0, it continue with the step number the way we would otherwise have popuated `events`.
It seemed also clearer.

src/dev_utils/network.rs
madadam: I don't entirely agree that it clarifies the loop, but as it's mostly matter of personal preference, I won't block the merge over this.

src/dev_utils/schedule.rs
jeanphilippeD: The repeated call to partition for each step was making the test grind to a halt when the number is big enough => sort first, and process the steps the same we we did before without the complexity.

src/dev_utils/network.rs
madadam: If `got_min` is different from `got_max` then the call to `check_blocks_all_in_sequence` below fails anyway, so I still don't see the point.

madadam: Re: `minmax` - it returns `MinMaxResult` which is a bit awkward to use, but it has `into_option` which you can then just `unwrap`:

```rust
let (min, max) = unwrap!(self.active_peers().map(...).minmax().into_option());
```

src/dev_utils/schedule.rs
madadam: Yes, but the start of the range is fixed (to the number of regular steps), so you can keep only the number of additional steps and then construct the range at the point of use.  It would reduce the number of moving parts and the possibility for things to get out of sync.

jeanphilippeD: In order to do that I would have to look inside the events to find the last LocalStep.
I guess If I were going this way I would use reverse().find(...).
But that seems more complicated.

I am not sure why you think there are additional moving steps. Before Schedule would prepare all the events to be executed, now it does the same, One way I can see that this could be improved Is by replacing the Range with a special iterator type that produce the LocalSteps.

Then we could avoid splitting this between schedule and network.rs and have:
```
while let Some(event) = queue.pop_front().or_else(|| additional_step.next()) {
``

Would that make sense?

File: src/dev_utils/network.rs (hunk 2 above)
Comment (jeanphilippeD) [reply src/dev_utils/network.rs]: > If `got_min` is different from `got_max` then the call to `check_blocks_all_in_sequence`

Because that other check is super slow and cause problem when generating with large number of events. So we want the fast check to be precise.

File: src/dev_utils/schedule.rs (hunk 4 above)
Comment (madadam) [reply src/dev_utils/schedule.rs]: Thinking about this a bit, I now think the range is probably fine, so leave it. Writing a special iterator for this isn't worth it and would only increase complexity.

File: src/dev_utils/network.rs [hunk 6]
```diff
@@ -349,12 +351,21 @@ impl Network {
...
+        let (got_min, got_max) = match self
+            .active_peers()
+            .map(|peer| peer.blocks_payloads().len())
+            .minmax()
+        {
+            MinMaxResult::NoElements => (0, 0),
+            MinMaxResult::OneElement(value) => (value, value),
+            MinMaxResult::MinMax(min, max) => (min, max),
+        };
+        if got_min < min_expected_observations || got_max > max_expected_observations {
```
madadam: >> If `got_min` is different from `got_max` then the call to `check_blocks_all_in_sequence`
>
> Because that other check is super slow and cause problem when generating with large number of events. So we want the fast check to be precise.

So is this faster because the `WrongBlocksNumber` check is now more likely to fail because we are checking all peers instead of just one?

src/dev_utils/network.rs
```diff
@@ -349,12 +351,21 @@ impl Network {
...
-        if got < min_expected_observations || got > max_expected_observations {
+        let (got_min, got_max) = match self
+            .active_peers()
+            .map(|peer| peer.blocks_payloads().len())
+            .minmax()
+        {
+            MinMaxResult::NoElements => (0, 0),
+            MinMaxResult::OneElement(value) => (value, value),
+            MinMaxResult::MinMax(min, max) => (min, max),
+        };
```
madadam: I think `minmax().into_option().unwrap_or((0, 0))` would more succinct an wouldn't require bringing `MinMaxResult` into scope.

jeanphilippeD: I've got a fixup patch for this.

src/dev_utils/network.rs
jeanphilippeD: yes.

madadam: OK, I see. I got it now, but still think it's somehow confusing. I'd maybe try to optimize `check_blocks_all_in_sequence` instead, but I won't block the PR over this anymore.

PR #261, opened 2019-02-28
MAID-3285 fix/parsec: avoid accusing Fork duplicatedly

maqi:
Within the current test of concensus_with_fork, Carol and Bob will accuse Alice of fork at two different events.
We are supposed to accuse fork only at the root event (which means only one event_hash).
Also, we need to consider in the scenario of 
```
A0 --- A1_0 --- A2_0
 \
  \-- A1_1 --- A2_1
           \
             \-- A2_2
```
fork accusations are only casted against A0 and A1_1
and in the scenario of 
```
A0 --- A1_0 --- A2_0
  \
   \-- A1_1 --- A2_1
    \
     \-- A1_2
```
fork accusation is only casted against A0
And this PR is to achieve this.

### DIFF
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -1771,12 +1771,25 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
 
     // Detect whether the event incurs a fork.
     fn detect_fork(&mut self, event: &Event<S::PublicId>) {
-        if self.peer_list.last_event(event.creator()) != event.self_parent() {
+        if self.is_first_fork(event) {
             if let Some(self_parent_hash) = self.graph.self_parent(event).map(|event| *event.hash())
             {
                 self.accuse(event.creator(), Malice::Fork(self_parent_hash));
             }
         }
+    }
+
+    fn is_first_fork(&self, event: &Event<S::PublicId>) -> bool {
+        let same_index_events = self
+            .peer_list
+            .events_by_index(event.creator(), event.index_by_creator());
+        // Having no event with the same index means no fork, meanwhile multiple and having the same
+        // self_parent means already cast forking accusation.
+        same_index_events
+            .filter_map(|other_event| self.graph.get(other_event))
+            .filter(|other_event| other_event.inner().self_parent() == event.self_parent())
+            .count()
+            == 1
     }
 
     fn detect_invalid_accusation(&mut self, event: &Event<S::PublicId>) {


src/parsec.rs
```diff
@@ -1685,9 +1685,21 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
     // Detect whether the event incurs a fork.
     fn detect_fork(&mut self, event: &Event<S::PublicId>) {
         if self.peer_list.last_event(event.creator()) != event.self_parent() {
-            if let Some(self_parent_hash) = self.graph.self_parent(event).map(|event| *event.hash())
+            let our_accusations = self.accusations_by_peer_since(PeerIndex::OUR, None);
+            if !self
```
jeanphilippeD: consider giving a name to this condition, maybe

```
let accused_peer_of_forking = ....
if !accused_peer_of_forking {
}

File: src/parsec.rs [hunk 2]
```diff
@@ -1685,9 +1685,21 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
     // Detect whether the event incurs a fork.
     fn detect_fork(&mut self, event: &Event<S::PublicId>) {
         if self.peer_list.last_event(event.creator()) != event.self_parent() {
```
madadam: This PR seems fine, but I wonder if we can do better by making the fork detection more robust, thus avoiding the need to detect duplicate accusations, which can be costly. I'm thinking something like this (pseudo-rust):

```rust
for other_event in self.peer_list.peer_events(event.creator()).rev() {
    if other_event == event.self_parent() {
        return; // No fork
    }

    if other_event.self_parent() == event.self_parent() {
        // Fork detected, raise accusation
        self.accuse(...)
    }
}
```

So instead of checking just the last event, we would check all events, starting from the most recent one and iterating down until we either find one with the same self-parent as `event`, which means fork, or we find `event.self_parent()` at which point we can terminate the search and declare no fork. Does it make sense?

maqi:
> Also is there a test for this?

This is the same as to the PR 255 for the test of consensus_with_fork.
@pierrechevalier83 : shall we create a separate task for the test (same to 3288) as well?

src/parsec.rs
jeanphilippeD: Does it pessimize the normal case to improve the corner case?
I assume the `if` is faster than the `for`?

maqi: I think with the `for` approach, for the normal case, we will return at the first event.

madadam: In the no-fork case, my proposal always terminates after the first event. In case of fork, it depends where the fork parent is. If it is close to the top of the graph, we only need to iterate a couple of events. But if it is at the bottom of the graph, we might potentially need to iterate all the events (of that peer). But even in this worst-case scenario I believe the approach is cheaper than looking for all previous accusations (fewer operations).

jeanphilippeD: Ok, I see. I think your approach is reasonable.
One thing to be mindful about, is the accusation list is mostly going to be empty most of the time, and only ever have very few events (it is very costly to be accused).

madadam: Thing is there is no accusation list. We have to iterate through the actual events in the graph and filter those which carry accusation. Which makes is somehow costly.

src/parsec.rs
```diff
@@ -1684,7 +1684,24 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+                other_event.inner().self_parent()
+            } else {
+                continue;
+            };
+            if other_event_self_parent == event.self_parent() {
+                fork_detected = true;
+                break;
+            }
+        }
+        if fork_detected {
```
jeanphilippeD: I am really struggling to understand this loop, it does so much, and there are thing that are unclear to me, so i cannot convince myself that the code is correct.

For example: Why `if Some(other_event) == event.self_parent()` would be no fork if it is not the first one?
It seems what we have here is detect if first fork, if so maybe make that a sub function would help clarify things.

Also maybe worth renaming to `detect_first_fork`.

```
if self.peer_list.last_event(event.creator()) != event.self_parent() {
    if self.is_first_fork(event) {
         if let Some(self_parent_hash) = self.graph.self_parent(event).map(|event| *event.hash()) {
             self.accuse(event.creator(), Malice::Fork(self_parent_hash));
         }
    }
}
```

```
fn is_first_fork(&mut self, event: &Event<S::PublicId>) {
        for other_event in self.peer_list.peer_events(event.creator()).rev().skip(1) {
            if Some(other_event) == event.self_parent() {
                return true;
            }
         return false;
}

// Or better:
fn is_first_fork(&mut self, event: &Event<S::PublicId>) {
         self.peer_list.peer_events(event.creator()).rev().skip(1).any(|other_event| Some(other_event) == event.self_parent() )
}
```


With this formulation I understand.
I also see that there is a problem: if a peer create many forks with the same parent, then we will report as many accusations.

src/parsec.rs
jeanphilippeD: BTW it seems accusations_by_peer_since could be implemented with `
peer_list.peer_events` if there ever was a performance issue, so this approach does not necessarily gain us much.

src/parsec.rs
madadam: > I also see that there is a problem: if a peer create many forks with the same parent, then we will report as many accusations.

Good catch! We need to check how many events with the same `index_by_creator` do we have already, and raise the fork accusation only if there is exactly one. Should be simple.

jeanphilippeD: I'm ok-ish with that.
But i would question whether the original approach was not a more direct/clear/less bug-prone solution to the problem we were trying to solve (i.e avoid duplicate accusations).

maqi: I am also thinking the case that: 
if fork one happens at A_1 (that is A_2_0 and A_2_1 hasing the same self_parent of A_1),
and 
fork two happens at A_3 (that is A_4_0 and A_4_1 hasing the same self_parent of A_2)
Then there will be two accusations to be casted, however we'd prefer only one accusation to be raised.

Given that, I'd prefer reverting back to the original approach. is that ok?

madadam: Those are two different forks though. Do we want to raise only one accusation in that case?

src/parsec.rs
```diff
@@ -1684,11 +1684,21 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+            if let Some(self_parent_hash) = self.graph.self_parent(event).map(|event| *event.hash())
+            {
+                if !self.has_accused_fork(event.creator(), &self_parent_hash) {
+                    self.accuse(event.creator(), Malice::Fork(self_parent_hash));
+                }
+            }
+        }
+    }
+
+    fn is_forking(&self, event: &Event<S::PublicId>) -> bool {
```
madadam: I think you could add something like this to the beginning of this function:

```rust
if self.peer_list.events_by_index(event.index_by_creator()).count() != 1 { 
    return false; 
}
```
and then you wouldn't need `has_accused_fork` at all.

src/parsec.rs
```diff
@@ -1684,14 +1684,41 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+            .peer_list
+            .events_by_index(event.creator(), event.index_by_creator())
+            .count()
+            != 1
+        {
+            return false;
+        }
+        for other_event in self.peer_list.peer_events(event.creator()).rev() {
+            // No fork
+            if Some(other_event) == event.self_parent() {
```
jeanphilippeD: I would really move that out of the loop, it simplify the loop so much because you do not need to get the self_parent(0 within the loop.

jeanphilippeD: Or is that already addressed by the if above, and you can just remove most of the body of this loop?

src/parsec.rs
```diff
@@ -1684,14 +1684,30 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
             {
                 self.accuse(event.creator(), Malice::Fork(self_parent_hash));
             }
         }
     }
 
+    fn is_first_fork(&self, event: &Event<S::PublicId>) -> bool {
+        let mut same_index_events = self
+            .peer_list
+            .events_by_index(event.creator(), event.index_by_creator());
```
madadam: I think you need consider only events that have the same self-parent as `event`, otherwise in a situation like this:

```
A0 --- A1_0 --- A2_0
  \
   \-- A1_1 --- A2_1
```
on inserting `A2_1` we end up accusing the peer of fork because `A2_0` and `A2_1` have the same `index-by-creator`, even though they have different self-parent so no fork should be detected.

maqi: that's why we are still checking `other_event.inner().self_parent() == event.self_parent()` when the count of same_index_events is 1.

madadam: Right, I somehow missed that! Looks good then.

maqi: There will be no fork accusation at A2_2, either on A1_2.
I agree that some unit tests could be helpful, however I think it will be better for a separate task, as discussed previously?

jeanphilippeD: So does it break consensus?

If a peer first see:
```
A0
  \
   \-- A1_1 --- A2_1
                       \
                         \-- A2_2
```
Will they detect and accuse A2_2 as a (first) fork? Will all the peers that have first seen the other branch not?
Is that ok? (everyone will see A0 forking anyway, so the peer will be kicked out. ok?, Leave some unconsensused events. not ok?)

maqi: hmm... I am not getting your question. 
If Carol adds A0, A1_1, A2_1 and A2_2 first, then when adding the other branch, it has to be `A1_0 then A2_0`, which will only cast fork accusation when adding A1_0 (against A0). 
There shall be no fork accusation against A2_2 ?

jeanphilippeD: Carol only see that side of the 1st fork for now. A1_0 & A2_0 do not exist for her yet.
A2_2 is a fork of A2_1, for Carol it is the first fork (imagine the A1_0 and A2_0 do not exists at all, then it is clear that A2_2 is the first fork. It is the same if Carol does not see them yet).

maqi: A2_2 itself won't be considered as a fork, unless you mean A2_2 is pointing to A1_1 ? (i.e. A2.2 and A2.1 has the same self_parent of A1_1) ?
In that case, then it will raise a fork accusation against A1_1.
However, other peers will do the same once they see this fork.
And, Carol will also caust accusation against A0 eventually once it sees the other side (A1_0)

jeanphilippeD: >  (i.e. A2.2 and A2.1 has the same self_parent of A1_1) ?
Yes.

> However, other peers will do the same once they see this fork.
Will they, even if they saw A2_0 first?

maqi: >Will they, even if they saw A2_0 first?
yes, because one fork accusation is agains A1_1, and the other is against A_0

jeanphilippeD: > yes, because one fork accusation is agains A1_1, and the other is against A_0

I cannot convinced myself that this is true because of this code:
```
        if let Some(other_event) = same_index_events.next() {
            if same_index_events.next().is_none() {
```

I see A1_0 & A2_0: no fork so far.
I see A1_1: fork with parent A0.
I see A2_1: no additional fork.
I see A2_2: other_event exists (A2_0), is_none return false (A2_1) => do not detect fork with parent A1_1.

If the same scenario happen with the peer not seeing A1_0, A2_0, then: detect fork with parent A1_1.

maqi: >is_none return false

`next` is a comsuming call. which means the first next call cosumes A2_1 into other_event and empties the same_index_events.

src/parsec.rs
```diff
@@ -1745,14 +1745,30 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
 
+    fn is_first_fork(&self, event: &Event<S::PublicId>) -> bool {
+        let same_index_events = self
+            .peer_list
+            .events_by_index(event.creator(), event.index_by_creator());
+        // Having no event with the same index means no fork, meanwhile multiple and having the same
+        // self_parent means already cast forking accusation.
+        same_index_events
+            .filter(|other_event| {
+                self.graph.get(*other_event).map_or(false, |other_event| {
```
jeanphilippeD: A style suggestion: try to avoid nesting when possible:

```
same_index_events
     .filter_map(|other_event| self.graph.get(*other_event))
     .filter(|other_event| other_event.inner().self_parent() == event.self_parent())
     .count()
     == 1
```

maqi: will take this change when rebasing

PR #262, opened 2019-03-04
MAID-3287 fix/parsec: avoid invalid accomplice accusation

maqi:
Within the current test of concensus_with_fork, at step 34, Eric accused Alice of Fork.
However, at step 47 Carol still accused Eric with Accomplice. This situation happens to many other accusations as well (that is A accused already, but B still cast A as accomplice).
This issue is resovled in this PR by ensuring that: accomplice check is only carried at the events that are not request/response/accusation .

### DIFF
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -1999,9 +1999,9 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
                 })
                 .unwrap_or(false);
 
-            // If this is a Request or an accusation for another malice then the peer might not
-            // have raised the accusation yet.
-            if event.is_request() || is_accusation {
+            // If this is a Request, Response or an accusation for another malice then the peer
+            // might not have raised the accusation yet.
+            if event.is_request() || event.is_response() || is_accusation {
                 return Ok(());
             }
 


src/parsec.rs
```diff
@@ -1991,6 +1991,19 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+                    if !self.graph.is_descendant(last_event, cur_event) {
+                        return Ok(vec![]);
+                    }
+                }
+                _ => return Ok(vec![]),
+            }
+        } else {
+            return Ok(vec![]);
+        }
+
```
octol: Ok I get what the code is doing, and looks good. Although I had hoped that the check at line 1962 would have caught it. Maybe we can perform this check there instead?

maqi: I can merge this with the current checks at 1962 into a separate function, say call it `peer_accusation_raised`.
Once I got the comments from others, I will carry out the update in one shot.

src/parsec.rs
```diff
@@ -1991,6 +1991,19 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         event: EventIndex,
         starting_event: Option<EventIndex>,
     ) -> Result<Accusations<T, S::PublicId>> {
+        if let Some(our_last_event) = self.peer_list.last_event(PeerIndex::OUR) {
```
jeanphilippeD: Even if it is possible to go to Jira task,
Would be nice to have a one sentence in PR or commit description of what 'invalid' we are dealing with.

src/parsec.rs
```diff
@@ -1991,6 +1991,19 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+            match (self.graph.get(our_last_event), self.graph.get(event)) {
+                (Some(last_event), Some(cur_event)) => {
+                    if !self.graph.is_descendant(last_event, cur_event) {
```
jeanphilippeD: Seems like having a `graph` `fn is_descendant(EventIndex,EventIndex)` (or other name) would simplify code in a lot of places including this one.

jeanphilippeD: Could you explain why this works?
I do not understand why checking for this decedant stop the problem described in the JIRA task.

maqi: will have a check.

maqi: The issue is that  we create separate accusation event when detecting a malice. 
However, if the created accusation event comes in with the same `detecting event` that peer detected the malice, we will add that `detecting event` first and think that peer shall be known of the malice but cannot find its accusation. 
Hence we cast the accomplice accusation.
This checks ensures our last_event has the direct link to the `detecting event`, which menas those `created accusation events` has to be known as well.

madadam: Agree. It would make it slightly easier to follow.

src/parsec.rs
madadam: Could you add a short comment explaining why this snippet of code is necessary (something like what you wrote [here](https://github.com/maidsafe/parsec/pull/262/files#r262412748))? It took me quite some time to understand it as it's in no way obvious. That said, as far as I can see it seems correct.

src/parsec.rs
```diff
@@ -1999,9 +1999,9 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
                 })
                 .unwrap_or(false);
 
-            // If this is a Request or an accusation for another malice then the peer might not
-            // have raised the accusation yet.
-            if event.is_request() || is_accusation {
+            // If this is a Request, Response or an accusation for another malice then the peer
+            // might not have raised the accusation yet.
+            if event.is_request() || event.is_response() || is_accusation {
```
pierrechevalier83: What about `requesting`?

maqi: When a peer creats a `requesting` event, it indicates that peer's `adding events` process, i.e. checking malice or cast accusations, has been completed. 
Hence in that case we shall consider a `should-to-be-casted accusation` from that peer has been added.

PR #264, opened 2019-03-06
chore/parsec: remove InvalidGossipCreator and test

Fraser999:


src/parsec.rs
```diff
@@ -1594,10 +1589,6 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         Ok(())
     }
 
-    fn detect_malice_after_process(&mut self, event_index: EventIndex) {
```
madadam: Maybe we can now rename `detect_malice_before_process` to just `detect_malice`?

PR #265, opened 2019-03-07
WIP Combine Opaque Payloads in ConsensusMode::Single with proof.

jeanphilippeD:
This version of the code combine opaque payloads in combined_vote_for.
We will want to replace vote_for with it before pushing.

Observation has an additional recursive variant
`paquePayloadCombined(Vec<Observation<T, P>>)`, Vote gain a
combined_signatures: Vec<P::Signature> to keep the signatures of all
the votes.
=> This is so we can split the vector into consensused blocks with one
   block per Observation.

This is more complicated than I wanted it to be.
It is somewhat complicated by Parsec supporting both Single and SuperMajority
ConsensusMode, and that optimization only work with ConsensusMode::Single.

Of note the original prototype (will push that in different branch) did
not cater for signatures and splitting the observation into multiple
blocks and was thus significantly simpler.

Of note, that optimization can also be implemented outside Parsec for
ConsensusMode::Single:
 - Create a new NetworkEvent (like we did for observation) that contain
   Multiple events with the signatures.
 - Accumulate OpaquePayloads with a signature in this single new NetworkEvent.
 - Before passing handle_request, handle_response or create_gossip, push the
   accumulated event to vote_for.
 - All these are consensused in a single block: need to handle splitting it
   again + merge/split issues.


src/block.rs
```diff
@@ -26,26 +27,50 @@ pub struct Block<T: NetworkEvent, P: PublicId> {
...
-    pub fn new(votes: &BTreeMap<P, Vote<T, P>>) -> Result<Self, Error> {
-        let payload = if let Some(vote) = votes.values().next() {
-            vote.payload().clone()
-        } else {
-            return Err(Error::MissingVotes);
+    pub fn new_blocks(votes: &BTreeMap<P, Vote<T, P>>) -> Result<Vec<Self>, Error> {
+        let (first_public_id, first_vote) = {
+            let mut vote_values = votes.iter();
+
+            let (first_public_id, first_vote) = if let Some((public_id, vote)) = vote_values.next()
```
pierrechevalier83: Maybe you can compactify:
```
let (first_public_id, first_vote) = vote_values.next().expect(Err(Error::MissingVotes))?;
```

PR #271, opened 2019-03-12
feat/test: provide malicious peer for tests

Fraser999:


dot_gen/src/main.rs
```diff
@@ -131,7 +133,7 @@ fn main() {
     let _ = scenarios
         .add("consensus_with_forks", |env| {
             let obs = ObservationSchedule {
-                genesis: peer_ids!("Alice", "Bob", "Carol", "Dave", "Eric"),
+                genesis: Genesis::new(peer_ids!("Alice", "Bob", "Carol", "Dave", "Eric"), 0),
```
pierrechevalier83: Given the number of places that need a similar change to this, I would be for keeping the old  `new` constructor identical and create a new  `new_with_malicious_peers` (or such) constructor. Would also be more explicit than an unidentified magic number.

Fraser999: That `Genesis` enum is a new addition to the code, so the same number of places will need changed.  I can add a second constructor as you suggest, but if you don't like passing a count of malicious peers through it, then maybe the most explicit way to go is just to not provide a constructor.  Then we'd have to do something like:
```rust
genesis: Genesis {
    good: peer_ids!("Alice", "Bob", "Carol", "Dave", "Eric"),
    malicious: BTreeSet::new(),
}
```

pierrechevalier83: I'm not against having `new` that only takes the peer_ids (to be a bit more concise in many places_ but let the ones that want something more specific (with malicious peers) do the verbose construction like you're suggesting (?)

src/parsec.rs
```diff
@@ -2299,39 +2310,61 @@ impl<T: NetworkEvent, S: SecretId> TestParsec<T, S> {
...
+    #[cfg(all(test, feature = "mock"))]
     pub fn new_event_from_observation(
         &mut self,
         self_parent: EventIndex,
         observation: Observation<T, S::PublicId>,
     ) -> Result<Event<S::PublicId>> {
         self.0.new_event_from_observation(self_parent, observation)
     }
+
+    pub fn event_context(&self) -> EventContextRef<T, S> {
```
madadam: Is this and the following functions necessary? `TestParsec` implemented `Deref<Target=Parsec>`.

src/dev_utils/peer.rs
```diff
@@ -78,9 +83,142 @@ pub enum NetworkView {
...
+                _ => false,
+            })
+            .unwrap_or(false);
+
+        if use_fork {
+            let (fork_index, forked_event) = unwrap!(self.forked_event.as_ref());
+            events
+                .into_iter()
+                .take_while(|&event| {
+                    unwrap!(self.test_parsec.graph().get_index(event.hash())) != *fork_index
```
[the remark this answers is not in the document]
Fraser999: I'm not sure it would either.  Also, this method is meant to be equivalent for the case of a malicious peer (hence I called it `events_to_gossip_to_peer` too), and we can't return an event index for the forked event since we didn't add it to our graph.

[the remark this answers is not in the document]
madadam: You could return `EventIndex::PHONY`, but that would be pushing it too hard I think. OK then, let's leave it.

src/parsec.rs
Fraser999: It's just that they need to be made public.

madadam: Right! Makes sense.

src/dev_utils/network.rs
```diff
@@ -285,10 +247,14 @@ impl Network {
         };
         let valid = self
             .peer(sender)
-            .parsec
             .gossip_recipients()
             .any(|valid_recipient| valid_recipient == recipient);
-        let result = self.peer_mut(sender).parsec.create_gossip(recipient);
+        let result = if self.peer(sender).is_malicious() && !self.peer(sender).has_misbehaved() {
```
maqi: hmm...  This sounds like `the malicious` is only about `fork`, as the later on call of `create_gossip_with_fork`.

src/dev_utils/network.rs
```diff
@@ -474,18 +440,22 @@ impl Network {
         Ok(())
     }
 
-    /// Check that no node has been accused of malice.
+    /// Check that no well-behaved peer has been accused of malice.
     fn check_unexpected_accusations(&self, peer_id: &PeerId) -> Result<(), ConsensusError> {
```
maqi: Same as above. Seems the `accusations` are only about fork.

src/dev_utils/schedule.rs
```diff
@@ -37,14 +37,78 @@ pub struct Request {
...
+    fn from_all_ids(mut peer_ids: BTreeSet<PeerId>, malicious_count: usize) -> Self {
+        assert!(malicious_count <= peer_ids.len());
+        if malicious_count == 0 {
+            return Self {
+                ids_of_good_peers: peer_ids,
+                ids_of_malicious_peers: BTreeSet::new(),
+            };
+        }
+
+        let split_off_at_id = unwrap!(peer_ids.iter().rev().nth(malicious_count - 1)).clone();
```
maqi: we are taking malice peers from fixed position, why have to reverse the iterater ?

src/dev_utils/schedule.rs
```diff
@@ -168,6 +232,8 @@ pub enum DelayDistribution {
 pub struct ScheduleOptions {
     /// Size of the genesis group
     pub genesis_size: usize,
+    /// Number of malicious peers included in the genesis group
+    pub malicious_genesis_count: usize,
```
maqi: where do we use this malicious_genesis_count?

src/dev_utils/network.rs
```diff
@@ -192,10 +152,10 @@ impl Network {
 
     /// Returns true if all peers hold the same sequence of stable blocks.
     fn check_blocks_all_in_sequence(&self) -> Result<(), ConsensusError> {
-        let first_peer = unwrap!(self.active_peers().next());
+        let first_peer = unwrap!(self.active_non_malicious_peers().next());
```
maqi: I see that most of the `active_peers` has been replaced by `active_non_malicious_peers`.
Just want to confirm, the only left over place (in Network::execute_event function for ScheduleEvent::AddPeer ) is really required?

src/dev_utils/network.rs
Fraser999: Yes, at the moment there's only a very specific type of malice the new peer type can perform (a fork with just two `Requesting` events having the same self-parent), and only once during the test.

Hopefully we can expand the capabilities of the malicious peer, but I think that's work for the future?

src/dev_utils/network.rs
Fraser999: Same answer :)

src/dev_utils/schedule.rs
Fraser999: I didn't have a strong preference here at all really.  I nearly went for picking the requested number of malicious peers randomly, but I didn't think it was worthwhile.

I guess I ended up favouring good peers then bad since that's the order they're listed in the new `Genesis` type... but there's no reason to require this ordering.

src/dev_utils/network.rs
Fraser999: We're using it to get a list of all "current peers" so we can pass that to the constructor of a new `Parsec`.  While it might not be strictly required right now, I still think we should keep it like that so newly-added peers also accept messages from the malicious peer(s).

src/dev_utils/schedule.rs
maqi: I'd prefer to avoid an extra `rev` operation, however as this won't hurt performance much,
I won't let this block the PR.

PR #272, opened 2019-03-12
Fix clippy and travis

jeanphilippeD:
Make Parsec a workspace to address issues with multi crates and travis.
Fix version discrepancies detected by workspace.
Remove osx clippy from travis.


Cargo.toml
```diff
@@ -49,4 +52,4 @@ required-features = ["mock"]
 
 [[bench]]
 name = "bench"
-harness = false
+harness = false
```
pierrechevalier83: You probably want that newline.

jeanphilippeD: yes, I'll add it with any other fix or when rebasing.

PR #277, opened 2019-03-18
MAID-3293: Fix our_unpolled_observations

madadam:
The bug was that when using `ConsensusMode::Single`, `our_unpolled_observation` sometimes returned observations that have already been polled. This PR also contains a test case exposing the bug.

### DIFF
--- a/dot_gen/src/main.rs
+++ b/dot_gen/src/main.rs
@@ -321,6 +321,29 @@ fn add_functional_tests(scenarios: &mut Scenarios) {
         })
         .seed([411278735, 3293288956, 208850454, 2872654992])
         .file("Alice", "alice.dot");
+
+    let _ = scenarios
+        .add(
+            "functional_tests::our_unpolled_observations_with_consensus_mode_single",
+            |env| {
+                let obs = ObservationSchedule {
+                    genesis: Genesis::new(peer_ids!("Alice", "Bob")),
+                    // Make the votes sufficiently far apart from each other so they are placed
+                    // each in its own block group.
+                    schedule: vec![
+                        (0, Opaque(Transaction::new("A"))),
+                        (1000, Opaque(Transaction::new("A"))),
+                    ],
+                };
+
+                Schedule::from_observation_schedule(env, &ScheduleOptions::default(), obs)
+            },
+        )
+        // Note: Make sure to only use seeds that make Alice reach consensus on Transaction(A) by
+        // Alice before Transaction(A) by Bob.
+        .seed([834576548, 1145967030, 3794692405, 640370552])
+        .consensus_mode(ConsensusMode::Single)
+        .file("Alice", "alice.dot");
 }
 
 fn add_dev_utils_record_smoke_tests(scenarios: &mut Scenarios) {
--- a/src/functional_tests.rs
+++ b/src/functional_tests.rs
@@ -414,12 +414,10 @@ fn our_unpolled_observations_with_consensus_mode_single() {
         x => panic!("Unexpected block payload: {:?}", x),
     }
     assert_eq!(blocks[0].proofs().len(), 1);
-    assert!(blocks[0]
-        .proofs()
-        .iter()
-        .any(|proof| proof.public_id() == alice.our_pub_id()));
+    assert!(blocks[0].is_signed_by(alice.our_pub_id()));
 
-    // Bob's vote is still in, but should not be returned here, as it's not "ours"
+    // Bob's vote is still in, but should not be returned here, as it's not "ours" (from Alice's
+    // point of view).
     assert_eq!(alice.our_unpolled_observations().next(), None);
 }
 
--- a/src/functional_tests.rs
+++ b/src/functional_tests.rs
@@ -10,23 +10,19 @@ use crate::block::Block;
 use crate::dev_utils::{parse_test_dot_file, Record};
 use crate::error::Error;
 use crate::gossip::{Event, Graph, GraphSnapshot};
-use crate::id::PublicId;
+use crate::id::{Proof, PublicId};
 use crate::meta_voting::MetaElectionSnapshot;
 use crate::mock::{self, PeerId, Transaction};
 use crate::observation::{ConsensusMode, Observation};
 use crate::parsec::TestParsec;
 use crate::peer_list::{PeerIndexSet, PeerListSnapshot, PeerState};
-use std::collections::BTreeSet;
+use std::{collections::BTreeSet, fmt::Debug};
 
-macro_rules! assert_err {
-    ($expected_error:pat, $result:expr) => {
-        match $result {
-            Err($expected_error) => (),
-            unexpected => panic!(
-                "Expected {}, but got {:?}",
-                stringify!($expected_error),
-                unexpected
-            ),
+macro_rules! assert_matches {
+    ($actual:expr, $expected:pat) => {
+        match $actual {
+            $expected => (),
+            ref unexpected => panic!("{:?} does not match {}", unexpected, stringify!($expected)),
         }
     };
 }
@@ -67,6 +63,27 @@ impl Snapshot {
 fn nth_event<P: PublicId>(graph: &Graph<P>, n: usize) -> &Event<P> {
     unwrap!(graph.iter_from(n).next()).inner()
 }
+
+/// Testing related extensions to `Iterator`.
+trait TestIterator: Iterator {
+    /// Returns the only element in the iterator. Panics if the iterator yields less or more than
+    /// one element.
+    fn only(mut self) -> Self::Item
+    where
+        Self: Sized,
+        Self::Item: Debug,
+    {
+        let item = unwrap!(self.next(), "Expected one element - got none.");
+        assert!(
+            self.by_ref().peekable().peek().is_none(),
+            "Expected one element - got more (excess: {:?}).",
+            self.collect::<Vec<_>>()
+        );
+        item
+    }
+}
+
+impl<I: Iterator> TestIterator for I {}
 
 #[test]
 fn from_existing() {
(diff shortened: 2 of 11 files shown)


src/parsec.rs
```diff
@@ -429,6 +425,28 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+        let mut matching_blocks = self
+            .consensused_blocks
+            .iter()
+            .flatten()
+            .filter(|block| block.payload() == payload);
+
+        // In `Supermajority` mode, check only if the payload matches, as there can be blocks not
+        // signed by us, yet with payloads voted for by us.
+        // In `Single` mode, on the other hand, check also that we signed it, to avoid false
+        // positives when there are blocks with the same payloads but signed by somone else.
```
Fraser999: Typo "someone".

src/functional_tests.rs
```diff
@@ -398,6 +398,31 @@ fn unpolled_and_unconsensused_observations() {
...
+        Observation::OpaquePayload(_) => (),
+        x => panic!("Unexpected block payload: {:?}", x),
+    }
+    assert_eq!(blocks[0].proofs().len(), 1);
+    assert!(blocks[0]
+        .proofs()
+        .iter()
+        .any(|proof| proof.public_id() == alice.our_pub_id()));
+
+    // Bob's vote is still in, but should not be returned here, as it's not "ours"
```
Fraser999: Maybe change to "as it's not Alilce's"?

src/functional_tests.rs
```diff
@@ -398,6 +398,31 @@ fn unpolled_and_unconsensused_observations() {
...
+    assert_eq!(alice.our_unpolled_observations().next(), None);
```
jeanphilippeD: I would recommend following Arrange/Act/Assert pattern, and combine assertion when possible.

```
// Arrange
let mut alice = Record::from(parse_test_dot_file("alice.dot")).play();

//Act
let (genesis_blocks, alice_block) = (alice.poll(), alice.poll());
let our_unpolled = alice.our_unpolled_observations().collect_vec();

// Assert
let keep_non_genesis = |block: &block| {
       match block.payload() {
            Observation::Genesis(_) => None,
            _ => Some(block),
    } 
};

let keep_non_opaque = |block: &block| {
       match block.payload() {
            Observation::OpaquePayload(_) => None,
            _ => Some(block),
    } 
};

let to_proofs = |block: &block| {
     block.proofs().iter().map(|proof| proof.public_id()).collect_vec()
    }

// Maybe I'd combine all these assert into one with a tuple or struct so on error, all the info are shown.
assert_eq!(
    genesis_blocks.map(keep_non_genesis).collect_vec(),
    vec![None]
);
assert_eq!(
    alice_block.map(keep_non_opaque).collect_vec()),
    (vec![None], vec![None])
);
assert_eq!(
    alice_block.map(to_proofs).collect_vec()),
    vec![vec![alice.our_pub_id()]]
);
assert_eq!(
    our_unpolled,
    vec![]);

File: src/functional_tests.rs [hunk 4]
```diff
@@ -398,6 +398,31 @@ fn unpolled_and_unconsensused_observations() {
     assert!(unpolled_observations.next().is_none());
 }
 
+#[test]
+fn our_unpolled_observations_with_consensus_mode_single() {
+    let mut alice = Record::from(parse_test_dot_file("alice.dot")).play();
```
jeanphilippeD: Wouldn't we be better off using a schedule directly for this test, as oppose to parsing a dot file?

jeanphilippeD: I like what you did with the consensus mode, but I'm not sure I understand why you needed it. All the benchmark play in Single Mode and did not need change in the dot file format. Would some of the change you did not necessary only if we loaded parsec from dot_input?

src/gossip/cause.rs
```diff
@@ -205,8 +206,7 @@ impl Cause<VoteKey<PeerId>, EventIndex, PeerIndex> {
                 other_parent,
             },
             Cause::Observation { vote, .. } => {
-                let (vote_key, observation) =
-                    VoteKey::new(vote, creator, ConsensusMode::Supermajority);
+                let (vote_key, observation) = VoteKey::new(vote, creator, consensus_mode);
```
jeanphilippeD: I'm not sure we have any code that exercise this code with consensus mode not super-majority. I do not think your new test does?

dot_gen/src/main.rs
```diff
@@ -321,6 +321,29 @@ fn add_functional_tests(scenarios: &mut Scenarios) {
...
+                        (1000, Opaque(Transaction::new("A"))),
+                    ],
+                };
+
+                Schedule::from_observation_schedule(env, &ScheduleOptions::default(), obs)
+            },
+        )
+        // Note: Make sure to only use seeds that make Alice reach consensus on Transaction(A) by
+        // Alice before Transaction(A) by Bob.
+        .seed([834576548, 1145967030, 3794692405, 640370552])
```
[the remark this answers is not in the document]
madadam: The observation schedule currently doesn't support explicitly specifying who should cast the vote. I needed Alice's vote to come before Bob's (the test setup needs it) and I think the only way was to find a seed where that happens randomly.

src/functional_tests.rs
madadam: In my mind the schedule is more for randomized integration tests. Functional tests should not be random. Also, if we used schedule here, why not use it everywhere else?

madadam: Hm. I now see there is `Record::set_consensus_mode` which I somehow missed. Do you want me to revert the changes use that instead?

src/gossip/cause.rs
madadam: It probably doesn't. Depends on how [this](https://github.com/maidsafe/parsec/pull/277#discussion_r266448888) is resolved, I might revert it.

src/functional_tests.rs
```diff
@@ -398,6 +398,31 @@ fn unpolled_and_unconsensused_observations() {
...
+        x => panic!("Unexpected block payload: {:?}", x),
+    }
+
+    let blocks = unwrap!(alice.poll());
+    match blocks[0].payload() {
+        Observation::OpaquePayload(_) => (),
+        x => panic!("Unexpected block payload: {:?}", x),
+    }
+    assert_eq!(blocks[0].proofs().len(), 1);
+    assert!(blocks[0]
```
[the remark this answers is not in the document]
madadam: I like the `is_signed_by` suggestion. I was actually thinking about exactly that, but in the end decided to be lazy and not do it. You changed my mind.

src/functional_tests.rs
jeanphilippeD: No, they seem like a good thing to have.

jeanphilippeD: So I'm happy for you to decide the best course of action.

src/functional_tests.rs
madadam: I'm failing to see the benefit of your approach. Mine is shorter and I believe clearer.

jeanphilippeD: http://wiki.c2.com/?ArrangeActAssert
"Clearly separates what is being tested from the setup and verification steps. "
=> What I found is that this result in more maintainable code as it avoid test growing arms and legs.

There are many way to code the Assert section thought.
The main difference is I'm testing more completely the collection (not assume it is of size 1, and not separate assertion on a single item into 2 steps). I found testing collections as a whole produce clearer code and test results

There are other way to apply the same principle though, I'm sure many much better than what I suggested, and I'd be happy with any version that take these into account (But I'd live with just following AAA pattern AAA comments not needed if clear).

Fraser999: I feel that AAA approach is more appropriate for unit tests rather than larger functional ones.  I prefer the original layout too.

jeanphilippeD: AAA is more unit test centric, but best testing practices tend to move away from free-form tests and provide structure to ensure consistency in tests. BDD (https://en.wikipedia.org/wiki/Behavior-driven_development) also give a very similar structure to tests with Given/When/Then, which basically serve the same purpose, and this is definitely a Functional testing practice.

I'll carry on pushing for structure, but I would not block this PR for this.

src/functional_tests.rs
```diff
@@ -10,23 +10,19 @@ use crate::block::Block;
...
-            unexpected => panic!(
-                "Expected {}, but got {:?}",
-                stringify!($expected_error),
-                unexpected
-            ),
+macro_rules! assert_matches {
+    ($actual:expr, $expected:pat) => {
+        match $actual {
+            $expected => (),
+            ref unexpected => panic!("{:?} does not match {}", unexpected, stringify!($expected)),
```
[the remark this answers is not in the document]
madadam: > Any reason to switch the order? I know CCPUNIT used expected first.

No strong reason. Just I think it reads more naturally this way. Also there is a [crate](https://crates.io/crates/matches) which implements very similar macro which uses this order. I thought about using it first, but then decided it's not worth bringing another dependency for such a trivial thing.

src/functional_tests.rs
```diff
@@ -402,19 +417,15 @@ fn unpolled_and_unconsensused_observations() {
 fn our_unpolled_observations_with_consensus_mode_single() {
     let mut alice = Record::from(parse_test_dot_file("alice.dot")).play();
 
-    let blocks = unwrap!(alice.poll());
-    match blocks[0].payload() {
-        Observation::Genesis(_) => (),
-        x => panic!("Unexpected block payload: {:?}", x),
-    }
+    let block = alice.poll().into_iter().flatten().only();
```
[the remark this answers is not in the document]
madadam: I guess. Not sure it matters in this case though.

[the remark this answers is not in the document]
jeanphilippeD: Only as far as consistent effective testing practice matters in maintaining a very large body of tests. I would not block this PR for this.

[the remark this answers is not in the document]
jeanphilippeD: it is much more important to have a test :-), so seeing this test already makes me happy.

PR #279, opened 2019-03-19
MAID-3081: Add test for invalid self parent

madadam:


src/functional_tests.rs
```diff
@@ -1206,4 +1206,27 @@ mod handle_malice {
...
+        let mut request = unwrap!(alice.create_gossip(&bob_id));
+
+        // Remove the second event from the request, so the event after it ends up with invalid
+        // self-parent.
+        let _ = request.packed_events.remove(1);
+
+        // Create another Parsec and handle the above request. It should return error.
+        let mut bob = TestParsec::from_genesis(bob_id, &genesis, ConsensusMode::Supermajority);
+        let result = bob.handle_request(&alice_id, request);
+        assert_err!(Error::UnknownSelfParent, result);
```
maqi: do we also need to check whether three is any event in the request has been added or not added into the graph of Bob?

src/functional_tests.rs
```diff
@@ -1206,4 +1206,27 @@ mod handle_malice {
...
+        // Create Parsec with three events: Initial, Genesis and OpaquePayload:
+        let mut alice =
+            TestParsec::from_genesis(alice_id.clone(), &genesis, ConsensusMode::Supermajority);
+        unwrap!(alice.vote_for(Observation::OpaquePayload(Transaction::new("X"))));
+
+        let mut request = unwrap!(alice.create_gossip(&bob_id));
+
+        // Remove the second event from the request, so the event after it ends up with invalid
+        // self-parent.
+        let _ = request.packed_events.remove(1);
```
jeanphilippeD: Is it possible to create that packed event without running Alice parsec at all (or desirable, i think it might well be)?

I'm wondering what is missing. new_from_observation first create packed content, then compute the signature, so I'm thinking we are probably not very far to be able to have a test function `packed_event_from_observation(observation, packed_parent)`.

One thing that is not super clear is which event we removed is it genesis, initial event, opaque payload or requesting.
That would be obvious if we were constructing the request, especially since we are not testing Alice here, bob, but that is not so clear.

src/gossip/event_context.rs
```diff
@@ -33,3 +35,38 @@ impl<'a, T: NetworkEvent, S: SecretId> Clone for EventContextRef<'a, T, S> {
 
 // ...neither does `#[derive(Copy)]`.
 impl<'a, T: NetworkEvent, S: SecretId> Copy for EventContextRef<'a, T, S> {}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::mock::{PeerId, Transaction};
+
+    pub(crate) struct EventContext {
```
jeanphilippeD: I would probably have split the move into a refactor commit to highlight what changed, but the commit is small enough anyway.

src/functional_tests.rs
```diff
@@ -1206,4 +1207,41 @@ mod handle_malice {
...
+        let bob_id = PeerId::new("Bob");
+        let genesis = btree_set![alice_id.clone(), bob_id.clone()];
+
+        let mut alice = EventContext::new(alice_id.clone());
+        let _ = alice
+            .peer_list
+            .add_peer(bob_id.clone(), PeerState::active());
+
+        let a_0 = Event::new_initial(alice.as_ref());
+        let a_0_index = alice.graph.insert(a_0).event_index();
```
jeanphilippeD: I would probably move all the above in a sub function `get_context_with_initial_event` or something.

src/gossip/event_context.rs
```diff
@@ -33,3 +35,38 @@ impl<'a, T: NetworkEvent, S: SecretId> Clone for EventContextRef<'a, T, S> {
 
 // ...neither does `#[derive(Copy)]`.
 impl<'a, T: NetworkEvent, S: SecretId> Copy for EventContextRef<'a, T, S> {}
+
+#[cfg(test)]
+mod tests {
```
maqi: we normally only put `test cases` under the `mod tests`, however here it seems we put a struct definition used for testing purpose inside a `mod tests`?
shall we better rename it?

madadam: We also have test utilities, not just tests themselves under `mod tests` in a couple of places, so I think this is fine? I don't think renaming the module to something like `test_utils` would bring much value.

PR #281, opened 2019-03-20
MAID-3282: add malice protection for Requesting type

Fraser999:
This PR adds validation that incoming messages are correct (wrt the sender correctly following the `Requesting` → `Request` → `Response` flow) and also similar validation for all sync events as they're added to the graph.

In the case of the former, the message isn't handled at all (no events are added to our graph) and an accusation of `UnprovableMalice` is made against the sender.

To allow efficient checking that a sync event's corresponding descendant hasn't yet been added, a hash-set containing all the `Requesting` and `Request` sync events which don't yet have their partner `Request` or `Response` sync event has been added to `Graph`.

This also required the `dev_utils::Record` to be updated for the case where, at the end of iterating through the parsed dot file creating actions, we send the final few events via a new Request.  For that message to be valid, we now also create a `Requesting` sync event by the sender.  This means that after running `Record::play()`, the parsec instance will have two events not contained in the input dot file.

### DIFF
--- a/src/dev_utils/record.rs
+++ b/src/dev_utils/record.rs
@@ -8,7 +8,7 @@
 
 use super::dot_parser::{parse_dot_file, ParsedContents};
 use crate::gossip::IndexedEventRef;
-use crate::gossip::{Cause, Event, Request, Response};
+use crate::gossip::{Cause, Event, PackedEvent, Request, Response};
 use crate::hash::Hash;
 use crate::mock::{PeerId, Transaction};
 use crate::observation::{ConsensusMode, Observation, ObservationKey, ObservationStore};
@@ -26,12 +26,13 @@ pub struct Record {
     our_id: PeerId,
     genesis_group: BTreeSet<PeerId>,
     actions: Vec<Action>,
-
     // Keys of the consensused blocks' payloads in the order they were consensused.
     consensus_history: Vec<ObservationKey>,
-
     // Consensus mode to play
     consensus_mode: ConsensusMode,
+    // True if when parsing the graph we had to add a final `Requesting` sync event and schedule a
+    // Request to be sent to us in order to learn of any remaining events.
+    added_final_requesting_event: bool,
 }
 
 impl Record {
@@ -70,8 +71,7 @@ impl From<ParsedContents> for Record {
             contents
                 .graph
                 .iter()
-                .filter_map(|event| extract_genesis_group(event.inner(), &contents.observations))
-                .next()
+                .find_map(|event| extract_genesis_group(event.inner(), &contents.observations))
                 .cloned(),
             "No event carrying Observation::Genesis found"
         );
@@ -84,40 +84,6 @@ impl From<ParsedContents> for Record {
         let mut actions = Vec::new();
         let mut skip_our_accusations = false;
         let mut known = vec![false; contents.graph.len()];
-
-        let collect_event_to_gossip = |other_parent: IndexedEventRef<_>, known: &mut Vec<bool>| {
-            let mut events_to_gossip = Vec::new();
-            let other_parent_tindex = other_parent.topological_index();
-            for event in contents.graph.ancestors(other_parent) {
-                let event_tindex = event.topological_index();
-                if known[event_tindex] && other_parent_tindex != event_tindex {
-                    continue;
-                } else {
-                    known[event_tindex] = true;
-                }
-
-                events_to_gossip.push(unwrap!(event.pack(contents.event_context())));
-            }
-            events_to_gossip.reverse();
-            events_to_gossip
-        };
-
-        let collect_remaining_event_to_gossip = |known: &mut Vec<bool>| {
-            let mut events_to_gossip = Vec::new();
-            let mut last_event = None;
-            for event in contents.graph.iter() {
-                let event_tindex = event.topological_index();
-                if known[event_tindex] {
-                    continue;
-                } else {
-                    known[event_tindex] = true;
-                }
-
-                events_to_gossip.push(unwrap!(event.pack(contents.event_context())));
-                last_event = Some(event);
-            }
-            (events_to_gossip, last_event)
-        };
 
         for event in &contents.graph {
             if event.topological_index() == 0 {
@@ -168,7 +134,8 @@ impl From<ParsedContents> for Record {
                         .get(other_parent.creator())
                         .map(|peer| peer.id().clone()));
 
-                    let events_to_gossip = collect_event_to_gossip(other_parent, &mut known);
+                    let events_to_gossip =
+                        collect_events_to_gossip(&contents, other_parent, &mut known);
 
                     if event.is_request() {
                         actions.push(Action::Request(src, Request::new(events_to_gossip)))
@@ -197,13 +164,14 @@ impl From<ParsedContents> for Record {
             }
         }
 
-        // Need a request to our peer with the remaining events
-        let (events_to_gossip, other_parent) = collect_remaining_event_to_gossip(&mut known);
-        if let Some(other_parent) = other_parent {
-            let src = unwrap!(contents
-                .peer_list
-                .get(other_parent.creator())
-                .map(|peer| peer.id().clone()));
+        let mut events_to_gossip = collect_remaining_events_to_gossip(&contents, &mut known);
+        let added_final_requesting_event = !events_to_gossip.is_empty();
+        if let Some(packed_event) = events_to_gossip.last() {
+            let src = packed_event.creator().clone();
+            let self_parent = packed_event.compute_hash();
+            let requesting_event =
+                PackedEvent::new_requesting(src.clone(), contents.our_id.clone(), self_parent);
+            events_to_gossip.push(requesting_event);
             actions.push(Action::Request(src, Request::new(events_to_gossip)));
         }
 
@@ -213,6 +181,7 @@ impl From<ParsedContents> for Record {
             actions,
             consensus_history: contents.meta_election.consensus_history,
             consensus_mode: contents.consensus_mode,
+            added_final_requesting_event,
         }
     }
 }
@@ -257,19 +226,62 @@ fn extract_genesis_group<'a>(
         })
 }
 
+fn collect_events_to_gossip(
+    contents: &ParsedContents,
+    other_parent: IndexedEventRef<PeerId>,
+    known: &mut Vec<bool>,
+) -> Vec<PackedEvent<Transaction, PeerId>> {
+    let mut events_to_gossip = Vec::new();
+    let other_parent_tindex = other_parent.topological_index();
+    for event in contents.graph.ancestors(other_parent) {
+        let event_tindex = event.topological_index();
+        if known[event_tindex] && other_parent_tindex != event_tindex {
+            continue;
+        } else {
+            known[event_tindex] = true;
+        }
+
+        events_to_gossip.push(unwrap!(event.pack(contents.event_context())));
+    }
+    events_to_gossip.reverse();
+    events_to_gossip
+}
+
+fn collect_remaining_events_to_gossip(
+    contents: &ParsedContents,
+    known: &mut Vec<bool>,
+) -> Vec<PackedEvent<Transaction, PeerId>> {
+    let mut events_to_gossip = Vec::new();
+    for event in contents.graph.iter() {
+        let event_tindex = event.topological_index();
+        if known[event_tindex] {
+            continue;
+        } else {
+            known[event_tindex] = true;
+        }
+
+        events_to_gossip.push(unwrap!(event.pack(contents.event_context())));
+    }
+    events_to_gossip
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
     use crate::parsec::get_graph_snapshot;
-    use std::cmp;
+    use std::fs;
+    use std::iter;
+    use std::path::PathBuf;
+    use std::thread;
+    use walkdir::WalkDir;
 
     #[derive(PartialEq, Eq, Debug)]
-    struct TrucatedHashes<'th> {
+    struct TruncatedHashes<'th> {
         actual_len: usize,
         hashes: &'th [Hash],
     }
 
-    impl<'th> TrucatedHashes<'th> {
+    impl<'th> TruncatedHashes<'th> {
         fn new_with_one_missing_element(hashes: &'th [Hash]) -> Self {
             Self {
                 actual_len: hashes.len() + 1,
@@ -277,26 +289,45 @@ mod tests {
             }
         }
 
-        fn new_with_one_trucated_element(hashes: &'th [Hash]) -> Self {
+        fn new_with_one_truncated_element(hashes: &'th [Hash]) -> Self {
             Self {
                 actual_len: hashes.len(),
-                hashes: &hashes[0..cmp::max(hashes.len(), 1) - 1],
+                hashes: &hashes[0..hashes.len().saturating_sub(1)],
+            }
+        }
+    }
+
+    // The path of the dot file used to construct the Record.  Used for printing while panicking.
+    struct PathPrinter(PathBuf);
+
+    impl Drop for PathPrinter {
+        fn drop(&mut self) {
+            if thread::panicking() {
+                let msg = format!("!  Record constructed from '{}'  !", self.0.display());
+                let border = iter::repeat('!').take(msg.len()).collect::<String>();
+                println!("\n{1}\n{}\n{1}\n", msg, border);
             }
         }
     }
 
     /// Run smoke test using given dot file.
-    /// Use ignore_last_events to skip fake request that replay generates to send remaining gossip events.
-    fn smoke(path: &str, ignore_last_events: usize) {
-        let contents = unwrap!(parse_dot_file(path));
+    fn smoke<P: AsRef<Path>>(path: P) {
+        let _p = PathPrinter(path.as_ref().to_owned());
+        let contents = unwrap!(parse_dot_file(path.as_ref()));
         let expected = Parsec::from_parsed_contents(contents);
         let expected_events = {
             let ignore_last_events = 0;
             get_graph_snapshot(&expected, ignore_last_events)
         };
 
-        let contents = unwrap!(parse_dot_file(path));
-        let replay = Record::from(contents);
+        let replay = unwrap!(Record::parse(path));
+        let ignore_last_events = if replay.added_final_requesting_event {
+            // Ignore the `Requesting` event we created when parsing the graph, and the associated
+            // `Request` we'll create when receiving the message.
+            2
+        } else {
+            0
+        };
         let actual = replay.play();
         let actual_events = get_graph_snapshot(&actual, ignore_last_events);
 
@@ -304,10 +335,10 @@ mod tests {
     }
 
     /// Run smoke test using given dot file checking the consensus history.
-    /// missing_one_consensus=true if the dump-graphs was taken when consensus is reached (last block missing).
+    /// `missing_one_consensus=true` if the dump-graphs was taken when consensus is reached (last
+    /// block missing).
     fn smoke_consensus_history(path: &str, missing_one_consensus: bool) {
-        let contents = unwrap!(parse_dot_file(path));
-        let replay = Record::from(contents);
+        let replay = unwrap!(Record::parse(path));
         let expected_history = replay.consensus_history();
 
         let actual = replay.play();
@@ -315,8 +346,8 @@ mod tests {
 
         if missing_one_consensus {
             assert_eq!(
-                TrucatedHashes::new_with_one_missing_element(&expected_history),
-                TrucatedHashes::new_with_one_trucated_element(&actual_history)
+                TruncatedHashes::new_with_one_missing_element(&expected_history),
+                TruncatedHashes::new_with_one_truncated_element(&actual_history)
             );
         } else {
             assert_eq!(expected_history, actual_history);
@@ -325,26 +356,39 @@ mod tests {
 
     #[test]
     fn smoke_parsec() {
-        let ignore_last_events = 0;
-        smoke("input_graphs/benches/minimal.dot", ignore_last_events)
+        let is_dot_file = |path: &PathBuf| {
+            path.extension()
+                .map(|extension| extension.to_string_lossy() == "dot")
+                .unwrap_or(false)
+        };
+        // 50kB seems to strike a reasonable balance between including quite a few dot files, while
+        // not having too big of an impact on the test's running time.
+        let max_file_size = 50_000;
+        let mut checked_at_least_one_file = false;
+        for path in WalkDir::new("input_graphs")
+            .into_iter()
+            .map(|entry| unwrap!(entry).path().to_owned())
+            .filter(is_dot_file)
+            .filter(|path| unwrap!(fs::metadata(path)).len() < max_file_size)
+        {
+            smoke(path);
+            checked_at_least_one_file = true;
+        }
+        assert!(
+            checked_at_least_one_file,
+            "All dot files are over {} bytes, so none were checked.",
+            max_file_size
+        );
     }
 
     #[test]
     fn smoke_other_peer_names() {
-        let ignore_last_events = 1;
-        smoke(
-            "input_graphs/dev_utils_record_tests_smoke_other_peer_names/annie.dot",
-            ignore_last_events,
-        )
+        smoke("input_graphs/dev_utils_record_tests_smoke_other_peer_names/annie.dot")
     }
 
     #[test]
     fn smoke_routing() {
-        let ignore_last_events = 1;
-        smoke(
-            "input_graphs/dev_utils_record_tests_smoke_routing/minimal.dot",
-            ignore_last_events,
-        )
+        smoke("input_graphs/dev_utils_record_tests_smoke_routing/minimal.dot")
     }
 
     #[test]
--- a/src/functional_tests.rs
+++ b/src/functional_tests.rs
@@ -15,7 +15,9 @@ use crate::meta_voting::MetaElectionSnapshot;
 use crate::mock::{self, PeerId, Transaction};
 use crate::observation::{ConsensusMode, Observation};
 use crate::parsec::TestParsec;
-use crate::peer_list::{PeerIndexSet, PeerListSnapshot, PeerState};
+#[cfg(feature = "malice-detection")]
+use crate::peer_list::PeerIndexSet;
+use crate::peer_list::{PeerListSnapshot, PeerState};
 use std::{collections::BTreeSet, fmt::Debug};
 
 macro_rules! assert_matches {
@@ -241,8 +243,8 @@ fn add_peer() {
     // Generated with RNG seed: [411278735, 3293288956, 208850454, 2872654992].
     let mut parsed_contents = parse_test_dot_file("alice.dot");
 
-    // The final decision to add Fred is reached in D_18, so pop this event for now.
-    let d_18 = unwrap!(parsed_contents.remove_last_event());
+    // The final decision to add Fred is reached in E_25, so pop this event for now.
+    let e_25 = unwrap!(parsed_contents.remove_last_event());
 
     let mut alice = TestParsec::from_parsed_contents(parsed_contents);
     let genesis_group: BTreeSet<_> = alice
@@ -252,7 +254,6 @@ fn add_peer() {
         .collect();
 
     let alice_id = PeerId::new("Alice");
-    let dave_id = PeerId::new("Dave");
     let fred_id = PeerId::new("Fred");
 
     assert!(!alice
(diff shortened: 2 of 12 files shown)


src/gossip/graph/mod.rs
```diff
@@ -155,12 +177,116 @@ impl<P: PublicId> Graph<P> {
...
+                #[cfg(any(test, feature = "testing"))]
+                assert!(other_parent.is_requesting());
+                other_parent.index
+            });
+            let _removed =
+                index_to_remove.map(|index| self.awaiting_associated_events.remove(&index));
+            #[cfg(any(test, feature = "testing"))]
+            assert_eq!(_removed, Some(true));
+        } else if event.is_response() {
+            let index_to_remove = self.associated_request(event).map(|associated_request| {
```
jeanphilippeD: why do you need associated_request instead of the other_parent index?

I do not understand, this function seem significantly more complicated than I would expect, what am i missing?

If adding to the graph a request or requesting, then add it to the hashset.
If adding to the graph a response or request, then remove the other parent hashset.

What am i missing?

src/gossip/graph/mod.rs
```diff
@@ -155,12 +177,116 @@ impl<P: PublicId> Graph<P> {
...
+
 #[cfg(any(all(test, feature = "mock"), feature = "testing"))]
 impl<P: PublicId> Graph<P> {
     /// Remove the topologically last event.
     pub fn remove_last(&mut self) -> Option<(EventIndex, Event<P>)> {
         let event = self.events.pop()?;
         let _ = self.indices.remove(event.hash());
+        #[cfg(feature = "malice-detection")]
+        {
+            let _ = self
```
jeanphilippeD: This only work if no malice, no?
Also, if awaiting_associated_events is a struct, then most should be implemented there.

src/gossip/messages.rs
```diff
@@ -24,8 +34,42 @@ impl<T: NetworkEvent, P: PublicId> Request<T, P> {
...
 
-    pub(crate) fn hash_of_last_event_created_by(&self, src: &P) -> Result<Option<EventHash>> {
-        hash_of_last_event_created_by(src, &self.packed_events)
+    /// Returns `Ok` if the final event is a `Requesting` by `sender` targeting us, and for which we
+    /// have not yet recorded an associated `Request` event.  Otherwise returns `Err`.
+    #[cfg(feature = "malice-detection")]
+    pub(crate) fn validate<S: SecretId<PublicId = P>>(
+        &self,
+        sender: &P,
+        ctx: EventContextRef<T, S>,
```
jeanphilippeD: That we need to pass the context suggest this is not the right place for that function.
Would it be better in parsec_helper or new file?

src/gossip/messages.rs
```diff
@@ -41,24 +85,64 @@ impl<T: NetworkEvent, P: PublicId> Response<T, P> {
...
+    ///   * for which we have not yet recorded an associated `Response` event
+    ///   * or if such an event immediately precedes a series of accusations by `sender`.
+    /// Otherwise returns `Err`.
+    #[cfg(feature = "malice-detection")]
+    pub(crate) fn validate<S: SecretId<PublicId = P>>(
+        &self,
+        sender: &P,
+        ctx: EventContextRef<T, S>,
+    ) -> Result<()> {
+        for packed_event in self.packed_events.iter().rev() {
```
jeanphilippeD: It feels somewhat more complicated than I would hope.
it seems it would be simpler to do as an in process thing since we are not trying to fight spam (i.e we will have processed all valid graph part and place them in the graph, but removed the invalid request/reponse and its descendent.

If it is done inside the unpack_and_add_events, as we are processing event, we can check each request other parent is a requesting with the correct destination, and each response other parent is a request with other parent having same creator as the response, and bail like we do for some of our malice detection.

src/dev_utils/record.rs
```diff
@@ -162,13 +128,19 @@ impl From<ParsedContents> for Record {
                         "Sync event without other-parent: {:?}",
                         *event
                     );
+                    if event.is_request() {
+                        assert!(other_parent.is_requesting());
```
maqi: better put some msg in case of assertion failure?
same comment to the other assertion below.

src/dev_utils/record.rs
```diff
@@ -197,13 +169,15 @@ impl From<ParsedContents> for Record {
...
 
-        // Need a request to our peer with the remaining events
-        let (events_to_gossip, other_parent) = collect_remaining_event_to_gossip(&mut known);
-        if let Some(other_parent) = other_parent {
-            let src = unwrap!(contents
-                .peer_list
-                .get(other_parent.creator())
-                .map(|peer| peer.id().clone()));
+        let mut events_to_gossip = collect_remaining_events_to_gossip(&contents, &mut known);
+        let added_final_requesting_event = !events_to_gossip.is_empty();
```
maqi: looking at this implementation, it seems the comment (`True if when parsing the graph we had`) of `added_final_requesting_event` and the name is bit confusing.
Do you mean whenever we parse the graph, we shall `add a final Requesting sync event and schedule a Request ` ?

src/dev_utils/record.rs
```diff
@@ -197,13 +169,15 @@ impl From<ParsedContents> for Record {
...
+        if let Some(packed_event) = events_to_gossip.last() {
+            // Need a request to our peer with the remaining events
+            let src = packed_event.creator().clone();
```
maqi: will there be chance that the last event in the `events_to_gossip` is not this peer?
in that case, `src` will be wrong?

src/gossip/graph/mod.rs
Fraser999: A lot of the complexity comes from handling response messages - more so than requests.

We need to ensure that the events in a response message should finish with a `Request` sync event, created by message sender, and that needs to have a `Requesting` as its other parent which is created by us and specifies the response sender as the recipient inside the `Requesting` event.

That's kinda complex, but what adds more is that a valid response message can have such a `Request` sync event not quite as the final one; that could validly be followed by a series of accusations all made by the sender.

In that case, when we receive that message and create our `Response` event, that will have the final accusation by the sender as its `other_parent`.  Hence the need to use `associated_request` here rather than just `other_parent`.

src/gossip/graph/mod.rs
```diff
@@ -24,13 +26,19 @@ use std::collections::BTreeSet;
 pub(crate) struct Graph<P: PublicId> {
     events: Vec<Event<P>>,
     indices: BTreeMap<EventHash, EventIndex>,
+    /// Indices of `Requesting` events with no associated descendant `Request`, and `Request`s with
+    /// no associated descendant `Response`.
+    #[cfg(feature = "malice-detection")]
+    awaiting_associated_events: FnvHashSet<EventIndex>,
```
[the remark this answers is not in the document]
Fraser999: I did consider that, but I don't think it would be worthwhile.  The bulk of the complexity around usage of `awaiting_associated_events` is in deducing what value should be passed as `other_parent` in your `insert()` method, and that's work which _should_ be done here in graph.rs rather than in a new struct.

If you disagree, I can try implementing this and see how it goes.

src/gossip/graph/mod.rs
Fraser999: Good point, I hadn't considered that last event being a malicious sync event.

src/gossip/messages.rs
Fraser999: The only other place I'd see this belonging is in parsec.rs really.  I'd be fine to move the two `validate()` methods there if you prefer?  I don't like moving it to parsec_helper or a new file since not only would that still need the context passed, but also the message.

An advantage to keeping it here is that we can destructure `packed_event.content.cause`.  If I move it to parsec, I think I'd need to add one or two getters to `PackedEvent` similar to those on `Event` like `requesting_recipient()`.

src/functional_tests.rs
```diff
@@ -241,8 +243,8 @@ fn add_peer() {
     // Generated with RNG seed: [411278735, 3293288956, 208850454, 2872654992].
     let mut parsed_contents = parse_test_dot_file("alice.dot");
 
-    // The final decision to add Fred is reached in D_18, so pop this event for now.
-    let d_18 = unwrap!(parsed_contents.remove_last_event());
+    // The final decision to add Fred is reached in E_25, so pop this event for now.
+    let e_25 = unwrap!(parsed_contents.remove_last_event());
```
maqi: this change is due to the change in the req-resp pattern, or something else?

src/gossip/graph/mod.rs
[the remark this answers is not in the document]
jeanphilippeD: I agree, and think it is worthwhile.
I think while the object will be simple, it will simplify the complex code by removing some of the moving pieces.

src/gossip/messages.rs
Fraser999: The issue is that we'd need to keep more state to track the events as we process them, or somehow flag the event which we think is the one which will decide whether we create our own sync event on the back of it.

Say I send a malicious response message which as its last packed events contains my malicious `Request` event, then a valid accusation made by me.  When you're processing the `Request`, you don't yet know if that final event (the accusation) is a valid `Request`, and that would change whether or not we want to create our own `Response` sync event.

jeanphilippeD: Not sure.
We do something similar for .

```
    fn add_event(&mut self, event: Event<S::PublicId>) -> Result<EventIndex> {
        let our = event.creator() == PeerIndex::OUR;
        if !our {
            self.detect_malice(&event)?;
        }

...
// for example that one will stop the event from being added.
self.detect_other_parent_by_same_creator(event)?;
```


What I mean, is if we find any event that does not match the correct requesting/request/response, then we get an error, and stop adding any events left in the gossip, like we do for detect_other_parent_by_same_creator. This will also happen when creating our own sync event, i.e if the gossip was not correct, we will keep it, but not link it to a response sync event for example.

src/gossip/graph/mod.rs
```diff
@@ -68,6 +85,11 @@ impl<P: PublicId> Graph<P> {
             }
         };
 
+        if needs_updated {
+            #[cfg(feature = "malice-detection")]
```
maqi: why not put the feature guard outside the if clause?

src/gossip/graph/mod.rs
```diff
@@ -155,12 +177,116 @@ impl<P: PublicId> Graph<P> {
...
+                sync_parent.is_request() && self.is_awaiting_associated_event(sync_parent.index)
+            })
+    }
+
+    fn update_awaiting(&mut self, index: EventIndex) {
+        let event = &self.events[index.0];
+        if event.is_requesting() {
+            let _inserted = self.awaiting_associated_events.insert(index);
+            #[cfg(any(test, feature = "testing"))]
+            assert!(_inserted);
```
maqi: seems we expect the above `self.awaiting_associated_events.insert` always return true.
In that case, shall we put a log_or_panic ?
same comment to the below similar assertion places.

src/gossip/graph/mod.rs
```diff
@@ -68,6 +85,11 @@ impl<P: PublicId> Graph<P> {
             }
         };
 
+        if needs_updated {
```
jeanphilippeD: Why do we need this needs_updated at all?
What breaks if we call update_awaiting all the time?

Should we avoid updating if the event was already there instead?
It seems this function can be called multiple times.

We do not want to re-insert the requesting, if request has been found, afterward, but then we call that function with the requesting event again.

So it seems it should be done within the Vacant code.

src/gossip/graph/mod.rs
jeanphilippeD: do the asserts in this function reflect a bug, or an invalid input that we detect for malice? (if the latter, they probably should not be there).

src/gossip/graph/mod.rs
```diff
@@ -155,12 +177,116 @@ impl<P: PublicId> Graph<P> {
...
+            let _inserted = self.awaiting_associated_events.insert(index);
+            #[cfg(any(test, feature = "testing"))]
+            assert!(_inserted);
+        } else if event.is_request() {
+            let _inserted = self.awaiting_associated_events.insert(index);
+            #[cfg(any(test, feature = "testing"))]
+            assert!(_inserted);
+            let index_to_remove = self.other_parent(event).map(|other_parent| {
+                #[cfg(any(test, feature = "testing"))]
+                assert!(other_parent.is_requesting());
```
jeanphilippeD: I'm not sure this assert should be here, is that not more of a malice detection?

src/gossip/messages.rs
Fraser999: That wouldn't solve the example I mentioned though.

src/gossip/graph/mod.rs
```diff
@@ -155,12 +189,75 @@ impl<P: PublicId> Graph<P> {
...
+                .map(|associated_request| {
+                    #[cfg(any(test, feature = "testing"))]
+                    assert!(associated_request.is_request());
+                    associated_request.index
+                });
+            let _removed =
+                index_to_remove.map(|index| self.awaiting_associated_events.remove(&index));
+            #[cfg(any(test, feature = "testing"))]
+            assert_eq!(_removed, Some(true));
+        }
```
jeanphilippeD: So I was thinking something along these lines:

```
fn get_awaiting_and_awaited_index(&self, index: EventIndex) {
    let event = &self.events[index.0];
    if event.is_requesting() {
        (Some(index), None)
    } else if event.is_request() {
        (Some(index), self.other_parent(event))
    } else if event.is_response() {
        (None, self
                .other_parent(event)
                .and_then(|other_parent| self.self_sync_ancestor(other_parent)))
    }
}

fn update_awaiting(&mut self, index: EventIndex) { 
    let (awaiting, awaited) = get_awaiting_and_awaited_index(index);
    if let Some(awaiting) = awaiting {
         let _inserted = self.awaiting_associated_events.insert(index);
    }
    if let Some(awaited) = awaited {
        let _removed = self.awaiting_associated_events.remove(awaited);
    }
}


// Or with the new struct, something like:
fn update_awaiting(&mut self, index: EventIndex) { 
    let (awaiting, awaited) = get_awaiting_and_awaited_index(index);
    let (_inserted, _removed) = self.awaiting_associated_events.insert(awaiting, awaited);
}
```

src/gossip/graph/mod.rs
```diff
@@ -155,12 +189,75 @@ impl<P: PublicId> Graph<P> {
...
     /// Remove the topologically last event.
     pub fn remove_last(&mut self) -> Option<(EventIndex, Event<P>)> {
         let event = self.events.pop()?;
         let _ = self.indices.remove(event.hash());
+        #[cfg(feature = "malice-detection")]
+        {
+            let _ = self
+                .awaiting_associated_events
+                .remove(&EventIndex(self.events.len()));
+            if event.is_request() {
```
jeanphilippeD: could you reuse the proposed get_awaiting_and_awaited_index here?

src/dev_utils/record.rs
Fraser999: I don't think I meant to commit this block.  I'll remove it since the second assertion is invalid in the case of accusations being made.  If you'd prefer me to keep it, let me know what additional info you'd want in the assertion and I'll add it.

src/dev_utils/record.rs
Fraser999: Yes, that's what the comment means :)
The comment was just moved from the previous location, but I agree it's probably less than helpful now.  I commented the new member `added_final_requesting_event` in the definition of `Record`.  Do you think that's enough so that this comment can just be removed?  Or should I more or less duplicate the comment from there?

src/dev_utils/record.rs
Fraser999: You mean so we'd be sending a `Request` message from ourself to ourself?  I don't think that's possible since all of the events we created have been turned into corresponding actions for us to do during `play()`.

I can add an assertion that `src != contents.our_id` if you want?

src/dev_utils/record.rs
```diff
@@ -257,66 +232,128 @@ fn extract_genesis_group<'a>(
...
+            if thread::panicking() {
+                let msg = format!("!  Record constructed from '{}'  !", self.0.display());
+                let border = iter::repeat('!').take(msg.len()).collect::<String>();
+                println!("\n{1}\n{}\n{1}\n", msg, border);
             }
         }
     }
 
     /// Run smoke test using given dot file.
-    /// Use ignore_last_events to skip fake request that replay generates to send remaining gossip events.
```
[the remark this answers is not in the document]
Fraser999: I did that when I defined `ignore_last_events` just below here.  I don't think it should still be a doc comment since it doesn't really apply to the signature/args of the function any more.

src/dev_utils/record.rs
```diff
@@ -325,26 +362,37 @@ mod tests {
...
     #[test]
     fn smoke_parsec() {
-        let ignore_last_events = 0;
-        smoke("input_graphs/benches/minimal.dot", ignore_last_events)
+        let is_dot_file = |path: &PathBuf| {
+            path.extension()
+                .map(|extension| extension.to_string_lossy() == "dot")
+                .unwrap_or(false)
+        };
+        let max_file_size = 50_000;
```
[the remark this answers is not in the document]
Fraser999: It yielded a decent number of extra dot files to get tested without slowing down this test much.  I don't really know if it's worthwhile adding that though as a comment, since it just changes the question from "Why 50,000?" to "How many extra dot files is enough?" and "What's an acceptable slowdown?"  I have added a check that we at least tested one dot file, in case our syntax changes again and all dot files end up over 50kB.

src/functional_tests.rs
Fraser999: This test was failing while I was still developing the code, and during debugging I noticed the name mismatch.  I don't know which previous commit caused the mismatch.

src/gossip/graph/mod.rs
Fraser999: > Why do we need this needs_updated at all?
What breaks if we call update_awaiting all the time?

I designed `update_awaiting()` to be called after we'd confirmed that this event wasn't already in our graph.  I _could_ change that behaviour so that `update_awaiting()` can be called unconditionally, but I don't think that's an improvement.

> Should we avoid updating if the event was already there instead?
It seems this function can be called multiple times.

I'm not quite following this.  You mean if the event was already in the graph, or in `awaiting_associated_events`?

> We do not want to re-insert the requesting, if request has been found, afterward, but then we call that function with the requesting event again.

I believe that's what `needs_updated` is doing for us here?

src/gossip/graph/mod.rs
Fraser999: I don't think that works for stable Rust.  I'd need to nest the `if` in a new scope to make that compile afaik.

src/gossip/graph/mod.rs
Fraser999: You mean change this to something like:
```rust
if !self.awaiting_associated_events.insert(index) {
    log_or_panic!("");
}
```
I could I guess - I just don't see the advantage?  Also, I'm not sure all these assertions will last for too long once we start writing tests for malicious behaviour

Fraser999: They would reflect a bug; I was using them to help sanity check that `awaiting_associated_events` was being synced correctly in the absence of proper unit tests.  However, I'll remove them since they've served their purpose and depending on how we want to handle malicious entries, may be invalid anyway.

src/gossip/graph/mod.rs
Fraser999: I'll remove it.

src/dev_utils/record.rs
maqi: I am ok with you to remove this block. :)

src/dev_utils/record.rs
maqi: I think it's ok just to remove this comment.

src/dev_utils/record.rs
maqi: I thought the `events_to_gossip` contains the events in our graph that need to be sent to a peer, where some of those events are created by other peers (received not created).
Meanwhile, the `src` here needs to be `self`? which means   `events_to_gossip.last()` has to be an event created by self, which doesn't being reflected in the collection code?

src/dev_utils/record.rs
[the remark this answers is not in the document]
maqi: Ok then. It's just mandatory number is always suspicious.  :)

src/functional_tests.rs
maqi: maybe better split it into a separte commit as it's not realted to any of your change.

src/gossip/graph/mod.rs
maqi: if the assertion will be removed eventually, then I am ok with leave it as now.

src/dev_utils/record.rs
Fraser999: No, these are all incoming messages to us.

So, when constructing the `Record` we iterate the parsed dot file, and when we see e.g. a `Request` event created by us, it represents us having received a request message from the other_parent creator.  We don't prebuild any messages from ourself, we just call the running parsec's `create_gossip()` and `handle_request()` to allow it to generate the outgoing messages.

src/dev_utils/record.rs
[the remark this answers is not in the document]
Fraser999: Yeah, true.  I'll add a comment :)

src/functional_tests.rs
Fraser999: I _really_ don't want to have to do that for such a trivial change!  This is only renaming a variable.

src/gossip/graph/mod.rs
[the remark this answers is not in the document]
Fraser999: With your suggested change of `get_awaiting_and_awaited_index()` added now, do you agree it's not worth creating a wrapper for the set?

src/parsec.rs
```diff
@@ -1721,6 +1713,65 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
 
+    fn detect_invalid_request(&mut self, event: &Event<S::PublicId>) -> Result<()> {
+        if !event.is_request() {
+            return Ok(());
+        }
+
+        let is_valid = self
+            .graph
+            .other_parent(event)
+            .map(|other_parent| {
```
jeanphilippeD: I think I would call the closure parameter `requesting_event` here to match the other function.
Within the closure we care what event it represent, not where it is from.

src/parsec.rs
```diff
@@ -1721,6 +1713,65 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+            .unwrap_or(false);
+
+        if is_valid {
+            return Ok(());
+        }
+
+        let packed_event = event.pack(self.event_context())?;
+        self.create_accusation_event(
+            event.creator(),
+            Malice::InvalidRequest(Box::new(packed_event)),
```
jeanphilippeD: any reason to have 2 types `InvalidRequest` and `InvalidResponse` instead of one maybe `InvalidRequestResponse`?

We could possibly even merge `detect_invalid_response` and `detect_invalid_request` with only the `is_valid` computation being different.

src/parsec.rs
```diff
@@ -1721,6 +1713,65 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+
+    fn detect_invalid_response(&mut self, event: &Event<S::PublicId>) -> Result<()> {
+        if !event.is_response() {
+            return Ok(());
+        }
+
+        let is_valid = self
+            .graph
+            .other_parent(event)
+            .and_then(|other_parent| self.graph.self_sync_ancestor(other_parent))
```
jeanphilippeD: I think this is fine, (this may skip other events that are not accusation, which would be wrong would it not?)

I'm thinking we will probably want to move that check in unpack in the future when it is just checking other parent, like we do if we do not find any parent (i.e in unpack we probably want to ensure Requesting/Request/Response pattern, but not necessarily check that it is from correct. (maybe event only check that other_parent of an event is always Requesting/Request/Response but not check further there)).

src/parsec.rs
```diff
@@ -1721,6 +1713,65 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+            .and_then(|request_event| {
+                self.graph
+                    .other_parent(request_event)
+                    .map(|requesting_event| (request_event, requesting_event))
+            })
+            .map(|(request_event, requesting_event)| {
+                Some(request_event.creator()) == requesting_event.requesting_recipient()
```
jeanphilippeD: Why do we check this? it seem it should have been validated by `detect_invalid_request` when request was added.

src/parsec.rs
```diff
@@ -1450,19 +1454,12 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
     // of `is_request`) from `src`, then add it to our graph.
-    //
-    // `opt_other_parent` will contain the other-parent this new sync event should use, unless the
-    // gossip message from the peer was empty, in which case this will be `None` and we'll just use
-    // `src`'s most recent event we know of.
     fn create_sync_event(
         &mut self,
-        src_index: PeerIndex,
         is_request: bool,
+        other_parent: EventIndex,
```
jeanphilippeD: I think this is a good bug fix.
Ideally it would be in its own commit. did that change an existing test that would cover it? should this bug fix have its own test?

src/parsec.rs
```diff
@@ -1732,6 +1713,65 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+    fn detect_invalid_request(&mut self, event: &Event<S::PublicId>) -> Result<()> {
+        if !event.is_request() {
+            return Ok(());
+        }
+
+        let is_valid = self
+            .graph
+            .other_parent(event)
+            .map(|other_parent| {
+                Some(event.creator()) == other_parent.requesting_recipient()
```
jeanphilippeD: should we check is_requesting()?

src/parsec.rs
```diff
@@ -1732,6 +1713,65 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+            .graph
+            .other_parent(event)
+            .and_then(|other_parent| self.graph.self_sync_ancestor(other_parent))
+            .and_then(|request_event| {
+                self.graph
+                    .other_parent(request_event)
+                    .map(|requesting_event| (request_event, requesting_event))
+            })
+            .map(|(request_event, requesting_event)| {
+                Some(request_event.creator()) == requesting_event.requesting_recipient()
```
jeanphilippeD: should we check is_request()?

src/parsec.rs
Fraser999: This was to match the current paradigm of being explicit in the `Malice` variants.  I'd rather not change this for now until we're sure there's no debugging benefit from keeping them separate.

src/parsec.rs
Fraser999: Yup, now that we don't add invalid `Request`s to the graph, this shouldn't ever return false.  Good catch.

src/parsec.rs
Fraser999: This isn't worthy of its own commit - it's fine as part of this.

There aren't any unit tests for these sorts of methods, but it's called in a few of the functional tests.

src/parsec.rs
Fraser999: That's implicit via `requesting_recipient()`, which will return `None` in that case.

src/parsec.rs
Fraser999: I don't think we'll need that once we nail down enforcing the Requesting→Request→Response pattern, but I'll add it for now.

src/parsec.rs
```diff
@@ -1645,14 +1637,14 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
 
         self.detect_other_parent_by_same_creator(event)?;
         self.detect_self_parent_by_different_creator(event)?;
+        self.detect_invalid_request(event)?;
```
maqi: at the begining of `detect_invalid_request & detect_invalid_response`, we will check `is_request or is_response`.
will there be a chance of this event is an invalid requesting ?

Fraser999: I don't think so.  If the packed event was successfully unpacked, then we have a valid `Requesting` since (at least for now) there are no rules prohibiting creating it at any point.

Also, the calls to `is_request()` and `is_response()` are basically filtering all events except those from the corresponding `detect_...` methods - they're not really validating anything on their own.

src/parsec.rs
jeanphilippeD: What I meant, I guess, will a test break if we reverted that part?
If so, I'm happy as it is.

src/gossip/graph/mod.rs
jeanphilippeD: > I designed update_awaiting() to be called after we'd confirmed that this event wasn't already in our graph.

jeanphilippeD: > I designed update_awaiting() to be called after we'd confirmed that this event wasn't already in our graph. 

> Should we avoid updating if the event was already there instead?
It seems this function can be called multiple times.

What I meant to say is this is exactly what the `match` is already doing, why do we have a second mechanism instead of using the `Entry::Vacant` branch.

src/parsec.rs
Fraser999: Yes, several tests fail if we use that old method of finding the other-parent.

src/gossip/graph/mod.rs
Fraser999: Can't remember what issue the borrow checker had before when I tried this, but with the recent changes, I can indeed just call `update_awaiting()` from inside the `Vacant` branch now.  Fixup commit pushed.

PR #282, opened 2019-03-21
MAID-3296: Fix interesting event detection in presence of forks

madadam:
This PR changes the implementation of the detection of interesting event to not be based on the `sees` relation, but instead on the `is_descendant`. This makes it resilient to forks as the result of `is_descendant` never changes retroactively (unlike `sees`).

### DIFF
--- a/src/gossip/event.rs
+++ b/src/gossip/event.rs
@@ -33,36 +33,6 @@ use std::cmp;
 use std::collections::BTreeMap;
 use std::fmt::{self, Debug, Display, Formatter};
 
-/// Provide a small interface to Event not dependent on PublicId. Serves as a test seam.
-pub(crate) trait AbstractEvent {
-    /// Returns whether this event can see `other`, i.e. whether there's a directed path from `other`
-    /// to `self` in the graph, and no two events created by `other`'s creator are ancestors to
-    /// `self` (fork).
-    fn sees<E: AsRef<Self>>(&self, other: E) -> bool
-    where
-        Self: Sized;
-
-    /// The vote payload_key for an Observation event
-    fn payload_key(&self) -> Option<&ObservationKey>
-    where
-        Self: Sized;
-
-    /// The PeerIndex for this event creator
-    fn creator(&self) -> PeerIndex
-    where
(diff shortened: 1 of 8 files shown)


src/parsec.rs
```diff
@@ -873,6 +873,8 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
 
         let peers_that_can_vote = self.voters();
 
+        let is_descendant = |x, y| self.graph.is_descendant(x, y);
```
jeanphilippeD: So I assume it has not impact on performance if there are no forks.
Could you show the perf result in PR description, particularly for 65'000 events.

madadam: There should be no performance impact, because if there are not forks, `is_descendant` works the same as `sees`. I'll post the results shortly.

src/gossip/event.rs
```diff
@@ -434,28 +404,6 @@ impl<P: PublicId> Event<P> {
     }
 }
 
-impl<P: PublicId> AbstractEvent for Event<P> {
-    fn sees<E: AsRef<Self>>(&self, other: E) -> bool {
-        self.sees(other)
```
jeanphilippeD: Is there any circumstance where we should ever use sees over is_descendant?
Should we remove sees?

Or should we rename is_descendant to sees?

madadam: I had the same question and apparently we still need `sees`. It's still used in `strongly_sees`, for example.

tests/integration_tests.rs
```diff
@@ -410,8 +410,6 @@ fn extensive_dynamic_membership() {
 // execute it when the feature is disabled.
 // TODO: remove this after MAID-3270 is completed.
 #[cfg(not(feature = "malice-detection"))]
-// TODO: remove this after MAID-3164 is completed.
-#[ignore]
 #[test]
 fn consensus_with_fork() {
```
jeanphilippeD: do we need more tests, seems quite sparse?

madadam: Maybe, but that would be outside of the scope of this PR, I think.

PR #286, opened 2019-03-28
Various updates

Fraser999:
The bulk of these changes come from [a single reformatting commit](https://github.com/maidsafe/parsec/commit/7691912498ce8ef541e56d26175505be77d1d699), so it may be worth excluding that commit when reviewing to see the other changes more easily.


src/dev_utils/dot_parser.rs
```diff
@@ -784,16 +784,43 @@ pub(crate) fn parse_dot_file<P: AsRef<Path>>(full_path: P) -> io::Result<ParsedC
...
-        filename,
-        &unwrap!(thread::current().name()).replace("::", "_"),
-    )
+        const THREADS_ARG: &str = "--test-threads";
+        let mut args_itr = env::args().skip_while(|arg| !arg.starts_with(THREADS_ARG));
+        match (
+            args_itr.next().as_ref().map(|x| &**x),
+            args_itr.next().as_ref().map(|x| &**x),
+        ) {
+            (Some(THREADS_ARG), Some("1")) | (Some("--test-threads=1"), _) => {
```
madadam: I guess you didn't use the named constant in the second branch because you would have to append `"=1"` to it which would make it not constant and therefore not usable as pattern? I wonder whether passing the string directly instead of through named constant wouldn't make things cleaner in this case.

Fraser999: I did have it that way originally, but `THREADS_ARG` is used in 3 places, so I thought it was slightly the better option.  I'm ok to change it back though if you prefer?

Fraser999:
Yes, I had to use nightly rustfmt to get access to `merge_imports` - it's not recognised by the stable version unfortunately.

src/dev_utils/dot_parser.rs
madadam: Nah, it's fine. But maybe I'd extract this whole thread detection thing into a separate function for clarity.

PR #288, opened 2019-03-29
MAID-3134: Calculate coin tosses after calculating meta-votes

madadam:
This PR modifies the way in which meta-votes are assigned to events:
1. First, temporary meta-votes are calculated using no coin tosses
2. Coins are tossed using the temporary meta-votes
3. Final meta-votes are calculated using the temp meta-votes and the coin tosses

### DIFF
--- a/src/meta_voting/meta_vote.rs
+++ b/src/meta_voting/meta_vote.rs
@@ -13,7 +13,7 @@ use std::{
 };
 
 // This holds the state of a (binary) meta vote about which we're trying to achieve consensus.
-#[derive(Clone, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, Copy, Default, PartialEq, Eq, Serialize, Deserialize)]
 pub(crate) struct MetaVote {
     pub round: usize,
     pub step: Step,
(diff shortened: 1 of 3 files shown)


src/parsec.rs
```diff
@@ -1146,19 +1160,23 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-            if let Some(meta_vote) = meta_vote {
-                if meta_vote.aux_value.is_some() {
-                    let event = self.graph.get(event_index)?;
-                    return Some(event.hash().least_significant_bit());
-                }
+            if meta_vote.aux_value.is_none() {
+                continue;
             }
+
+            let event = self.graph.get(event_index)?;
```
maqi: This looks the same as in the PR 289?
same question that why switch to use `continue` explicitly?

also leave this comment to ensure the same change will be auto-resolved.

madadam:
> I'm wondering if it'd be worth making `MetaVote::next()` private and adding two new public constructors which enforce the flow a bit better? E.g.
> ...
> I don't even know if it'd be worthwhile creating a new type for these temp votes (`struct TempVote(MetaVote)`) to make it slightly safer? That's probably overkill though.

Sounds useful. I'll think about it.

src/meta_voting/meta_vote.rs
```diff
@@ -83,20 +83,41 @@ impl MetaVote {
         if is_voter {
             initial.estimates = BoolSet::from_bool(initial_estimate);
         }
-        Self::next(&[initial], others, &BTreeMap::new(), total_peers, is_voter)
+        Self::next_votes(&[initial], others, &BTreeMap::new(), total_peers, is_voter)
     }
 
-    pub fn next(
+    /// Create temporary next meta-votes. They must be finalized by calling `temp_final` before
```
maqi: `next_final` ?

PR #289, opened 2019-03-29
Fix coin value search

madadam:
1. Do not stop the search when the current event has no meta-votes. Continue on the next event instead.
2. When new meta-election is started and there are no remaining unconsensused events, set `continue_consensus_start_index` to `graph.len()` instead of `usize::MAX`, so that events added afterwards are still considered when searching for coin value.

### DIFF
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -1146,19 +1146,25 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
                 continue;
             }
 
-            let meta_votes = self.meta_election.populated_meta_votes(event_index)?;
-            let meta_vote = meta_votes
-                .get(peer_index)
+            let meta_vote = self
+                .meta_election
+                .populated_meta_votes(event_index)
+                .and_then(|meta_votes| meta_votes.get(peer_index))
                 .into_iter()
                 .flatten()
                 .find(|meta_vote| meta_vote.round_and_step() >= (round, Step::GenuineFlip));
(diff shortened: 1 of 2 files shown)


src/parsec.rs
```diff
@@ -1141,24 +1141,25 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-                if meta_vote.aux_value.is_some() {
-                    let event = self.graph.get(event_index)?;
-                    return Some(event.hash().least_significant_bit());
-                }
+            if meta_vote.aux_value.is_none() {
+                continue;
             }
+
+            let event = self.graph.get(event_index)?;
+            return Some(event.hash().least_significant_bit());
```
maqi: just wondering why switch from the old style of  
`if let Some(meta_vote) = meta_vote {
    if meta_vote.aux_value.is_some() {
            let event = self.graph.get(event_index)?;
            return Some(event.hash().least_significant_bit());
    }
} `
to use `continue` explicitly ?

madadam: No big reason, just I find is slightly more readable.

maqi: @Fraser999 : do you have any preference here?

Fraser999: No _strong_ preference, but I did wonder about the reason for the change too :)

If I'd written it, I'd probably have defaulted to what was there before (fewer LoC).  But I think there's a push to avoid nesting where possible.

madadam: If you guys feel strongly about it I can revert it.

PR #290, opened 2019-04-01
MAID-3300: Add support for single node sections

madadam:


src/dev_utils/network.rs
```diff
@@ -648,9 +652,12 @@ impl Network {
...
 
     fn allow_addition_of_peer(&self) -> bool {
+        // Only allow another addition if the currently joined peers still form supermajority
+        // after all the joining peers have joined. This is because the joining peers might not
+        // necessarily vote for adding the new peer - only the currently joined ones are guaranteed
+        // to do so.
         let joined_count = self.num_with_network_view(NetworkView::Joined);
-        // Increment by 1 to include the potential new joiner in the joining count.
-        let joining_count = self.num_with_network_view(NetworkView::Joining) + 1;
+        let joining_count = self.num_with_network_view(NetworkView::Joining);
```
maqi: It seems the newly added comment doesn't explain why we no longer require the additional plus one.
Is the removal of `plus one` required to support `single node section` ?

src/parsec.rs
```diff
@@ -670,13 +670,11 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
             return Ok(PostProcessAction::Continue);
         }
 
-        if !get_known_event(self.our_pub_id(), &self.graph, event_index)?.is_sync_event() {
-            // Only add meta events for sync events
-            return Ok(PostProcessAction::Continue);
```
maqi: hmm...  the previous flow will return immediately in case of non-sync_events, however now it will continue to other process create_blocks. Is it expected?

src/parsec.rs
```diff
@@ -1271,6 +1269,18 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
     }
 
     fn compute_consensus(&self, event_index: EventIndex) -> Vec<ObservationKey> {
+        // If the creator of the current event is the only voter so far and the event carries
```
maqi: `is the only known voter` ?

src/parsec.rs
```diff
@@ -1271,6 +1269,18 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+        // a payload, decide it immediately.
+        if let Ok(event) = self.get_known_event(event_index) {
+            if let Some(payload_key) = event.payload_key() {
+                let voter_indices = self.peer_list.voters().map(|(peer_index, _)| peer_index);
+                if iter::once(event.creator()).eq(voter_indices) {
+                    return vec![*payload_key];
+                }
+            }
+        }
+
```
Fraser999: I wonder if we should be doing something like `if self.voter_count() <= 1 { return vec![]; }` to keep the previous level of optimisation?

src/dev_utils/network.rs
madadam: > It seems the newly added comment doesn't explain why we no longer require the additional plus one.

It does not explain the change itself, but I think it explains why the function works the way it does. But maybe it can be worded better? 

> Is the removal of plus one required to support single node section ?

Yes. The previous implementation would not allow sections of size 1 or 2 to grow.

Fraser999: I think that `+1` was required though, since if this method returns `true`, we'll add another joining peer, and that one might not get voted for by enough peers.

madadam: My assumption is that all the `joined` ones will vote for the addition and the `joining` ones may or might not. So in the worst case all the `joining` ones will become `joined` before the vote to add the new one is consensused, but even then there is still supermajority of the ones that voted and so the new peer gets added. Am I wrong?

src/parsec.rs
madadam: > it will continue to other processes such as create_blocks...

It won't, because the meta-event is not created and so [this line](https://github.com/maidsafe/parsec/pull/290/files#diff-98b4a6ae0c53dc7ababf2d21eb815263R1284) will return `None` which in turn makes `compute_consensus` return an empty `vec` and `process_event` take [this branch](https://github.com/maidsafe/parsec/pull/290/files#diff-98b4a6ae0c53dc7ababf2d21eb815263R680).

src/dev_utils/network.rs
Fraser999: I'm trying to remember the details myself.  The [commit which added it](https://github.com/Fraser999/parsec/commit/d1a727ec87639377ae43221e51847b20a3baf4df) points to the issue being related to adding and dropping peers.

madadam: That commit restricts addition of peers where previously there was no such restriction. My change only slightly relaxes the restriction but I think it should still be correct.

Fraser999: It might be down to one of the joined ones being killed before it votes?  I can't recall the details, but I guess soak testing should show us before too long.

src/parsec.rs
```diff
@@ -1271,6 +1269,19 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
     }
 
     fn compute_consensus(&self, event_index: EventIndex) -> Vec<ObservationKey> {
+        // If the creator of the current event is the only known voter and the event carries
+        // a payload, decide it immediately.
+        if let Ok(event) = self.get_known_event(event_index) {
+            if let Some(payload_key) = event.payload_key() {
+                if iter::once(event.creator()).eq(self.voters()) {
```
maqi: what's the diff between `self.peer_list.voters()` and `self.voters()` ?

src/dev_utils/network.rs
madadam: OK, you might be right. I just got a test failure (in the `add_many_peers_and_vote` test). I'm thinking of reverting it but adding a special case for networks of size 1 and 2.

src/parsec.rs
madadam: I think the difference used to be more significant in the past but now `self.voters()` is pretty much just a cache of `self.peer_list.voters().map(|(peer_index, _)| peer_index)`

src/parsec.rs
```diff
@@ -1271,6 +1269,19 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+                    return vec![*payload_key];
+                } else if self.voters().len() <= 1 {
```
Fraser999: I was thinking this would need to be outside the `if let Some(payload_key) = event.payload_key()` block?  If we're inside that `if`, then won't we already be returning in the `None` arm of `self.meta_election.populated_meta_votes` below since there are only meta votes on sync events, not those with payloads?

PR #292, opened 2019-04-03
MAID-3301: Fix failures in the consensus_with_fork test when malice-detection is enabled

madadam:
This PR fixes issues that caused the `consensus_with_fork` test to fail:

1. Update the test framework to account for raised accusations when calculating expected number of blocks.
2. Modify parsec so that it accepts events even from removed peers if those peers haven't yet reached the consensus on their own removal.

### DIFF
--- a/src/dev_utils/schedule.rs
+++ b/src/dev_utils/schedule.rs
@@ -465,6 +465,16 @@ impl ObservationSchedule {
             })
             .count()
     }
+
+    fn count_expected_accusations(&self) -> usize {
+        if cfg!(feature = "malice-detection") {
+            // One accusation per malicious peer as currently the malicious peers commit only one
+            // malice each.
+            self.genesis.ids_of_malicious_peers.len()
+        } else {
+            0
+        }
+    }
 }
 
 pub struct StepObservationSchedule {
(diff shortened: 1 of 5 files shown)


src/dev_utils/schedule.rs
```diff
@@ -465,6 +465,18 @@ impl ObservationSchedule {
             })
             .count()
     }
+
+    #[cfg(feature = "malice-detection")]
+    fn count_expected_accusations(&self) -> usize {
```
Fraser999: This could be handled without conditional compilation by doing:
```rust
if cfg!(feature = "malice-detection") {
    // One accusation per malicious peer as currently the malicious peers commit only one
    // malice each.
    self.genesis.ids_of_malicious_peers.len()
} else {
    0
}
```

madadam: That's still conditional compilation the way I see it :wink:  I don't mind making the change though.

Fraser999: I know what you mean - dunno if I'm using proper Rust terminology.  But both arms of the `if` statement are at least checked by the compiler unconditionally (I'd expect whichever is false to get optimised away eventually by the compiler), so I don't consider it conditional compilation.

src/peer_list/peer.rs
```diff
@@ -92,6 +102,12 @@ impl<P: PublicId> Peer<P> {
     }
 }
 
+#[derive(Debug)]
+enum Presence {
+    Present(PeerState),
+    Removed(EventIndex),
```
pierrechevalier83: I would put a comment here to indicate that this event index isn't necessarily the index of the event at which the guy was removed in their pov, but the index of the event in which I learnt they were removed. (Just to avoid possible confusion later)

PR #293, opened 2019-04-04
Make dot file dumper & parser supports malice accusation

maqi:
This PR contains the work of making dot file dumper & parser supports malice accusation.
So far, there are only two malice supported: Fork and InvalidAccusation. 
As the testing case that is currently being worked on only requires these two.

To support other malice types, the main blocking point is how a `packed_event` to be dumped and parsed. This will require a further discussion.

### DIFF
--- a/src/dev_utils/dot_parser.rs
+++ b/src/dev_utils/dot_parser.rs
@@ -18,8 +18,8 @@ use crate::{
     },
     mock::{PeerId, Transaction},
     observation::{
-        ConsensusMode, Observation, ObservationHash, ObservationInfo, ObservationKey,
-        ObservationStore,
+        ConsensusMode, Malice, MaliceInput, Observation, ObservationHash, ObservationInfo,
+        ObservationKey, ObservationStore,
     },
     peer_list::{PeerIndex, PeerIndexMap, PeerIndexSet, PeerList, PeerState},
     round_hash::RoundHash,
@@ -356,8 +356,10 @@ fn parse_cause() -> Parser<u8, CauseInput> {
     let response = seq(b"Response").map(|_| CauseInput::Response);
     let observation =
         (seq(b"Observation(") * parse_observation() - sym(b')')).map(CauseInput::Observation);
+    let malice = (seq(b"Observation(") * parse_accusation() - sym(b')'))
+        .map(|(offender, malice_input)| CauseInput::Malice(offender, malice_input));
 
-    prefix * (initial | requesting | request | response | observation) - newline()
+    prefix * (initial | requesting | request | response | observation | malice) - newline()
 }
 
 fn parse_last_ancestors() -> Parser<u8, BTreeMap<PeerId, usize>> {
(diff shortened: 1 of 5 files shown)


src/observation.rs
```diff
@@ -115,6 +115,29 @@ pub enum Malice<T: NetworkEvent, P: PublicId> {
 }
 
 impl<T: NetworkEvent, P: PublicId> Malice<T, P> {
+    #[cfg(feature = "dump-graphs")]
```
jeanphilippeD: I think this need to move to the dump_graph file to follow the same pattern started for dump_graph, where it own the responsibility of outputting things the way it needs.

It can be a free function like for observation.

src/dev_utils/dot_parser.rs
```diff
@@ -1077,8 +1106,11 @@ fn create_events(
             next_event_details.last_ancestors.clone(),
             peer_list,
             &mut parsed_contents.observations,
+            &name_hash_map,
         );
 
+        let _ = name_hash_map.insert(ev_id.clone(), *next_event.hash());
```
jeanphilippeD: name_hash_map somewhat confusing as this is not a HashMap...

src/dev_utils/dot_parser.rs
```diff
@@ -1077,8 +1106,11 @@ fn create_events(
             next_event_details.last_ancestors.clone(),
             peer_list,
             &mut parsed_contents.observations,
+            &name_hash_map,
```
jeanphilippeD: Could we pass malice info directly here containing the correct hash.
Or maybe pass a different Cause that contain the hash for the malice (so as to avoid adding a new parameter)?

jeanphilippeD: > what we want to know is not the hash of the malice, but the hash of the event that malice is accusing against.

Yes, and we compute it somewhere down the call stack.
Should we compute it here and pass the Cause with the info we need instead of in 2 parts Cause + map?

src/dump_graph.rs
```diff
@@ -1070,6 +1072,31 @@ mod detail {
         }
     }
 
+    fn write_malice_to_string<T: NetworkEvent, P: PublicId>(
+        malice: &Malice<T, P>,
+        graph: &Graph<P>,
+    ) -> String {
+        match malice {
+            Malice::Fork(event_hash) => {
+                let short_name = unwrap!(graph.get_by_hash(event_hash)).short_name();
```
jeanphilippeD: Sorry, did not read properly: Could you use the short name facilities provided by dump graph, not the event one that does not work with many nodes, and that we should probably remove.

src/dump_graph.rs
```diff
@@ -1070,6 +1072,31 @@ mod detail {
...
+            }
+            Malice::Accomplice(event_hash, boxed_malice) => {
+                let short_name = unwrap!(graph.get_by_hash(event_hash)).short_name();
+                format!(
+                    "Accomplice({}, {})",
+                    short_name,
+                    write_malice_to_string(boxed_malice, graph),
+                )
+            }
+            _ => panic!("unsupported yet"),
```
jeanphilippeD: I would list them explicitly so we get compilation error if we add a new one and need to decide whether we want to dump them.

src/dump_graph.rs
maqi: you mean the facility of `event_to_short_name`? 
It belongs to DotWriter, which has DotObserver (where the write_malice_to_file happens) as its member.  :(
I think there is issue of the ownership. And even I managed it, it requires additional parameters to be passed around. :(

src/dump_graph.rs
maqi: That will be a lot to list out. And I think a panic is obvious enough?

src/dump_graph.rs
jeanphilippeD: Yes, but the functionalities are in `short_peer_ids`.
Though I see forks were added and so only need peer_list.

So you could replace that member with a new struct that contains `short_peer_ids` and `peer_list` and provide that functionality.

Using the short_name function is just not correct.

src/dump_graph.rs
```diff
@@ -1072,31 +1056,82 @@ mod detail {
...
-            }
-            Malice::Accomplice(event_hash, boxed_malice) => {
-                let short_name = unwrap!(graph.get_by_hash(event_hash)).short_name();
-                format!(
-                    "Accomplice({}, {})",
-                    short_name,
-                    write_malice_to_string(boxed_malice, graph),
+            Malice::Fork(event_hash) => format!(
+                "Fork({})",
+                sanitise_event_short_name(
```
jeanphilippeD: maybe add in this function opening:

```
let get_short_name_by_hash = |event_hash| {
                sanitise_event_short_name(
                    unwrap!(graph.get_by_hash(event_hash)),
                    peer_list,
                    short_peer_ids
                )
};

And then use it in the different branch of the match to avoid duplication.

File: src/gossip/event.rs [hunk 7]
```diff
@@ -422,7 +424,7 @@ impl<P: PublicId> Event<P> {
     }
 
     /// Returns the first char of the creator's ID, followed by an underscore and the event's index.
-    #[cfg(any(test, feature = "testing"))]
+    #[cfg(any(test, feature = "testing", feature = "dump-graphs"))]
```
jeanphilippeD: You should not need that anymore, why do we need it?
Same for the others

PR #294, opened 2019-04-05
Create accusations before the sync event

Fraser999:
This changes the order in which accusations of malice are added to the graph.  Previously we would buffer these as we handled an incoming message, then add them all after we'd created our sync event.

This PR changes that process to add them all just before we create our sync event.

The only thing in this PR which I'd specifically draw attention to is the removal of the functional test `accomplice_negative()`.  I didn't see any value in it, but I'm maybe missing something.


src/dev_utils/network.rs
```diff
@@ -455,7 +455,7 @@ impl Network {
             .unpolled_accusations()
             .find(|(offender, malice)| match malice {
                 Malice::Fork(..) => !self.peer(offender).has_misbehaved(),
-                _ => malice.is_provable(),
+                _ => true,
```
maqi: why previously we diff provable/unprovable ?

src/functional_tests.rs
```diff
@@ -658,122 +659,120 @@ mod handle_malice {
...
-        let mut dave = initialise_parsec(dave_id.clone(), &genesis, None);
-        assert!(!dave.graph().contains(&a_0_hash));
-        assert!(!dave.graph().contains(&a_1_hash));
-
-        // Send gossip from Alice to Dave.
-        let message = unwrap!(alice.create_gossip(&dave_id));
-        unwrap!(dave.handle_request(&alice_id, message));
-        assert!(dave.graph().contains(&a_0_hash));
-        assert!(dave.graph().contains(&a_1_hash));
+    fn initialise_alice_bob_and_carol() -> (TestPeer, TestPeer, TestPeer) {
```
maqi: do we really need these initialisers in this format?
can we do something like `fn initialise_instances(num: usize) -> vec![TestPeer]` ?

src/functional_tests.rs
```diff
@@ -784,23 +783,18 @@ mod handle_malice {
...
-        ));
+        let duplicated_payload = Observation::OpaquePayload(Transaction::new("ABCD"));
+        let first_duplicate = unwrap!(carol
+            .new_event_from_observation(carol.our_last_event_index(), duplicated_payload.clone()));
 
-        let first_duplicate_clone = unwrap!(carol.new_event_from_observation(
-            carol.our_last_event_index(),
-            Observation::OpaquePayload(Transaction::new("ABCD")),
-        ));
+        let first_duplicate_clone = unwrap!(carol
```
maqi: can't clone from `first_duplicate` directly ?

src/dev_utils/network.rs
Fraser999: I don't know, and I couldn't see a need for it, so I made the test stricter.

src/functional_tests.rs
Fraser999: That's what `initialise_genesis_parsecs()` does, but these helpers just reduce the amount of duplicated boilerplate needed in the tests.  I think the tests are more readable if we assign names like "Alice", "Bob", etc. to the `TestPeer`s rather than just using e.g. `peers[0]`.

src/functional_tests.rs
Fraser999: Correct - `Event` doesn't implement `Clone`.

madadam: Is there any reason to not have `Clone` for `Event` ?

Fraser999: I think it's probably good that we don't; makes it impossible to accidentally clone when we should be using the `IndexedEventRef` paradigm.

src/gossip/event_utils.rs
```diff
@@ -61,13 +65,25 @@ impl IndexSet {
     }
 }
 
+impl Debug for IndexSet {
+    fn fmt(&self, formatter: &mut Formatter) -> fmt::Result {
+        let values = self.0.iter().collect::<BTreeSet<_>>();
+        for (index, value) in values.iter().enumerate() {
```
jeanphilippeD: https://docs.rs/itertools/0.8.0/itertools/fn.join.html ?

madadam: Or https://docs.rs/itertools/0.8.0/itertools/trait.Itertools.html#method.format.

src/parsec.rs
```diff
@@ -1659,6 +1638,77 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
 
 #[cfg(feature = "malice-detection")]
 impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
+    fn create_accusation_events(&mut self, other_parent: EventIndex) -> Result<()> {
+        let pending_accusations = mem::replace(&mut self.pending_accusations, vec![]);
+        for (offender, malice) in pending_accusations {
+            let _ = self
```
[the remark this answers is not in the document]
Fraser999: That's not doing the same thing though is it?  `create_accusation_event()` just returns a single optional accusation.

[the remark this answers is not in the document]
pierrechevalier83:
```
if let Some(unhandled) =  self.create_accusation_event(offender, malice, other_parent)? {
    self.pending_accusations.push((offender, unhandled));
}
```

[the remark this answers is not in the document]
pierrechevalier83: I agree with using `.map()` over nesting  in general, but the fact you have to discard the return value here is a hint in my eyes that it is not the appropriate construct for this particular situation.

PR #295, opened 2019-04-08
MAID-3290: add tests for accusations of forks

Fraser999:


src/parsec.rs
```diff
@@ -2054,7 +2054,16 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
             .all(|accused_event| event.is_descendant_of(accused_event))
     }
 
-    fn accused_event_is_fork_but_not_seen_by(
+    fn events_with_self_parent(
```
madadam: I think this function is either incorrectly implemented or misnamed. Consider:

```
A2,0     A2,1
|        |
A1,0     A1,1
|        |
+--------+
|
A0
```
When called on `A1,0` it returns `[A2,0, A2,1]` but `A1,0` is not self-parent of `A2,1`.

src/parsec.rs
```diff
@@ -2067,7 +2076,14 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
                         .map(|event| (accused_event, event))
                 })
-                .map(|(accused_event, event)| !event.sees_fork_by(accused_event.creator()))
+                .map(|(accused_event, event)| {
+                    // `event` must be a descendant of more than one fork branch to be in a position
+                    // to know about the fork.
+                    self.events_with_self_parent(accused_event)
+                        .filter(|forked_event| event.is_descendant_of(forked_event))
+                        .count()
+                        < 2
```
madadam: Wouldn't this give incorrect result in situation like this:

```

+------------------------B4 
|                        |
|       +----------------B3
|       |                |
A2,0    A2,1      A2,2   |
|       |         |      .
|       +----+----+      .
|            |           .
A1,0         A1,1
|            |
+------------+
|
A0
```
when `accused_event` is `A1,1` and `event_index` is `B4` ?

src/functional_tests.rs
```diff
@@ -975,17 +975,152 @@ mod handle_malice {
...
+        let mut response = unwrap!(carol.handle_request(&bob_id, request));
+        assert!(carol.graph().contains(&a_2_0.compute_hash()));
+        let expected_malice = Malice::Fork(a_1.compute_hash());
+        let expected_accusation = vec![(&alice_id, &expected_malice)];
+        assert_peer_has_accused(&carol, expected_accusation.clone());
+        unwrap!(bob.handle_response(&carol_id, response));
+        assert!(bob.graph().contains(&a_2_1.compute_hash()));
+        assert_peer_has_accused(&bob, expected_accusation.clone());
+
+        // Send a request from Dave to Carol and send the response.  Dave should accuse A_1.
```
maqi: Actually, here we are also confirming Carol only accused Alice once and only once.
Shall we mention this in the comment?

src/functional_tests.rs
```diff
@@ -1062,6 +1062,85 @@ mod handle_malice {
         assert_peer_has_accused(&dave, both_accusations.clone());
     }
 
+    #[test]
+    //    A_3,0    A_3,1     A_3,2
```
maqi: the testing code seems doesn't have the event of `A_3_2` ?
does it really not required?

PR #297, opened 2019-04-11
chore/parsec: update API poll() method

Fraser999:
This reverts `Parsec::poll()` to yield a single block at a time rather than in batches.  This is a more convenient API for Routing to use at the moment as it avoids extra state being held in the case where some blocks from a polled batch need to be handled differently.

The previous behaviour of `poll()` is provided as a testing-only method named `batch_poll()`.  This is used by tests to allow rigorous validation of the signatories of any given block.  Since all block groups are handled identically by all peers, we can definitively know which peers should and shouldn't have signed each block.


src/parsec.rs
```diff
@@ -366,7 +366,29 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
     /// Once the owning peer has been removed from the section (i.e. a block with payload
     /// `Observation::Remove(our_id)` has been made stable), then no further blocks will be
     /// enqueued. So, once `poll()` returns such a block, it will continue to return `None` forever.
-    pub fn poll(&mut self) -> Option<BlockGroup<T, S::PublicId>> {
+    pub fn poll(&mut self) -> Option<Block<T, S::PublicId>> {
+        let mut block_group = self.batch_poll()?;
+
+        if cfg!(any(test, feature = "testing")) && block_group.is_empty() {
+            log_or_panic!("self.consensused_blocks.front() is empty");
```
madadam: This can be replaced by simple `panic`, although I wonder why we need this at all?

Fraser999: Fair point about the `panic`, but don't you feel it's worth not just silently swallowing what would be a logic error otherwise?

madadam: It's pretty harmless error - we would end up returning `None` in case the block group was empty. I don't mind keeping the check if you think it's worth it.

Fraser999: Yeah, true, albeit we're breaking the documented behaviour.  I think you've convinced me though - I'll just remove it.

PR #299, opened 2019-04-13
chore/all: fix new clippy warnings

Fraser999:


src/dev_utils/misc.rs
```diff
@@ -26,7 +26,9 @@ pub trait TestIterator: Iterator {
         item
     }
 
-    fn is_sorted(mut self) -> bool
+    // TODO: Remove this method once the equivalent one in the std lib is stable.
+    // https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.is_sorted
+    fn itr_is_sorted(mut self) -> bool
```
madadam: What is the point of this rename?

Fraser999: The new compiler started warning about a name conflict here with a soon-to-be-stable std method.

madadam: I see. Thought that will be the case. :+1:

PR #300, opened 2019-04-15
MAID-3299 test new Requesting/Request/Response pattern

maqi:


src/functional_tests.rs
```diff
@@ -761,6 +761,302 @@ mod handle_malice {
...
+            bob.our_pub_id().clone(),
+            b_1.compute_hash(),
+            alice_requesting_hash,
+        );
+
+        let request = Request {
+            packed_events: vec![b_0.clone(), b_1.clone(), invalid_req.clone()],
+        };
+        assert_matches!(
+            alice.handle_request(bob.our_pub_id(), request),
```
jeanphilippeD: This does not seem right, the other_parent is not the second alice event.

```
Bob construct invalid packed Request with Alice's 2nd last event as other parent (ensure it's not a Requesting(Bob)).
Call Alice unpack_and_add_events(vec![Bob's Request]) - should fail.
```

src/functional_tests.rs
```diff
@@ -761,6 +761,302 @@ mod handle_malice {
...
+        let a_0 = unwrap!(nth_event(bob.graph(), 2).pack(bob.event_context()));
+        let a_1 = unwrap!(nth_event(bob.graph(), 3).pack(bob.event_context()));
+        let alice_requesting = unwrap!(nth_event(bob.graph(), 4).pack(bob.event_context()));
+        let invalid_req = PackedEvent::new_request(
+            bob.our_pub_id().clone(),
+            b_1.compute_hash(),
+            alice_requesting_hash,
+        );
+        let request = Request {
+            packed_events: vec![b_0, b_1, a_0, a_1, alice_requesting, invalid_req],
```
jeanphilippeD: Are these events not suppose to be in reverse order?

src/functional_tests.rs
```diff
@@ -761,6 +761,302 @@ mod handle_malice {
...
+        let request = Request {
+            packed_events: vec![b_0.clone(), b_1.clone(), invalid_req.clone()],
+        };
+        assert_matches!(
+            alice.handle_request(bob.our_pub_id(), request),
+            Err(Error::InvalidMessage)
+        );
+
+        let request = Request {
+            packed_events: vec![b_0, b_1, a_0, a_1, alice_requesting, invalid_req],
```
jeanphilippeD: Not clear that it matches the description.
```
Call Carol unpack_and_add_events(vec![Alice's Requesting, Bob's Request]) - should fail.
```

src/functional_tests.rs
```diff
@@ -761,6 +761,302 @@ mod handle_malice {
...
+        let request = unwrap!(alice.create_gossip(bob.our_pub_id()));
+        let response = unwrap!(bob.handle_request(alice.our_pub_id(), request));
+
+        // If the response be sent to Carol, a response event shall not be created.
+        assert_matches!(
+            carol.handle_response(bob.our_pub_id(), response),
+            Err(Error::UnknownOtherParent)
+        );
+        // The Carol's graph only contains Bob and Carol's inital and gensis events.
+        assert_eq!(carol.graph().len(), 4);
```
jeanphilippeD: not super clear, maybe better to check against a list of event.
Also, why 4? is there not Carol, bob initial event and genesis events?

src/functional_tests.rs
```diff
@@ -761,6 +761,302 @@ mod handle_malice {
...
+
+        // Create invalid response from Carol.
+        let c_0 = unwrap!(nth_event(carol.graph(), 0).pack(carol.event_context()));
+        let c_1 = unwrap!(nth_event(carol.graph(), 1).pack(carol.event_context()));
```
jeanphilippeD: These would be a lot clearer if we had an helper nth_peer_event(carol).
Then it would be obvious the naming is correct, and we do not assign bob's event to c_1 for example.

src/functional_tests.rs
```diff
@@ -761,6 +761,302 @@ mod handle_malice {
...
+        };
+
+        assert_matches!(
+            bob.handle_response(carol.our_pub_id(), response),
+            Err(Error::InvalidEvent)
+        );
+
+        let b_0 = unwrap!(nth_event(bob.graph(), 0).pack(bob.event_context()));
+        let b_1 = unwrap!(nth_event(bob.graph(), 1).pack(bob.event_context()));
+        let a_0 = unwrap!(nth_event(bob.graph(), 2).pack(bob.event_context()));
```
jeanphilippeD: same about non obvious names even more so => a_0 is 2, is not really clear

src/functional_tests.rs
```diff
@@ -761,6 +761,302 @@ mod handle_malice {
...
+        };
+        assert_matches!(
+            carol.handle_request(bob.our_pub_id(), request),
+            Err(Error::InvalidMessage)
+        );
+    }
+
+    #[test]
+    fn invalid_response_wrong_recipient() {
+        let (mut alice, mut bob, mut carol, mut dave) =
```
jeanphilippeD: having 3 tests instead of 1 may make each test easier to read.
Would also need to have more specific names which would be helpful.

src/functional_tests.rs
```diff
@@ -761,6 +761,302 @@ mod handle_malice {
...
+        let request = unwrap!(alice.create_gossip(carol.our_pub_id()));
+        let alice_requesting_hash = *nth_event(alice.graph(), 2).hash();
+
+        // Bob shall not create response for that request, but can put Alice's requesting event into
+        // its graph.
+        assert_matches!(
+            bob.handle_request(alice.our_pub_id(), request),
+            Err(Error::InvalidMessage)
+        );
+        assert!(bob.graph().contains(&alice_requesting_hash));
```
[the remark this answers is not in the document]
maqi: I was following the steps in the task description. 
At least will be easier for the reviewers to checking whether there is anything missed?

src/functional_tests.rs
maqi: No.

src/functional_tests.rs
maqi: Carol has to built up the knowledge of Alice and Bob first. 
Otherwise the self_parent & other_parent info will be missing.

src/functional_tests.rs
maqi: There is: Carol's initial and gensis, then Bob's inital and gensis.
Checking against a list of events is a good idea.

src/functional_tests.rs
jeanphilippeD: funnily I think they are the same, should it be a_1?

src/functional_tests.rs
maqi: no sure what you mean.
Isn't `nth_event` did the job?

src/functional_tests.rs
maqi: in Bob's graph, the second is Event(A_0).
any better naming suggestion?

src/functional_tests.rs
maqi: As explained above, I was just following the task description, hoping easier for checking.

src/functional_tests.rs
jeanphilippeD: Would it not be Alice, Bob and Carol inital and gensis events?
Bob added Alices to be ancestor of the request event he created?

src/functional_tests.rs
maqi: Ah, right. Yeah, it shall be `A_1`, not `alice_requesting`

src/functional_tests.rs
jeanphilippeD: nth did the job, as in we were able to get the event we wanted, but it is not clear.
Especially more so for the other example:

```
// Has to understand what 2 is:
let a_0 = unwrap!(nth_event(bob.graph(), 2).pack(bob.event_context()));
// Now can use 0 in name and function call:
let a_0 = unwrap!(nth_peer_event(bob, alice.id(), 0).pack(bob.event_context()));
```

src/functional_tests.rs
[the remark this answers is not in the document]
jeanphilippeD: I think this is not necessarily incompatible: could be 2 tests that keep the link with task like:

fn invalid_request_wrong_recipient_to_requesting()
fn invalid_request_wrong_recipient_to_other_node()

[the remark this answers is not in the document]
maqi: then probably there are too many duplicated setups of the testing environemnt :(
I think the soak test won't like this, as the execution time is also sensitive.

src/functional_tests.rs
maqi: No. The response from bob doesn't contain the events of Alice.

src/functional_tests.rs
```diff
@@ -761,6 +761,302 @@ mod handle_malice {
...
+        assert_matches!(
+            alice.handle_request(bob.our_pub_id(), request.clone()),
+            Err(Error::InvalidEvent)
+        );
+
+        assert_matches!(
+            dave.handle_request(bob.our_pub_id(), request),
+            Err(Error::InvalidEvent)
+        );
+    }
```
Fraser999: I think this part of the task is missing from the tests:
> When failures happen in tests below, check accusations are raised with the appropriate invalid event packed in them, and that the invalid events aren't added to graphs.

src/functional_tests.rs
```diff
@@ -761,6 +761,302 @@ mod handle_malice {
...
+
+        // Bob shall not create response for that request, but can put Alice's requesting event into
+        // its graph.
+        assert_matches!(
+            bob.handle_request(alice.our_pub_id(), request),
+            Err(Error::InvalidMessage)
+        );
+        assert!(bob.graph().contains(&alice_requesting_hash));
+
+        // Create invalid request from Bob.
```
Fraser999: Maybe good to expand a little.  E.g.
```rust
// Create invalid request from Bob.  (Bob should not use Alice's `Requesting(Carol)` event
// as an other-parent for his `Request` event).
```
Or maybe this would be more obvious if `alice_requesting` was named `alice_requesting_to_carol`?

src/functional_tests.rs
```diff
@@ -761,6 +761,302 @@ mod handle_malice {
...
+        let (mut alice, mut bob, mut carol) =
+            unwrap!(initialise_genesis_parsecs(3).into_iter().collect_tuple());
+
+        // Send request from Alice to Bob.
+        let request = unwrap!(alice.create_gossip(bob.our_pub_id()));
+        let alice_requesting_hash = *nth_event(alice.graph(), 2).hash();
+
+        assert!(bob.handle_request(alice.our_pub_id(), request).is_ok());
+
+        // Create invalid request from Bob.
```
Fraser999: Again, more details would help.

src/functional_tests.rs
Fraser999: Also "ensure it's not a Requesting(Bob)" to me means there should be assertion in the code.  Partly it future-proofs the test (should the sequences of events change in the future) and partly it helps document what we're testing for.

src/functional_tests.rs
```diff
@@ -761,6 +761,302 @@ mod handle_malice {
...
+        let b_0 = unwrap!(nth_event(bob.graph(), 0).pack(bob.event_context()));
+        let b_1 = unwrap!(nth_event(bob.graph(), 1).pack(bob.event_context()));
+        let a_0 = unwrap!(nth_event(bob.graph(), 2).pack(bob.event_context()));
+        let a_1 = unwrap!(nth_event(bob.graph(), 3).pack(bob.event_context()));
+        let alice_requesting = unwrap!(nth_event(bob.graph(), 4).pack(bob.event_context()));
+        let invalid_req = PackedEvent::new_request(
```
Fraser999: This would actually be a valid `Request` event.  The only reason the test is passing is that we call `alice.handle_request()` below; we should be calling `alice.handle_response()`.  If we passed that current message (as a response message) to alice, it would be accepted and the test would fail.

src/functional_tests.rs
```diff
@@ -761,6 +761,302 @@ mod handle_malice {
...
+        );
+
+        let request = Request {
+            packed_events: vec![b_0.clone(), b_1.clone(), invalid_req.clone()],
+        };
+        assert_matches!(
+            alice.handle_request(bob.our_pub_id(), request),
+            Err(Error::InvalidMessage)
+        );
+
```
Fraser999: If we called `assert!(!alice.graph().contains(&invalid_req_hash));` as per the task description, that would fail here.  It's an invalid message because it should be a response.  To be a valid request message, it must finish with a `Requesting` event.

src/functional_tests.rs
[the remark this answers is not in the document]
jeanphilippeD: I do not understand this, these test are very small, are you saying each test have a very big cost?

src/functional_tests.rs
jeanphilippeD: Why? It seems it should.

src/functional_tests.rs
```diff
@@ -761,6 +761,302 @@ mod handle_malice {
...
+        let request = Request {
+            packed_events: vec![b_0, b_1, a_0, a_1, alice_requesting, invalid_req],
+        };
+        assert_matches!(
+            carol.handle_request(bob.our_pub_id(), request),
+            Err(Error::InvalidMessage)
+        );
```
Fraser999: Likewise, if we called `assert!(!carol.graph().contains(&invalid_req_hash));` - that would fail too.  Carol will have added all the events from the message, but then returned `InvalidMessage` since it doesn't finish in a `Requesting(Carol)` event created by Bob.

src/functional_tests.rs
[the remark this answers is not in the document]
Fraser999: I'm not saying they have a big cost (not sure if that's what Qi's saying though?)

What I'm saying is that the the description in the task makes sense to me as a single functional test.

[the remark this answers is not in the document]
jeanphilippeD: sorry I was replying to Qi, I did not see your answer then,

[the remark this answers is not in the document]
jeanphilippeD: I'm ok if you are both happier with this format.

src/functional_tests.rs
maqi: I think expanding the comment makes more sense.

src/functional_tests.rs
maqi: Ah, yeah, I shall add them in.

src/functional_tests.rs
maqi: The task description says
"3, Call Alice unpack_and_add_events(vec![Bob's Request]) - should fail."
You mean we shall actually use `alice.handle_response` to accept this, and still expect handle_response return with an error?
or you mean if we do so, handle_response returns ok, which indicates a bug?

Fraser999: I think it should be fine to use `handle_response()` for this purpose here - it should return an error if the response message does actually contain an invalid `Request` event created by Bob.  (The one currently being created is valid, but if you make the change as JP suggested it won't be.)  That invalid event shouldn't get added to anyone's graph, and accusations should be made.

However, all this is true if we just call `unpack_and_add_events()` too.  As long as we're checking for invalid events not being added to graphs and being added to appropriate accusations, I don't think there's much difference between either option.

src/functional_tests.rs
maqi: the task description says `Call Carol unpack_and_add_events`, however the pub functions we have access are `handle_response` and `handle_request`, so sometime really confusing with which one to use. 
Do you mean we need to compose a `requesting` event using `new_requesting` function at this step instead?

src/functional_tests.rs
maqi: same question as in https://github.com/maidsafe/parsec/pull/300/files#r275776119

src/functional_tests.rs
maqi: Bob received A_0 & A_1 just from Alice, hence doesn't include them in the response msg to Alice for the optimisation purpose.

src/functional_tests.rs
Fraser999: I very strongly agree that having message types named the same as events is particularly confusing.  But in the functional tests, we're using `TestParsec`s which exists to expose private methods and helpers for testing purposes, so we could just add wrapper for `unpack_and_add_events()` there if you feel that would simplify things?

> Do you mean we need to compose a requesting event using new_requesting function at this step instead?

We shouldn't need to here, since we expect Alice to stop iterating through the packed events in the message when she hits Bob's invalid one.  It doesn't matter whether we add more valid packed events after that one - she shouldn't process them.

src/functional_tests.rs
maqi: I think here we are more interested in the info of its position in the graph, not the topological position of the event creator.
actually, I could just use another set of names, say b_0, b_1, b_2 to mapping there graph index (I used current naming set just to help me for debugging).
I think  the function of `nth_peer_event` is a bit of overshot. And we may still need to keep `nth_event` for some usage.

src/functional_tests.rs
maqi: first, I'd prefer not to open another function as public.
second, I'd like this set of tests to simulate the normal usage (i.e. use handle_response/handle_request only), to give an example how requestting-request-response pattern works.

Calling `unpack_and_add_events()` will hide the pattern a lot to me.

Fraser999: I'm fine with either approach, as long as sticking to just the existing public API of `TestParsec` doesn't introduce a lot of extra code to the tests.  But I doubt that'll be the case here; so go with whatever you prefer :)

src/functional_tests.rs
maqi: > we could just add a wrapper

as mentioned in https://github.com/maidsafe/parsec/pull/300#discussion_r275781779
I'd use handle_request/handle_response to show how the pattern works, instead of creating an extra wrapper, which could to be further confusing.

src/functional_tests.rs
jeanphilippeD: > (I used current naming set just to help me for debugging).

jeanphilippeD: > (I used current naming set just to help me for debugging).

Then it suggest these names are documenting and useful.

> I think the function of nth_peer_event is a bit of overshot. And we may still need to keep nth_event for some usage.

I do not suggest removing nth_event, just offering multiple options so the test can use the most appropriate.

src/functional_tests.rs
jeanphilippeD: a_0 is fine, the problem is that a_0 is at index 2 of bob.

maqi: I could take it from Alice's graph, if it helps. :)

jeanphilippeD: That is another option. I think that would also solve that particular concern (if it is ok to build message from multiple peers).

src/functional_tests.rs
```diff
@@ -761,6 +774,347 @@ mod handle_malice {
...
+        // its graph.
+        assert_matches!(
+            bob.handle_request(alice.our_pub_id(), request),
+            Err(Error::InvalidMessage)
+        );
+        assert!(bob.graph().contains(&alice_requesting_hash));
+
+        // Create invalid request from Bob.  (Bob should not use Alice's `Requesting(Carol)` event
+        // as an other-parent for his `Request` event).
+        let alice_requesting_hash = *nth_event(alice.graph(), 2).hash();
```
jeanphilippeD: Does that not set again alice_requesting_hash to its same value (set above)?

src/functional_tests.rs
```diff
@@ -761,6 +774,347 @@ mod handle_malice {
...
+    }
+
+    #[test]
+    fn invalid_request_wrong_recipient() {
+        let (mut alice, mut bob, carol, mut dave) =
+            unwrap!(initialise_genesis_parsecs(4).into_iter().collect_tuple());
+
+        // Create request from Alice to Carol, but sent to Bob.
+        let request = unwrap!(alice.create_gossip(carol.our_pub_id()));
+        let alice_requesting_hash = *nth_event(alice.graph(), 2).hash();
```
jeanphilippeD: Maybe something for the future, but it would be good to have small wrapper functions that hide implementation detail to help the important documenting part of the test pop out.

Ideally we would write in this main test function:
```
let request = alice.create_request_for(carol);
// Or easier to write: let request = create_request_from_for(alice, carol);
let alice_requesting_hash = nth_event_hash(alice, 2);

...

let invalid_req = pack_new_request(bob, b_1_hash, alice_requesting_hash);
```

src/functional_tests.rs
```diff
@@ -761,6 +774,347 @@ mod handle_malice {
...
+            Some(invalid_request),
+            None,
+            Error::InvalidEvent,
+            Some(&expected_malice),
+            Some(&invalid_req_hash),
+        );
+
+        packed_events = take_packed_events(&bob, 5);
+        packed_events.push(invalid_req);
+        let invalid_request = Request { packed_events };
```
jeanphilippeD: Could you describe in a very short sentence (1 line) this request as a comment.

src/functional_tests.rs
```diff
@@ -761,6 +774,347 @@ mod handle_malice {
...
+        // Send request from Alice to Bob.
+        let request = unwrap!(alice.create_gossip(bob.our_pub_id()));
+        let alice_requesting_hash = *nth_event(alice.graph(), 2).hash();
+
+        assert!(bob.handle_request(alice.our_pub_id(), request).is_ok());
+
+        // Create invalid request from Bob.  (Bob should not use Alice's `Requesting(Carol)` event
+        // as an other-parent for his `Request` event).
+        let b_1_hash = *nth_event(bob.graph(), 1).hash();
+        let valid_req = unwrap!(nth_event(bob.graph(), 5).pack(bob.event_context()));
```
jeanphilippeD: Helper:
```
let valid_req = nth_packed_event(&bob, 5);
```

src/functional_tests.rs
```diff
@@ -761,6 +774,347 @@ mod handle_malice {
...
+        );
+        // The Carol's graph only contains Bob and Carol's inital and gensis events.
+        assert_eq!(carol.graph().len(), 4);
+
+        // Create invalid response from Carol.  (Carol should not use Bob's `Request(Alice)` event
+        // as an other-parent for his `Response` event).
+        let c_1_hash = *nth_event(carol.graph(), 1).hash();
+        let bob_request = unwrap!(nth_event(bob.graph(), 5).pack(bob.event_context()));
+        let bob_request_hash = bob_request.compute_hash();
+        let invalid_resp =
```
jeanphilippeD: helper:
```
let invalid_resp = pack_new_response(&carol, c_1_hash, bob_request_hash);
```

src/functional_tests.rs
```diff
@@ -518,10 +518,23 @@ mod handle_malice {
...
 
+    fn take_packed_events<T: NetworkEvent, S: SecretId>(
+        peer: &TestParsec<T, S>,
+        n: usize,
+    ) -> Vec<PackedEvent<T, S::PublicId>> {
+        let mut packed_events = Vec::new();
+        for i in 0..n {
+            packed_events.push(unwrap!(
+                nth_event(peer.graph(), i).pack(peer.event_context())
+            ));
```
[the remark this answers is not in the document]
maqi: how to limit the number of entries? we don't want all the entries sometime :(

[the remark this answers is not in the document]
jeanphilippeD: This is what `take` function does: https://doc.rust-lang.org/nightly/core/iter/trait.Iterator.html#method.take

These implementations unless I'm mistaken are functionally equivalent.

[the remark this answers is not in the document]
maqi: ah, right, missed that `take`. thx

src/functional_tests.rs
maqi: a copy/paste error  :(

src/functional_tests.rs
maqi: I am not sure about the `create_request_for` and `nth_event_hash`, however `pack_new_request` seems a good suggestion.

I will switch to use  `pack_new_request` in the next fixup commit.

src/functional_tests.rs
maqi: knowledge of Alice and Bob and plus the invalid_req.

jeanphilippeD: Yes, that will be a good comment to add.

src/functional_tests.rs
jeanphilippeD: The reason, is having to say .graph and .our_pub_id really had noise to the test case, so we decipher it more than we read it. In these long test case, any noise removal is a plus.

src/functional_tests.rs
```diff
@@ -761,6 +774,347 @@ mod handle_malice {
...
+        );
+
+        packed_events = take_packed_events(&bob, 5);
+        packed_events.push(invalid_req);
+        let invalid_request = Request { packed_events };
+
+        assert_handling_invalid_req_or_resp(
+            &mut bob,
+            &mut carol,
+            Some(invalid_request),
```
Fraser999: If the `packed_events` finish with a `Request` event, we should be passing it as a response message, not a request.

src/functional_tests.rs
```diff
@@ -761,6 +774,347 @@ mod handle_malice {
...
+        // Send request from Alice to Bob.
+        let request = unwrap!(alice.create_gossip(bob.our_pub_id()));
+        let response = unwrap!(bob.handle_request(alice.our_pub_id(), request));
+
+        // If the response be sent to Carol, a response event shall not be created.
+        assert_matches!(
+            carol.handle_response(bob.our_pub_id(), response),
+            Err(Error::UnknownOtherParent)
+        );
+        // The Carol's graph only contains Bob and Carol's inital and gensis events.
```
Fraser999: Typos: "initial" and "genesis"

src/functional_tests.rs
maqi: This is only being used in two places. I'd prefer not to add an extra helper so far.
unless there is more places to use it.

jeanphilippeD: I would propose the same suggestion even if there was only 1 usage.

It is not how many time it is used.
It is how readable it is in the context. Because many things happen in this test, to keep it readable, we express things with as little noise as possible so it read well.

src/functional_tests.rs
```diff
@@ -761,6 +774,347 @@ mod handle_malice {
         assert!(carol.pending_accusations().is_empty());
     }
 
+    fn assert_handling_invalid_req_or_resp(
+        sender: &mut TestPeer,
+        receiver: &mut TestPeer,
+        invalid_req: Option<Request<Transaction, PeerId>>,
+        invalid_resp: Option<Response<Transaction, PeerId>>,
+        _expected_error: Error,
```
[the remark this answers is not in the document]
maqi: Probably we can make this change as a separate task & PR ?
As I think there will be some other places can be updated as well?

src/functional_tests.rs
```diff
@@ -761,6 +774,347 @@ mod handle_malice {
...
+        expected_malice: Option<&Malice<Transaction, PeerId>>,
+        invalid_hash: Option<&EventHash>,
+    ) {
+        assert!(invalid_req.is_some() || invalid_resp.is_some());
```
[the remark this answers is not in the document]
maqi: That's the original pattern I tried.
However, as we are using `assert_matches` inside, there will be complaining of unused request / response.
I could add another line of `assert!(!(invalid_req.is_none() && invalid_resp.is_none()));` to enusre there is only one passed down ?

src/functional_tests.rs
Fraser999: I agree with @maqi here in that a helper which is so small and used so infrequently is more of a hindrance than a helper.

src/functional_tests.rs
maqi: this is to simulate the step of `Call Carol unpack_and_add_events(vec![Alice's Requesting, Bob's Request]) - should fail` , which says the Request to be the last.
Also, `unpack_and_add_events` will be called immediately inside `handle_request` or `handle_response`, which I don't see it will cause difference at this stage?

src/functional_tests.rs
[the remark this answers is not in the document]
Fraser999: Would you propose to just match explicitly then in this function rather than use the `assert_matches!` and do a separate PR which updates that as well?

It seems like it's not much work to bundle this change (even as a separate commit if you prefer) with this PR, but as long as the we're not using that macro in here, and we have a task to remove it altogether as a priority, then I'd be ok with that.

src/functional_tests.rs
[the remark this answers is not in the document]
Fraser999: I see that as more justification for replacing the two optionals with a single enum.

src/functional_tests.rs
jeanphilippeD: Sub function are not to only be used for code de-duplication, so how many times it is used is not very relevant at all.
There is also a level of abstraction consideration here, this test function should not focus on the implementation of how to get a pack event.

When I read, I know what it intend to do.
`let valid_req = nth_packed_event(&bob, 5);`

When I read:
`let valid_req = unwrap!(nth_event(bob.graph(), 5).pack(bob.event_context()));`

I have to parse different parts and ignore the noise (the important bit being `nth_event, bob, 5 and pack`). At this point I've lost the focus of what the test does, to go pass how it does it.

jeanphilippeD: But I made that point before so I let you decide.

src/functional_tests.rs
Fraser999: Yes, but if you prefer to not call `unpack_and_add_events()` and instead use the top-level `handle_request/response()` then I think you should pass messages which are only as malformed as is needed for the test.

We do currently get all the way through `unpack_and_add_events()` inside `handle_request()` and then fail when we go to create our own sync event, since we find that our `Request` would have their `Request` as an other-parent.  If we add some check for a malformed message into the start of `handle_request()`, this test might still pass, but for the wrong reason - we wouldn't even get as far as handling the invalid event in `unpack_and_add_events()` - we'd already have returned an error and not called `unpack_and_add_events()` at all.

This wouldn't be such an issue if we used `unpack_and_add_events()` directly.  But if you still prefer to stick to the top-level methods, then I think they should be getting called with messages which only have that one invalid event, but are otherwise valid.

src/functional_tests.rs
Fraser999: Yes, at this level I feel it's personal preference.  I would normally avoid writing a helper which only wraps one line of code, even more so if it's not used often.

src/functional_tests.rs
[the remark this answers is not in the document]
maqi: I will `derive PartialEq for Error, Request and Response, and replace calls to assert_matches with assert_eq` in this PR.
And leave the work of removing that macro totally as a separate task/PR

src/functional_tests.rs
maqi: Ok, I get what you mean. I will change to use the proper `handle_response`.

src/functional_tests.rs
jeanphilippeD: OK, make sense now. Can you say that in the comment above the assert as it does not seem obvious.
i.e something like `Bob's response was for Alice and so do not include Alice's events.'

maqi: Ok, I will put the comment in

src/functional_tests.rs
Fraser999: I was mildly against this suggestion before I saw the change, but not enough to comment.  Having seen the change now, I do feel in this specific case we've made the code harder to read.  The main reason is because of the horrible duplication of the names, where a request can mean an event or a message.  By making this change, it's not immediately clear whether the helper is returning an event or a message.

I'm not asking for this helper to be changed again - it's not a big deal, but I feel we could have done more here to make it clearer (via variable names and function names) whether requests and responses are events or messages.

src/functional_tests.rs
```diff
@@ -761,6 +772,317 @@ mod handle_malice {
...
+    }
+
+    #[test]
+    fn duplicate_requests() {
+        let (mut alice, mut bob, mut carol) =
+            unwrap!(initialise_genesis_parsecs(3).into_iter().collect_tuple());
+
+        // Send request from Alice to Bob.
+        let request_msg = unwrap!(alice.create_gossip(bob.our_pub_id()));
+        let alice_requesting_hash = *nth_event(alice.graph(), 2).hash();
```
Fraser999: It might be a bit more robust, and clearer to the intent if we get this hash from Bob's valid `Request`:

```rust
let b_last = nth_event(bob.graph(), bob.graph().len() - 1);
let alice_requesting = unwrap!(bob.graph().get(unwrap!(b_last.other_parent())));
let invalid_req = pack_new_request(&bob, *b_last.hash(), *alice_requesting.hash());
```

src/functional_tests.rs
```diff
@@ -761,6 +772,317 @@ mod handle_malice {
...
+        assert!(alice
+            .handle_response(bob.our_pub_id(), response_msg)
+            .is_ok());
+
+        // Create invalid response from Alice.  (Alice should not use Bob's `Request(Alice)` event
+        // as an other-parent for his `Response` event)
+        let valid_resp =
+            unwrap!(nth_event(alice.graph(), alice.graph().len() - 1).pack(alice.event_context()));
+        let a_last_event_hash = valid_resp.compute_hash();
+        let bob_request = unwrap!(nth_event(bob.graph(), 5).pack(bob.event_context()));
```
Fraser999: I think it may be clearer and more robust to get this event hash from `valid_resp`:

```rust
let bob_request_hash = unwrap!(valid_resp.other_parent());
```

src/functional_tests.rs
```diff
@@ -761,6 +772,314 @@ mod handle_malice {
...
+
+    #[test]
+    fn invalid_request_wrong_type_of_other_parent() {
+        let (mut alice, mut bob, mut carol) =
+            unwrap!(initialise_genesis_parsecs(3).into_iter().collect_tuple());
+
+        // Send request from Alice to Bob.
+        let request = unwrap!(alice.create_gossip(bob.our_pub_id()));
+        let a_1_hash = *nth_event(alice.graph(), 1).hash();
+        assert_ne!(
```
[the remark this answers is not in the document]
jeanphilippeD: If an assert needs a comment, maybe better to put it in the assert output is possible
assert!(..., "why this is important");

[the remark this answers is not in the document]
Fraser999: Yes, that would do the job too.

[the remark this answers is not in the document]
maqi: This assertion is added due to the comment of https://github.com/maidsafe/parsec/pull/300#discussion_r275748918
Shall I put it as a printout in the assertion, or you prefer another way of assert?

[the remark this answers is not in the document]
Fraser999: I think the best option is to have `assert!(a_1 != Requesting(Bob), "This should not be a 'Requesting(Bob)' event to ensure a 'Request' using this as its other-parent is invalid.")`

src/functional_tests.rs
```diff
@@ -761,6 +772,314 @@ mod handle_malice {
...
+        let (mut alice, mut bob, mut carol, mut dave) =
+            unwrap!(initialise_genesis_parsecs(4).into_iter().collect_tuple());
+
+        // Send request from Alice to Bob.
+        let request = unwrap!(alice.create_gossip(bob.our_pub_id()));
+        let response = unwrap!(bob.handle_request(alice.our_pub_id(), request));
+
+        // If the response be sent to Carol, a response event shall not be created.
+        assert_eq!(
+            carol.handle_response(bob.our_pub_id(), response),
```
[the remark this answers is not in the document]
maqi: Ah, yes, it shall contain A_0 and A_1, I will add them in.

src/functional_tests.rs
```diff
@@ -761,6 +772,317 @@ mod handle_malice {
...
+            unwrap!(initialise_genesis_parsecs(3).into_iter().collect_tuple());
+
+        // Send request from Alice to Bob.
+        let request_msg = unwrap!(alice.create_gossip(bob.our_pub_id()));
+        assert!(bob.handle_request(alice.our_pub_id(), request_msg).is_ok());
+
+        // Create invalid response from Alice.  (Alice should not use Bob's second last event as an
+        // an other-parent for his `Response` event)
+        let a_last_hash = *nth_event(alice.graph(), alice.graph().len() - 1).hash();
+        let b_second_last =
```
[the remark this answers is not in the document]
maqi: Ah, yes, I got it in a wrong way, will swap they two.

src/functional_tests.rs
```diff
@@ -761,6 +772,317 @@ mod handle_malice {
...
+
+        // Send request from Alice to Bob.
+        let request_msg = unwrap!(alice.create_gossip(bob.our_pub_id()));
+        let response_msg = unwrap!(bob.handle_request(alice.our_pub_id(), request_msg));
+
+        assert!(alice
+            .handle_response(bob.our_pub_id(), response_msg)
+            .is_ok());
+
+        // Create invalid response from Alice.  (Alice should not use Bob's `Request(Alice)` event
```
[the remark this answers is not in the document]
maqi: I will reword the comment

src/functional_tests.rs
```diff
@@ -761,6 +772,333 @@ mod handle_malice {
...
+    fn duplicate_requests() {
+        let (mut alice, mut bob, mut carol) =
+            unwrap!(initialise_genesis_parsecs(3).into_iter().collect_tuple());
+
+        // Send request from Alice to Bob.
+        let request_msg = unwrap!(alice.create_gossip(bob.our_pub_id()));
+        let mut response_msg = unwrap!(bob.handle_request(alice.our_pub_id(), request_msg));
+
+        // Have Bob create an invalid Request event.  (Bob should not use Alice's
+        // `Requesting(Carol)` event as an other-parent for his Request event.)
```
Fraser999: This still has the copy-paste error.  It should be similar to your wording in the response test which is "Alice has already assigned a Response to the Request, hence can not use it again."

src/functional_tests.rs
```diff
@@ -761,6 +772,333 @@ mod handle_malice {
...
+            packed_events: packed_events
+                .iter()
+                .chain(response_msg.packed_events.iter())
+                .cloned()
+                .collect(),
+        };
+        // If the response be sent to Carol, a response event shall not be created.
+        assert_eq!(
+            carol.handle_response(bob.our_pub_id(), response_to_carol),
+            Err(Error::UnknownOtherParent)
```
Fraser999: This is still showing that we're not hitting the point of the test.  Bob gave Carol a malformed message - one which didn't contain enough events, not just one which has a `Request` to a different peer as its final event.

I think we should change this to something like:

```rust
// Send request from Alice to Bob.
let request_msg = unwrap!(alice.create_gossip(bob.our_pub_id()));
assert!(bob.handle_request(alice.our_pub_id(), request_msg).is_ok());

// If the response be sent to Carol, a response event shall not be created.
let packed_events = take_packed_events(&bob, bob.graph().len());
assert_eq!(
    carol.handle_response(bob.our_pub_id(), Response { packed_events }),
    Err(Error::InvalidMessage)
);
```

src/functional_tests.rs
```diff
@@ -761,6 +772,333 @@ mod handle_malice {
...
+        );
+        // The Carol's graph shall then only contain Alice, Bob and Carol's initial and genesis
+        // events.
+        assert_eq!(carol.graph().len(), 6);
```
Fraser999: I think rather than just checking the length, it'd be safer to check that Carol's graph doesn't contain a `Response` event created by her; or if it's easier, just check her last event in the graph isn't that.

PR #301, opened 2019-04-19
chore/test: remove macro

Fraser999:
This removes a macro used by the functional tests which could easily be misused, an example being highlighted by [this comment thread](https://github.com/maidsafe/parsec/pull/300#discussion_r276644857).

### DIFF
--- a/src/functional_tests.rs
+++ b/src/functional_tests.rs
@@ -22,15 +22,6 @@ use std::collections::BTreeSet;
 
 type TestPeer = TestParsec<Transaction, PeerId>;
 
-macro_rules! assert_matches {
-    ($actual:expr, $expected:pat) => {
-        match $actual {
-            $expected => (),
-            ref unexpected => panic!("{:?} does not match {}", unexpected, stringify!($expected)),
-        }
-    };
-}
-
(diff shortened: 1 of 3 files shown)


src/functional_tests.rs
```diff
@@ -293,9 +284,12 @@ fn remove_peer() {
     );
 
     // Try calling `create_gossip()` for Eric shall result in error.
-    assert_matches!(
+    assert_eq!(
         alice.create_gossip(&eric_id),
-        Err(Error::InvalidPeerState { .. })
```
maqi: I am not sure, however this looks like a possible reason we'd better still keep the `assert_matches` macro.
I mean, If we know the exact Error, then we shall use assert_eq; if we only know a general one, maybe assert_match is more suitable ?

Fraser999: The only way I'd be in favour of keeping the macro would be if it were updated to disallow a misuse.  I don't know how to do that easily, and there's only a single place where it'd be useful now (in `our_unpolled_observations_with_consensus_mode_single()`).

jeanphilippeD: It seem to make sense to me to have the full content in the assert.

PR #302, opened 2019-04-25
fix/meta-election: decide if others already did

pierrechevalier83:
Most of the logic for picking up a decision from other nodes is already
in-place:
In `MetaVoteCounts::new`, we pick any decision from any other voter and
apply it to ourselves.
In `MetaVotes::calculate_new_decision`, we use this decision from the
counts as the default value for a meta_vote.

Where the logic stopped working was in the multi-stage calculation of
the estimates:
* We first evaluate a temporary value for the meta-election that
  doesn't take into account the coin toss
* We then evaluate the coin toss value
* We then finalise the meta-election evaluation using the coin toss

Prior to this change, we wouldn't pass the other meta-votes when
finalising the meta-election evaluation. This would mean that
MetaVoteCounts::new wouldn't actually pick up on the decision that other
voter made anymore. This would mean that the decision reached in our
temp meta-vote would actually be discarded and we wouldn't shortcut the
decision process.

This bug caused issues in soak testings of the routing crate with the
real Parsec.

### DIFF
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -41,7 +41,7 @@ use itertools::Itertools;
 #[cfg(any(test, feature = "testing"))]
 use std::ops::{Deref, DerefMut};
 use std::{
-    collections::{BTreeMap, BTreeSet, VecDeque},
+    collections::{BTreeMap, BTreeSet, HashMap, VecDeque},
     iter,
     marker::PhantomData,
     mem, usize,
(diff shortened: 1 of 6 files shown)


src/parsec.rs
```diff
@@ -1093,9 +1093,15 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
             };
 
             for (peer_index, temp_votes) in &context.temp_votes {
+                let other_votes = Self::peer_meta_votes(&ancestors_meta_votes, peer_index);
```
madadam: We now call `peer_meta_votes` twice per each `peer_index` (first time on [line 1082](https://github.com/maidsafe/parsec/pull/302/files#diff-98b4a6ae0c53dc7ababf2d21eb815263R1082)). Not sure how expensive that call is, but it might be worth optimizing.

pierrechevalier83: I thought of this and thought it was a premature optimization since I didn't know if it was a bottleneck or not, so probably not. I'll benchmark it, though: just for piece of mind :)

pierrechevalier83: BTW, I've benchmarked it and it seems to actually be a big pessimisation. (varying between 25% and 200% worst with malice-detection enabled). I'll clean it up. Will also add some testing.

pierrechevalier83:
Deprecated by PR 303

PR #303, opened 2019-04-29
WIP fix/meta-election: decide if others already did

maqi:


src/parsec.rs
```diff
@@ -1374,16 +1375,43 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+                    }
+                }
+            }
+        }
+
+        if undecided_peers.len() == 1 {
+            let factor = if true_votes == false_votes { 4 } else { 2 };
+            if undecided_votes > factor * max_votes {
+                if let Some(undecided_peer) = undecided_peers.iter().last() {
+                    let _ = decided_meta_votes.insert((*undecided_peer, true_votes >= false_votes));
```
pierrechevalier83: If you're doing it right, there can't be true_votes and false_votes at the same time. Also, it is not necessary to wait for `factor * max_votes`... Also, I think it makes more sense to fix this in `meta_votes` rather than patch it up on the calling site in `parsec` which is already a bit of a God-struct...

maqi: >  there can't be true_votes and false_votes at the same time

pierrechevalier83: But then, maybe it would make sense to use PR 302 as a starting point and improve from there ;)

maqi: Well, It's just this PR get the tests passing, and seems to be a work can be completed within a predictable time.
Meanwhile, PR 302 still failing that test, and could be a long shot. :(

src/parsec.rs
```diff
@@ -1092,12 +1092,51 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+                    .map(|vote| vote.decision.is_some())
+                    .unwrap_or_else(|| false)
+                {
+                    let _ = decided.insert(peer_index, final_meta_votes);
+                } else {
+                    let _ = undecided.insert(peer_index, final_meta_votes);
+                }
+            }
+
+            if undecided.len() == 1
```
pierrechevalier83: This seems pretty arbitrary to me:
if anyone reaches a decision in the same meta-election as us, we should decide. Not only if we're the only node who haven't smelled the coffee yet...

BTW, I'm not 100% sure all these votes refer to the same election, so that would be a bug if they didn't.

src/parsec.rs
```diff
@@ -1092,12 +1092,51 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
                 temp_votes,
             };
 
+            let mut decided = BTreeMap::new();
```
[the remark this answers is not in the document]
maqi: we also need the collected votes for the next step, where you can see I used `chain` to take the advange of the container.
If only use a single bool, we still need an extra container to store the votes.

src/parsec.rs
maqi: >if anyone reaches a decision in the same meta-election as us, we should decide. Not only if we're the only node who haven't smelled the coffee yet
hmm... I don't get what you mean. You mean `we should decide when ANY one reaches a decision`? doesn't this mean `one vote takes all` ?

madadam: I'm also not convinced this is right. Why only do this when **all but one** peers decided? However, @pierrechevalier83 's idea of taking a decision when **at least one** peer decided also didn't seem right to me (my gut reaction is this opens us up for some kind of attack, but I don't have a proof for it). Maybe we can take the decision when **at least one third** decided (that means at least one honest)? But then the problem is what if not everybody decided the same way? We can have some peers deciding `true` and others `false`. What should we take then?

src/parsec.rs
```diff
@@ -1092,12 +1092,51 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+                    let _ = undecided.insert(peer_index, final_meta_votes);
+                }
+            }
+
+            if undecided.len() == 1
+                && decided.len() == voters.len() - 1
+                && undecided
+                    .values()
+                    .last()
+                    .map(|votes| votes.len() > 3)
```
madadam: Don't understand this condition. Where does the number `3` come from?

src/parsec.rs
```diff
@@ -1092,12 +1092,51 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+                    .unwrap_or_else(|| false)
```
madadam: This can be written more simply as `unwrap_or(false)`.

src/parsec.rs
```diff
@@ -1092,12 +1092,51 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+            {
+                let decision = decided
+                    .values()
+                    .last()
+                    .map(|votes| votes.iter().last().map(|vote| vote.decision))
+                    .unwrap_or_else(|| None);
```
madadam: This can be simplified to:
```rust
let decision = decided
    .values()
    .last()
    .and_then(|votes| votes.last()) // `votes` is `Vec`, so it already  has `last`. No need to do `.iter().last()`
    .and_then(|vote| vote.decision);
```
It would also make the type of `decision` cleaner: `Option<bool>` instead of `Option<Option<bool>>`.

src/parsec.rs
maqi: To avoid follow others' decision too quick. `>3` means we at least need to complete one round (3 steps)

src/parsec.rs
```diff
@@ -1092,12 +1092,51 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+                for (peer_index, votes) in undecided.iter() {
```
madadam: If you do `for ... in mem::replace(&mut undecided, BTreeMap::new())` then you wouldn't need to `clone` the `votes` below.

src/parsec.rs
maqi: > Why only do this when all but one peers decided

src/parsec.rs
madadam: Why do we need at least one round though? I think this needs a comment with explanation.

maqi: there will be failing test without this :(

src/parsec.rs
madadam: OK, but then issue with not everyone deciding the same way still holds. Currently we are seemingly arbitrarily taking the decision of the last peer and it isn't clear to me why. Also, the "last peer" depends on the ordering by `PeerIndex` which is arbitrary and not even consistent across different nodes.

maqi: I agree it's bit arbitraily.
However, it's not taking the decision of the `last peer`, but `peers' last vote`. Once a peer's vote reached decision, votes vector won't grow anymore, so take `last` shall be safe.

madadam: No, [here](https://github.com/maidsafe/parsec/pull/303/files#diff-98b4a6ae0c53dc7ababf2d21eb815263R1123) you are taking the `last` entry of `decided` and because `decided` is `BTreeMap`, `last` means last according to however the keys are ordered. As the key is `PeerIndex` in this case, the ordering is arbitrary and not consistent.

maqi: Ah, yes.
I will note this down. 
It maybe ok, as all peers shall have the same decision.
Or we shall have a count across all the decided peers.

madadam: > all peers shall have the same decision

I don't think this is necessarily the case, but maybe we can add additional condition: copy the decision only if everybody who decided, decided the same way?

maqi: According to the parsec invariants, the decisions of all peers shall be consistent, i.e. decided the same way.
We can confirm this with Pierre tomorrow, once he is back.

madadam: They are consistent in the sense that for any particular peer, every other peer will see them reaching the same decision. But decisions of different peers might still be different.

maqi:
PR is closed as the issued is turned to be in the concrete coin, and be resolved by the PR 310

PR #306, opened 2019-05-08
Change has_unconsensused_observations to has_unpolled_observations

madadam:
This PR replaces `has_unconsensused_observations` with `has_unpolled_observations` which returns `true` also for observation that have already been consensused but their corresponding blocks have not yet been polled.


src/parsec.rs
```diff
@@ -403,11 +403,11 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
             .unwrap_or(false)
     }
 
-    /// Check if there are any observations that have been voted for but not yet consensused - i.e.
-    /// if there is a gossip event containing a vote for a payload that is not yet a part of a
-    /// stable block.
-    pub fn has_unconsensused_observations(&self) -> bool {
+    /// Check if there are any observations that have been voted for but not yet polled - that is,
+    /// a block containing that observation hasn't yet been retrieved by calling `poll`.
```
maqi: I think this comments doesn't cover the scope of this function well.
A `polling` call will only drain `consensused_blocks`, however the `unconcensused observastions` will only be `drained` once consensused. This comment give the impression that a polling call of a block containing any observation will drain it event it's not concensused.

madadam: Got you. I'll try to reword it.

PR #307, opened 2019-05-10
Improve MetaVoteCounts + add unit tests

madadam:


src/meta_voting/meta_vote_counts.rs
```diff
@@ -46,52 +53,207 @@ impl MetaVoteCounts {
...
 
-            if counts.decision.is_none() {
-                counts.decision = vote.decision;
+            if decision.is_none() {
+                decision = vote.decision;
             }
         }
-        counts
+
+        Self {
```
maqi: why not construct a mut counts at the beginning as previous? so we can avoid using those temp variables of est/bin/aux/dec .

src/parsec.rs
```diff
@@ -1063,6 +1065,14 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         }
 
         let voters = self.voters();
+        let voters_len = match NonZeroUsize::new(voters.len()) {
+            Some(num) => num,
+            None => {
+                // No voters
+                return Ok(());
```
maqi: better have a log_or_panic here?

src/meta_voting/meta_vote_counts.rs
```diff
@@ -46,52 +53,207 @@ impl MetaVoteCounts {
...
+            aux_value: Some(true),
+            decision: None,
+        };
+        let vote1 = MetaVote {
+            round: 4,
+            step: Step::ForcedFalse,
+            estimates: BoolSet::Single(true),
+            bin_values: BoolSet::Single(true),
+            aux_value: Some(true),
+            decision: None,
```
maqi: regarding the testing dataset, better have `BoolSet::Both` involved, and having a decision to be not None?
maybe worth to have another test of `only_votes_with_the_same_round_and_step_as_parent_not_counted` to contains votes of having `BoolSet::Both` and a decision to be not None ?

madadam: The purpose of this test is only to verify that only votes on the same round/step as the parent vote are counted. It doesn't matter what the actual values are, only that the counts would be different if the test assumption was violated. The other test `smoke` is more about how different values affect the counts. And it already involves `BoolSet::Both` and non-`None` `decision`. 

That said, I can expand the tests to make them more thorough and explicit.

maqi: The `BoolSet::Both` will affects the counts on both true and false, meanwhile a `some` decision will be collected by the count as well. So I think they shall be reflected.
However, I think you don't need to expand the existing tests.
Just having a new test, with the parent vote as a non-voter, and having the other_votes' dataset contains `BoolSet::Both` and `Some(decision)`, shall be enough to cover the scope.

PR #309, opened 2019-05-15
Add failing test based on test from routing

madadam:


madadam:
This turned out to not be as useful as we hoped. Closing.

PR #311, opened 2019-05-29
Add related_info to Observation::Genesis

madadam:
- Add a new `genesis_info` parameter to `Parsec::from_genesis` for specifying the related_info on the genesis event.
- Some dot files had to be modified/regenerated because the hash of `Observation::Genesis` changed.
- A bogus clippy lint silenced by slightly reshuffling the code (clippy issue: https://github.com/rust-lang/rust-clippy/issues/4133)

### DIFF
--- a/input_graphs/dev_utils_record_tests_smoke_other_peer_names/annie.dot
+++ b/input_graphs/dev_utils_record_tests_smoke_other_peer_names/annie.dot
@@ -5,6 +5,7 @@
 ///   Carol: PeerState(VOTE|SEND|RECV)
 ///   Claire: PeerState(VOTE|SEND|RECV)
 /// }
+/// consensus_mode: Supermajority
 digraph GossipGraph {
   splines=false
   rankdir=BT
@@ -1163,7 +1164,7 @@ digraph GossipGraph {
 
 /// ===== meta-elections =====
 /// consensus_history:
-/// ece3eaf86a7f7396aed2e4bdd0c67cf7b67df5aabbd5eced04924b154d0f3841
+/// 5d1bc22c2cce50873bffc1ba00ecbecce52085dc3bd15e8b43f5167bf72962cd
 
 /// interesting_events: {
 ///   Annie -> ["An_19"]
@@ -1174,10 +1175,6 @@ digraph GossipGraph {
 /// all_voters: {Annie, B, Carol, Claire}
 /// unconsensused_events: {"An_13", "B_9", "Ca_6", "Cl_10"}
 /// meta_events: {
-///   An_13 -> {
-///     observees: {}
-///     interesting_content: []
-///   }
 ///   An_14 -> {
 ///     observees: {}
 ///     interesting_content: []
@@ -1353,10 +1350,6 @@ digraph GossipGraph {
 ///     observees: {}
 ///     interesting_content: []
 ///   }
-///   B_9 -> {
-///     observees: {}
-///     interesting_content: []
-///   }
 ///   B_10 -> {
 ///     observees: {}
 ///     interesting_content: []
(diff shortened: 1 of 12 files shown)


src/parsec.rs
```diff
@@ -1966,9 +1971,9 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         };
 
         let starting_index = self.peer_list.accomplice_event_checkpoint_by(creator);
-        for (_, malice) in
-            self.detect_accomplice_for_our_accusations(event_index, starting_index)?
-        {
+        let accusations =
```
madadam: Note to reviewers: this change is there only to silence a clippy false-positive, which should be fixed in the next clippy version (https://github.com/rust-lang/rust-clippy/issues/4133).

input_graphs/dev_utils_record_tests_smoke_other_peer_names/annie.dot
```diff
@@ -1219,126 +1216,126 @@ digraph GossipGraph {
 ///     interesting_content: []
 ///     meta_votes: {
 ///         stage est bin aux dec
-///       An: 0/0   f   -   -   -
-///       B: 0/0   t   -   -   -
-///       Ca: 0/0   t   -   -   -
-///       Cl: 0/0   t   -   -   -
+///       An: 0/0   f   -   -   -
```
maqi: why there is extra space come up?
I'd prefer not to have this change to reduce the amount of diff

src/observation.rs
```diff
@@ -27,7 +27,12 @@ use std::{
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 pub enum Observation<T: NetworkEvent, P: PublicId> {
     /// Genesis group
-    Genesis(BTreeSet<P>),
+    Genesis {
+        /// Members of the genesis group.
+        group: BTreeSet<P>,
+        /// Extra arbitrary information for use by the client.
```
maqi: Seems this `Extra arbitrary information` is actually the `gensis_info`.
Will it be better to mention it in this comment? so the reader can have a better idea what this arbitrary information is about?

input_graphs/dev_utils_record_tests_smoke_other_peer_names/annie.dot
```diff
@@ -1174,10 +1175,6 @@ digraph GossipGraph {
 /// all_voters: {Annie, B, Carol, Claire}
 /// unconsensused_events: {"An_13", "B_9", "Ca_6", "Cl_10"}
 /// meta_events: {
-///   An_13 -> {
```
maqi: Just wondering, why the meta_events of these events can now be removed?
is it related to the change of hash, or the added `consensus_mode: Supermajority`, or because of the new added `related_info` ?

madadam: I only regenerated the file with the dot-gen tool. Maybe something changed since this file was created/modified the last time? No idea.

src/observation.rs
madadam: I named it `related_info` because we already have that in the `Add` and `Remove` variants and I think it makes sense to be consistent here. Maybe I can instead rename the `from_genesis` argument to, say, `genesis_related_info`?

maqi: up to you if you think a renaming will be better.
I just expect a bit more explicit comment :)

input_graphs/dev_utils_record_tests_smoke_other_peer_names/annie.dot
maqi: ok. we shall be safe as long as the test still passing.

Issue #312, opened 2019-05-30
panic in example at rng.gen_range


iancoleman:
The basic example will sometimes panic.

```
cargo run --release --example=basic --features=mock -- --initial-peers=20 --opaque=10 --seed=4138645876,637453496,194668239,3557059928
```
ends after round 120 with

```
thread 'main' panicked at 'Rng.gen_range called with low >= high', /home/ian/.cargo/registry/src/github.com-1ecc6299db9ec823/rand-0.4.2/src/lib.rs:519:9
```

Probable source for fix:

Related to
https://rust-random.github.io/rand/rand/trait.Rng.html#panics
and happens at
https://github.com/maidsafe/parsec/blob/8c0d9d00e8027f8c8212b4ad03c6ffadc3f22252/examples/basic.rs#L504
where the second term is problematic

pierrechevalier83:
Thanks a lot for the report @iancoleman :smile: 

The panick itself can easily be fixed by avoiding calling `gen_range(1, 1)`, returning `1` instead in that special case. We do get stuck, though with this seed and that change.

I started exploring this and got some observations:

We can reproduce it with a slightly smaller example:
```
cargo run --release --example=basic --features=mock -- --initial-peers=16 --opaque=0 --seed=4138645876,637453496,194668239,3557059928 --add-peers 0 --remove-peers 1
```
16 peers, 1 to remove, 0 to add, no opaque observations.

What happens in this case is that 16 nodes actually attempt to remove 13 nodes.
At the end, `Fred` doesn't seem to be participating anymore and `Alice` and `Kent` are stuck forever trying to remove `Fred`. This makes sense since they're only `2` out of `3` which is less than a supermajority.

I'm not sure exactly what we expect, but I am surprised in a  first place that we try to remove so many nodes with the `remove-peers 1` parameter.

@Fraser999, maybe you will remember better than me what we're aiming for in this example.

Issue #313, opened 2019-06-06
Basic example thinks it's OK to remove all peers but two, but it isn't.


pierrechevalier83:
I found this bug while investigating https://github.com/maidsafe/parsec/issues/312. It is related, but a separate issue:

The calculations we do to ensure we don't remove too many peers allow an attempt to remove all peers but two.

This isn't OK because once we are down to 3 peers, when we attempt to remove one of them, it becomes unresponsive (by design of the example), which means that we would need for 2 peers out of 3 to reach consensus on removing another. This is impossible as that one inactive peer is now `>= 1/3` of the peers, so they have the power of stalling consensus, which they do.

With the fix to issue 312 in place, it is easy to reproduce with an invocation such as this one:
```
cargo run --release --example=basic --features=mock -- --initial-peers=3 --opaque=0  --add-peers 0 --remove-peers 1
```

This is only the simplest expression of the problem. It can be reproduced with any arguments as long as `initial-peers - remove-peers == 2`.

I will raise one PR with 2 commits: one for issue 312 and one for this issue.

PR #314, opened 2019-06-06
Fix issues #312 and #313

pierrechevalier83:


examples/basic.rs
```diff
@@ -501,7 +505,18 @@ impl Environment {
             && self.params.remove_peers_count != self.peers_removed_count
             && self.rng.gen_weighted_bool(REMOVE_PEERS_CHANCE)
         {
-            self.rng.gen_range(1, (self.peers.len() + 2) / 3)
+            assert!(self.peers_removed_count < self.params.remove_peers_count);
+            let respect_remove_peers_count =
+                self.params.remove_peers_count - self.peers_removed_count;
+            let less_than_a_third = (self.peers.len() + 2) / 3;
+            let max_peers_to_remove = std::cmp::min(respect_remove_peers_count, less_than_a_third);
```
Fraser999: Should you import `std::cmp` at the top of the file to be consistent?

pierrechevalier83:
> Did you ran some 100's of times?

No, but I did run manually with a few meaningful examples that give me confidence in the correctness of the fix (varying the number of peers to remove and seeing the effect in the output).

PR #316, opened 2019-06-11
chore/deps: update dependencies

m-cat:


jeanphilippeD:
This would require updating routing similarly. Could we close this and re-open it when we need it and are ready to upgrade it in `routing` crate as well?

m-cat:
This wouldn't affect routing since it points to a specific parsec commit (`8c0d9d00e`). Also, my fleming version of https://github.com/maidsafe/routing/pull/1646 updates the necessary routing dependencies.

PR #318, opened 2019-06-13
refactor/parsec: simplify adding accusations and sync events.

jeanphilippeD:
We will add a new kind of event to add at the end with DkgMessages.
Clarify how we are doing it with our existing events.

### DIFF
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -1376,16 +1376,11 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
     // Constructs a sync event to prove receipt of a `Request` or `Response` (depending on the value
     // of `is_request`) from `src`, then add it to our graph.
     fn create_sync_event(&mut self, is_request: bool, other_parent: EventIndex) -> Result<()> {
-        if self.peer_list.last_event(PeerIndex::OUR).is_none() {
-            self.pending_events.push(PendingEvent::Sync {
-                is_request,
-                other_parent,
-                _phantom: PhantomData,
-            });
-            return Ok(());
-        };
-
-        self.add_sync_event(is_request, other_parent)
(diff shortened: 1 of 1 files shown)


src/parsec.rs
```diff
@@ -1554,9 +1556,10 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
             )?;
 
             let _ = self.add_event(event)?;
-            Ok(None)
+            Ok(())
```
maqi: The both branch return with `Ok(())`, can we make the function not requires returning with `Result`?
or at least move `Ok(())` out of the if clause to save one line?

jeanphilippeD: There are `?` in the code, so Result is necessary.

maqi: Then at least move `Ok(())` out of the if-clause ?

src/parsec.rs
```diff
@@ -1445,28 +1440,42 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-                } => self.add_sync_event(is_request, other_parent)?,
-                #[cfg(feature = "malice-detection")]
-                PendingEvent::Accusation {
-                    offender,
-                    malice,
-                    other_parent,
-                } => {
-                    let _ = self
-                        .add_accusation_event(offender, malice, other_parent)?
-                        .map(|unhandled| self.pending_accusations.push((offender, unhandled)));
```
pierrechevalier83: I don't know for sure what it entails, but I think this behaviour of pushing unhandled accusations to the pending_accusations got lost in the refactoring, which makes it more than a pure refactoring.
Am I missing something?

jeanphilippeD: No, it got moved inside add_accusation_event (unless I got it wrong in another way).

PR #320, opened 2019-06-18
Dkg in gossip refactor attempt

jeanphilippeD:


jeanphilippeD:
For reference, we probably do not want this refactor now.

PR #321, opened 2019-06-20
initiate DKG using a consensused event

### LINKED ISSUE #323: Enable initiating DKG from user of Parsec
PR #321  : New consensus event to start DKG

jeanphilippeD:
This change create a public way to start a DKG and get a result and update associated test.
This change does not address replaying dump-graph that include DKG.

Note:
-Initial commit started with an approach that avoided storing Rng in Parsec but was flawed.
 We have to trigger the DKG on event consensus whether we voted for it yet or not.
-The rang::Rng in the interface make consumer of Parsec depend on a specific version of rand.
 This can be addressed if we decide it would be a better approach when handling dump-graph and replay, possibly using std::io::Read as a stable interface.

Closes #323

### DIFF
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -109,7 +109,11 @@ struct Peer {
 }
 
 impl Peer {
-    fn from_genesis(our_id: PeerId, genesis_group: &BTreeSet<PeerId>) -> Self {
+    fn from_genesis(
+        our_id: PeerId,
+        genesis_group: &BTreeSet<PeerId>,
+        secure_rng: Box<dyn Rng>,
+    ) -> Self {
         Self {
             id: our_id.clone(),
             parsec: Parsec::from_genesis(
@@ -117,6 +121,7 @@ impl Peer {
                 genesis_group,
                 vec![],
                 ConsensusMode::Supermajority,
+                secure_rng,
             ),
             observations: vec![],
             blocks: vec![],
@@ -127,6 +132,7 @@ impl Peer {
         our_id: PeerId,
         genesis_group: &BTreeSet<PeerId>,
         section: &BTreeSet<PeerId>,
+        secure_rng: Box<dyn Rng>,
     ) -> Self {
         Self {
             id: our_id.clone(),
@@ -135,6 +141,7 @@ impl Peer {
                 genesis_group,
                 section,
                 ConsensusMode::Supermajority,
+                secure_rng,
             ),
             observations: vec![],
             blocks: vec![],
@@ -472,15 +479,15 @@ impl Environment {
         };
 
         // Set up the requested number of peers and random network events.
-        env.genesis_group = (0..env.params.genesis_peer_count)
+        let genesis_group: BTreeSet<_> = (0..env.params.genesis_peer_count)
             .map(|_| env.new_peer_id())
             .collect();
 
-        env.peers = env
-            .genesis_group
+        env.peers = genesis_group
             .iter()
-            .map(|id| Peer::from_genesis(id.clone(), &env.genesis_group))
+            .map(|id| Peer::from_genesis(id.clone(), &genesis_group, Box::new(env.rng.new_rng())))
             .collect();
+        env.genesis_group = genesis_group;
 
         env.opaque_observations = (0..env.params.opaque_event_count)
             .map(|_| parsec::Observation::OpaquePayload(env.rng.gen()))
@@ -561,7 +568,12 @@ impl Environment {
             self.peers_added_count += 1;
             println!("Adding {:?}", new_peer_id);
             let section = self.peers.iter().map(|peer| peer.id.clone()).collect();
-            let new_peer = Peer::from_existing(new_peer_id.clone(), &self.genesis_group, &section);
+            let new_peer = Peer::from_existing(
+                new_peer_id.clone(),
+                &self.genesis_group,
+                &section,
+                Box::new(self.rng.new_rng()),
+            );
             self.peers.push(new_peer);
         }
     }
--- a/src/dev_utils/mod.rs
+++ b/src/dev_utils/mod.rs
@@ -19,6 +19,7 @@ mod peer;
 mod peer_statuses;
 #[cfg(feature = "testing")]
 pub mod proptest;
+mod pseudo_random;
 #[cfg(any(all(test, feature = "mock"), feature = "testing"))]
 mod record;
 mod schedule;
@@ -30,11 +31,12 @@ pub(crate) use self::dot_parser::ParsedContents;
 #[cfg(any(all(test, feature = "mock"), feature = "testing"))]
 pub use self::record::Record;
 pub use self::{
-    environment::{Environment, RngChoice},
+    environment::Environment,
     misc::TestIterator,
     network::{ConsensusError, Network},
     peer::{NetworkView, Peer, PeerStatus},
     peer_statuses::PeerStatuses,
+    pseudo_random::{new_common_rng, new_rng, RngChoice, RngDebug},
     schedule::*,
 };
 
--- a/src/dev_utils/network.rs
+++ b/src/dev_utils/network.rs
@@ -642,7 +642,8 @@ impl Network {
             ScheduleEvent::StartDkg(peers) => {
                 // All valid peers should vote for new DKG.
                 for peer_id in self.running_peers_ids() {
-                    self.peer_mut(&peer_id).vote_for_new_dkg(peers.clone(), rng)
+                    self.peer_mut(&peer_id)
+                        .vote_for(&ParsecObservation::StartDkg(peers.clone()));
                 }
             }
         }
--- a/src/dev_utils/network.rs
+++ b/src/dev_utils/network.rs
@@ -7,6 +7,7 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use super::{
+    new_rng,
     peer::{NetworkView, Peer, PeerStatus},
     schedule::{Schedule, ScheduleEvent, ScheduleOptions},
     Observation,
@@ -537,13 +538,30 @@ impl Network {
                 let genesis_ids = genesis.all_ids();
                 let good_peers = genesis
                     .ids_of_good_peers()
-                    .map(|id| Peer::from_genesis(id.clone(), &genesis_ids, self.consensus_mode));
-                let malicious_peers = genesis.ids_of_malicious_peers().map(|id| {
-                    Peer::malicious_from_genesis(id.clone(), &genesis_ids, self.consensus_mode)
-                });
+                    .map(|id| {
+                        Peer::from_genesis(
+                            id.clone(),
+                            &genesis_ids,
+                            self.consensus_mode,
+                            new_rng(rng),
+                        )
+                    })
+                    .collect_vec();
+                let malicious_peers = genesis
+                    .ids_of_malicious_peers()
+                    .map(|id| {
+                        Peer::malicious_from_genesis(
+                            id.clone(),
+                            &genesis_ids,
+                            self.consensus_mode,
+                            new_rng(rng),
+                        )
+                    })
+                    .collect_vec();;
 
                 self.peers = good_peers
-                    .chain(malicious_peers)
+                    .into_iter()
+                    .chain(malicious_peers.into_iter())
                     .map(|peer| (peer.id().clone(), peer))
                     .collect();
 
@@ -574,6 +592,7 @@ impl Network {
                         &self.genesis,
                         &current_peers,
                         self.consensus_mode,
+                        new_rng(rng),
                     ),
                 );
             }
--- a/src/dev_utils/network.rs
+++ b/src/dev_utils/network.rs
@@ -639,10 +639,10 @@ impl Network {
 
                 self.peer_mut(&voting_peer_id).vote_for(&observation);
             }
-            ScheduleEvent::StartDkg(_) => {
-                // For test purpose: simulate starting DKG on all nodes at the same time.
+            ScheduleEvent::StartDkg(peers) => {
+                // All valid peers should vote for new DKG.
                 for peer_id in self.running_peers_ids() {
-                    self.peer_mut(&peer_id).dkg_start_consensus(rng)
+                    self.peer_mut(&peer_id).vote_for_new_dkg(peers.clone(), rng)
                 }
             }
         }
--- a/src/dev_utils/pseudo_random.rs
+++ b/src/dev_utils/pseudo_random.rs
@@ -1,0 +1,48 @@
+// Copyright 2019 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use maidsafe_utilities::SeededRng;
+use rand::{Rng, SeedableRng, XorShiftRng};
+use std::fmt;
+
+pub trait RngDebug: Rng + fmt::Debug {}
+
+impl RngDebug for SeededRng {}
+impl RngDebug for XorShiftRng {}
+
+#[derive(Clone, Copy, Debug)]
+pub enum RngChoice {
+    SeededRandom,
+    #[allow(unused)]
+    Seeded([u32; 4]),
+    SeededXor([u32; 4]),
+}
+
+/// Create a new Rng: Use once per test.
+pub fn new_common_rng(seed: RngChoice) -> Box<dyn RngDebug> {
+    match seed {
+        RngChoice::SeededRandom => Box::new(SeededRng::new()),
+        RngChoice::Seeded(seed) => Box::new(SeededRng::from_seed(seed)),
+        RngChoice::SeededXor(seed) => {
+            let rng = Box::new(XorShiftRng::from_seed(seed));
+            println!("Using {:?}", rng);
+            rng
+        }
+    }
+}
+
+/// Create a new RNG using a seed generated from random data provided by `rng`.
+pub fn new_rng<R: Rng>(rng: &mut R) -> Box<dyn Rng> {
+    let new_seed = [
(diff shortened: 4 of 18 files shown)


src/parsec.rs
```diff
@@ -1466,7 +1487,7 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
                 Err(_) => true,
                 Ok(block) => {
                     // Do not leak internal blocks to Parsec consumer
-                    !block.payload().is_dkg_message()
+                    !block.payload().is_dkg_internal()
```
pierrechevalier83: Based on this call site, I would recommend the name: `is_internal` and I would add a comment on top of the function that says:
```
// TODO: fix by making these 2 variants part of a private enum instead
```
(as discussed earlier, but postponed)

src/dev_utils/network.rs
```diff
@@ -639,10 +639,10 @@ impl Network {
 
                 self.peer_mut(&voting_peer_id).vote_for(&observation);
             }
-            ScheduleEvent::StartDkg(_) => {
-                // For test purpose: simulate starting DKG on all nodes at the same time.
+            ScheduleEvent::StartDkg(peers) => {
+                // All valid peers should vote for new DKG.
                 for peer_id in self.running_peers_ids() {
-                    self.peer_mut(&peer_id).dkg_start_consensus(rng)
+                    self.peer_mut(&peer_id).vote_for_new_dkg(peers.clone(), rng)
```
maqi: Just wondering what's the difference between the `peers` in the StartDkg, and the set in the `self.running_peers_ids` ?

src/parsec.rs
```diff
@@ -967,6 +973,21 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+        &mut self,
+        peers: BTreeSet<S::PublicId>,
+        rng: &mut rand::Rng,
+    ) -> Result<()> {
+        let threshold = peers.len().saturating_sub(1) / 3;
+        if let Ok(result) = KeyGen::new(self.peer_list.our_id(), peers.clone(), threshold, rng) {
+            self.pending_key_gen.push(result);
+            self.vote_for(Observation::DkgStart(peers))
+        } else {
+            Err(Error::FailedDkg)
```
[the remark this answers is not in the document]
jeanphilippeD: None of the variants have content, I thought originally of adding it but it seemed strange.
Will had an error level log though.

src/dev_utils/schedule.rs
```diff
@@ -343,7 +343,7 @@ pub enum ObservationEvent {
     AddPeer(PeerId),
     RemovePeer(PeerId),
     Fail(PeerId),
-    StartDkg,
+    StartDkg(BTreeSet<PeerId>),
```
maqi: better have a comment here to explain what is this set of peers?

src/observation.rs
```diff
@@ -60,6 +60,9 @@ pub enum Observation<T: NetworkEvent, P: PublicId> {
     },
     /// Vote for an event which is opaque to Parsec.
     OpaquePayload(T),
+    /// Internal only: No blocks with it.
+    /// Start the DKG when receive supermajority consensus.
```
maqi: so this `supermajority` is among the set of peers in the `DkgStart`, right?

src/parsec.rs
jeanphilippeD: I'll use `is_internal`.
I think I'll avoid putting refactoring TODO in the code though, especially since we'll probably need to put some thought into exactly what it will need to look like.

src/observation.rs
```diff
@@ -60,6 +60,9 @@ pub enum Observation<T: NetworkEvent, P: PublicId> {
...
+    DkgStart(BTreeSet<P>),
```
maqi: we have `DkgStart` in `Observation`, but `StartDkg` in `Schedule`.
shall they be consistent?

src/dev_utils/network.rs
jeanphilippeD: `peers` could potentially be a subset. Eventually, it could also be a superset.

src/observation.rs
jeanphilippeD: No, this is a normal transparent event like Add or Accusation, that will reach consensus on super-majority. Maybe I'll remove that part of the comment and try to make it more like the Add and other comments.

src/parsec.rs
```diff
@@ -93,6 +93,8 @@ pub(crate) type KeyGenId = usize;
 pub struct Parsec<T: NetworkEvent, S: SecretId> {
     // The PeerInfo of other nodes.
     peer_list: PeerList<S>,
+    // Set of pending distributed key generation.
+    pending_key_gen: Vec<(KeyGen<S>, Option<Part>)>,
```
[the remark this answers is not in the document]
jeanphilippeD: No, they are pending, we voted for it, but we are waiting for consensus to start it.
It is moved to the other map when they actually start.

src/parsec.rs
```diff
@@ -841,6 +844,12 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
 
                 self.handle_remove_peer(event_index, offender)
             }
+            Some(Observation::DkgStart(peers)) => {
+                if self.handle_dkg_start_consensus(&peers).is_none() {
+                    warn!("Ignoring DkgStart with Error");
```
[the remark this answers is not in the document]
jeanphilippeD: `Not starting DKG on DkgStart consensus because of error` ?

src/parsec.rs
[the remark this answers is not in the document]
pierrechevalier83: OK. Not sure I understand the need for an `Option<Part>`. Can you explain?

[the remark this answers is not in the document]
jeanphilippeD: Should it? the way we use it is by destructuring it and using the components, so we can't really encapsulate it. At best we can give this tuple a name and make it a struct with public fields. Not sure that would help things.

Did you have a name in mind?

src/parsec.rs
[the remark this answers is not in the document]
pierrechevalier83: > None of the variants have content
`InvalidPeerState` and `InvalidSelfState` do.
But I think consistency shouldn't be the very top concern in code. In decreasing order of importance: "do we have all the correctness properties we need (as proven by tests)", then "is our code easy to read", then "is it easy to debug", then "is it consistent". Consistency is like the last concern once all other concerns are addressed, imho :smile: (But yes, it is important too)

src/parsec.rs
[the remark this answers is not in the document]
jeanphilippeD: This is what KeyGen::new returns: We need all these elements when actually starting the DKG.

This is the part that must be sent around in gossip.
The reason it is an option is if the current Peer is not a member of the DKG (That property is not used/tested yet but was imported with the initial version of the code).

src/parsec.rs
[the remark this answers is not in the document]
jeanphilippeD: Correct, I misread that, very small state though, so we would quickly go into variant too large warning. Also, I think for logging purpose we are probably better logging it at the source.

I feel we probably want to use these variants to be able to recover from an error, and so knowing the internal of why the DKG fail is probably not something that user of Parsec can do anything about it.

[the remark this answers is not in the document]
jeanphilippeD: Do you think a log when the error is converted is not enough?

src/dev_utils/network.rs
maqi: is this `peers` the list of peers we have when we `start DKG` ?
just wondering why not to use `self.running_peers_ids` directly

jeanphilippeD: Because they are a different set.
The running_peers_ids are the voters of Parsec, and the peers are the DKG participants. They can be completely different, even if for now they are the same. When splitting, we will need to have DKG with 1/2 of the Parsec voters, and with node Ageing, we will need DKG with 1/2 of the Parsec voters and also non voters.

src/observation.rs
jeanphilippeD: Used `StartDkg` So it is a verb/Action like Add/Remove

src/parsec.rs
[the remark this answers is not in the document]
pierrechevalier83: > At best we can give this tuple a name and make it a struct with public fields. Not sure that would help things.

Yes, that's what I have in mind. Because then the names can explain what these mean. In general, I think tuples are convenient but easy to overuse which puts the responsibility on the reader to parse them out and figure out what they mean.

In this case, I was unsure about the exact meaning of the part but well named and commented struct would have explained it to me.

What about:
```
/// All the information needed to start a DKG instance.
/// We must transition from a PendingKeyGen to a started key generation as
/// an outcome of consensus to simulate synchrony in the asynchronous context of PARSEC
struct PendingKeyGen<S: SecretId> {
    /// Instance of key generation to be started once consensus is reached
    /// on the event containing this struct
    key_gen : KeyGen<S>,
    /// Part to share with other nodes
    /// None if we are participating in the Dkg, but not one of the nodes for which a key share is generated
    part : Option<Part>,
}
``` 
(Feel free to take liberties. I'm using it as an example of something I could have read and would have brought clarity to me)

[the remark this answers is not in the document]
jeanphilippeD: I think the problem that you are hitting here is that you are trying to give a name to what is essentially an aggregate of 2 things only related because they are both needed at the start of the process.

All of the documentation essentially repeating what is or should be in doc for KeyGen::new.
And the name is really inaccurate, it is PendingKeyGen and something else.
Also, both member need to be public.

I am not clear you gain anything compared to a simple comment at the start of `pending_key_gen`.
Maybe it could be something like:

```
    // Set of KeyGen::new results that are pending use when the associated StartDkg reach consensus.
    pending_key_gen: Vec<(KeyGen<S>, Option<Part>)>,
```

src/parsec.rs
```diff
@@ -953,6 +953,7 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         Some(())
     }
 
+    // Require that vote_for_new_dkg to vote for the consensused event.
```
pierrechevalier83: Maybe it's a grammar thing but this isn't clear to me.
What about:
```
// This function must be called on consensus on a `StartDkg` observation.
// This allow the call-site to meet two preconditions:
// * All DKG participants will call it in the same order, and
// * Incidentally, it will guarantee that `pending_key_gen` contains an
//   entry for this set of peers.
```

src/parsec.rs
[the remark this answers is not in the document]
pierrechevalier83: OK. I'm happy with the comment. It's more of a taste different for where you create a named type or not, so I won't insist.

jeanphilippeD:
Ok, the additional commit should address the problem: DKG should process should now be sound.
Also the Rng is injected. It may be that that we would want to use a `Box<dyn std::io::Read>` and then create for ourselves, maybe within the key_gen mod an actual Rng using `rand::rngs::adapter::ReadRng`.

It is probably outside the scope of this PR to replay dump-graphs with DKG gossip, but as mentioned in one of the commit there are multiple approach we can choose.

examples/basic.rs
```diff
@@ -472,15 +479,15 @@ impl Environment {
...
             .collect();
 
-        env.peers = env
-            .genesis_group
+        env.peers = genesis_group
             .iter()
-            .map(|id| Peer::from_genesis(id.clone(), &env.genesis_group))
+            .map(|id| Peer::from_genesis(id.clone(), &genesis_group, Box::new(env.rng.new_rng())))
             .collect();
+        env.genesis_group = genesis_group;
```
maqi: just want to confirm that this part of change `env.genesis_group => genesis_group` is required just because we now need to pass an extra `Box::new(env.rng.new_rng())` in the `from_genesis` function, right?

src/dev_utils/mod.rs
```diff
@@ -19,6 +19,7 @@ mod peer;
 mod peer_statuses;
 #[cfg(feature = "testing")]
 pub mod proptest;
+mod pseudo_random;
```
maqi: wondering can this new mod be part of the `seeded_rng` or similar to it being in the maidsafe_util crate?

src/dev_utils/network.rs
```diff
@@ -537,13 +538,30 @@ impl Network {
...
+                            self.consensus_mode,
+                            new_rng(rng),
+                        )
+                    })
+                    .collect_vec();;
 
                 self.peers = good_peers
-                    .chain(malicious_peers)
+                    .into_iter()
+                    .chain(malicious_peers.into_iter())
```
maqi: I'd expect this change is not required? is it because the above there is extra `collect_vec` be used?

src/parsec.rs
```diff
@@ -117,6 +117,8 @@ pub struct Parsec<T: NetworkEvent, S: SecretId> {
     // parsec instances.
     #[cfg(any(test, feature = "testing"))]
     ignore_process_events: bool,
+    // Secure RNG
```
maqi: why it's secure? or the rng is for secure purpose?
the comment is better to be expanded little bit?

src/dev_utils/pseudo_random.rs
```diff
@@ -0,0 +1,48 @@
...
+}
+
+/// Create a new Rng: Use once per test.
+pub fn new_common_rng(seed: RngChoice) -> Box<RngDebug> {
+    match seed {
+        RngChoice::SeededRandom => Box::new(SeededRng::new()),
+        RngChoice::Seeded(seed) => Box::new(SeededRng::from_seed(seed)),
+        RngChoice::SeededXor(seed) => {
+            let rng = Box::new(XorShiftRng::from_seed(seed));
+            println!("Using {:?}", rng);
```
maqi: pushed by mistake?

src/dev_utils/pseudo_random.rs
```diff
@@ -0,0 +1,48 @@
...
+
+pub trait RngDebug: Rng + fmt::Debug {}
+
+impl RngDebug for SeededRng {}
+impl RngDebug for XorShiftRng {}
+
+#[derive(Clone, Copy, Debug)]
+pub enum RngChoice {
+    SeededRandom,
+    #[allow(unused)]
```
maqi: does this exception really required?

src/dev_utils/pseudo_random.rs
```diff
@@ -0,0 +1,48 @@
+// Copyright 2018 MaidSafe.net limited.
```
maqi: 2019?

src/parsec.rs
jeanphilippeD: Will update the comment, yes, the provided RNG need to be cryptographically secure.

src/dev_utils/pseudo_random.rs
jeanphilippeD: This is the code it replaced:

jeanphilippeD: This is where it comes from:
```
    pub fn with_consensus_mode(seed: RngChoice, consensus_mode: ConsensusMode) -> Self {
        let rng: Box<RngDebug> = match seed {
            RngChoice::SeededRandom => Box::new(SeededRng::new()),
            RngChoice::Seeded(seed) => Box::new(SeededRng::from_seed(seed)),
            RngChoice::SeededXor(seed) => {
                let rng = Box::new(XorShiftRng::from_seed(seed));
                println!("Using {:?}", rng);
                rng
            }
        };
```

So no changes here.

src/dev_utils/pseudo_random.rs
jeanphilippeD: Same, only moved.
Yes it is needed since it is only really used when re-compilling with a specific seed.

src/dev_utils/pseudo_random.rs
jeanphilippeD: Will update

examples/basic.rs
jeanphilippeD: Yes, the borrow checker complain otherwise

src/dev_utils/mod.rs
jeanphilippeD: I think we want to get away from maidsafe_util. In any case, better to first get it right within the crate before moving it to another crate.

src/dev_utils/network.rs
jeanphilippeD: Yes, before good_peers and malicious_peers were already into iterators.
Now they are vector. Still borrow checker.

maqi: Can we then remove the new added `collect_vec`, to keep the same as previous?

jeanphilippeD: No, the borrow checker complain because `new_rng(rng)` take a mutable reference on rng, so we cannot have both closure at the same time. That is why using collect_vec we use the closure immediatly, one after the other.

src/parsec.rs
```diff
@@ -944,17 +959,20 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
-    #[allow(unused)]
-    /// Start DKG: Needs to be called at the same block for all Nodes.
-    pub fn handle_dkg_start_consensus(&mut self, rng: &mut rand::Rng) -> Option<()> {
-        let voters: BTreeSet<S::PublicId> = self
-            .peer_list
-            .voters()
-            .map(|(_, p)| p.id().clone())
-            .collect();
+    // This function must be called on consensus on a `StartDkg` observation.
+    fn handle_dkg_start_consensus(&mut self, peers: &BTreeSet<S::PublicId>) -> Option<()> {
```
fizyk20: Why an `Option<()>`, and not just a `bool`, which is conceptually the same?

jeanphilippeD: Not changed from before I think.
Mostly so i could use `?` in the code, could fix it if you want.

src/dev_utils/pseudo_random.rs
```diff
@@ -0,0 +1,48 @@
...
+#[derive(Clone, Copy, Debug)]
+pub enum RngChoice {
+    SeededRandom,
+    #[allow(unused)]
+    Seeded([u32; 4]),
+    SeededXor([u32; 4]),
+}
+
+/// Create a new Rng: Use once per test.
+pub fn new_common_rng(seed: RngChoice) -> Box<RngDebug> {
```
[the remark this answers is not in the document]
jeanphilippeD: I'll add it.

src/parsec.rs
fizyk20: No need, I don't feel strongly about it. Possibility of using `?` is a good enough reason for me, too :+1:

PR #335, opened 2019-07-23
feat/parsec: dump and reload KeyGen

jeanphilippeD:
To allow tests that load Parsec using `Parsec::from_parsed_contents`
when a DKG is in progress, dump existing KeyGen as serialized data
and reload them.

This is a very basic facility, and we may want to update it in the
future to provide more useful information in the dump-graphs.

Update dot_gen to support extracting a specific dump in a serie so
we can pick a graph that contain a DKG in progress.

Test:
Add new test that exercise dump and reload.

### DIFF
--- a/dot_gen/src/main.rs
+++ b/dot_gen/src/main.rs
@@ -361,6 +361,20 @@ fn add_dev_utils_record_smoke_tests(scenarios: &mut Scenarios) {
         .seed([1, 2, 3, 4])
         .file("Alice", "alice.dot")
         .dump_mode(DumpGraphMode::OnParsecDrop);;
+
+    let _ = scenarios
+        .add("dev_utils::record::tests::smoke_partial_dkg", |env| {
+            let peer_ids = peer_ids!("Alice", "Bob", "Carol");
+            let dkg_peer_ids = peer_ids!("Alice", "Bob", "Carol", "Dave", "Eric");
+            let obs = ObservationSchedule {
+                genesis: Genesis::new(peer_ids.clone()),
+                schedule: vec![(1, StartDkg(dkg_peer_ids))],
+            };
+            Schedule::from_observation_schedule(env, &ScheduleOptions::default(), obs)
+        })
+        .seed([1, 2, 3, 4])
+        .file("Alice-005", "alice.dot")
+        .dump_mode(DumpGraphMode::OnConsensus);
 }
 
 fn add_benches(scenarios: &mut Scenarios) {
--- a/src/dev_utils/dot_parser.rs
+++ b/src/dev_utils/dot_parser.rs
@@ -83,6 +83,7 @@ struct ParsedFile {
     graph: ParsedGraph,
     meta_election: ParsedMetaElection,
     secure_rng_values: Vec<u32>,
+    serialized_key_gens_and_next_id: Option<Vec<u8>>,
 }
 
 struct ParserCtx {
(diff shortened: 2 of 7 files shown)


dot_gen/src/main.rs
```diff
@@ -545,7 +559,7 @@ struct Scenario {
     name: String,
     seed: RngChoice,
     schedule_fn: Box<FnMut(&mut Environment) -> Schedule>,
-    files: BTreeMap<String, String>,
+    files: BTreeMap<(String, Option<u32>), String>,
```
fizyk20: The optional `u32` could use a comment as to what is its purpose - maybe even replacing this tuple with a proper struct could be a good idea.

src/parsec.rs
```diff
@@ -2360,6 +2362,19 @@ impl Parsec<Transaction, PeerId> {
                 .map(|info| info.consensused = true);
         }
 
+        if let Some(serialized_key_gens_and_next_id) =
+            &parsed_contents.serialized_key_gens_and_next_id
+        {
+            // `deserialise` require `Serialize` which is only available with `dump-graphs`
```
fizyk20: Maybe you could elaborate on why we need to deserialise things outside of the `dump-graphs` functionality (and also maybe how `Cursor` helps here).

jeanphilippeD: I'll try to clarify, in comment. Cursor is really just a way to get std::io::Read object so we can use that specific function from maidsafe_utils::serialisation. The problem with the one that take a vector is that it is over constrained and require Serialize.

We need to deserialize everytime we are loading dot file as part of `testing` feature which does not require dump-graphs.

dot_gen/src/main.rs
jeanphilippeD: Turn out there is an existing trick that I can use here, so I will not add that functionality:
see functional_tests::unpolled_and_unconsensused_observations
using "Alice-005" as the peer name.

src/dev_utils/dot_parser.rs
```diff
@@ -127,18 +128,29 @@ fn parse_file(ctx: &Rc<ParserCtx>) -> Parser<u8, ParsedFile> {
         + parse_peer_list(ctx)
         + parse_consensus_mode()
         + parse_secure_rng()
+        + parse_key_gens()
         + parse_graph()
         + parse_meta_election(ctx)
         - parse_end())
     .map(
-        |(((((our_id, peer_list), consensus_mode), secure_rng_values), graph), meta_election)| {
+        |(
```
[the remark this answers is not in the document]
jeanphilippeD: Yes. Well it is a bit messy, I wonder if there is a better way to express that without writting down that scary tuple of tuple.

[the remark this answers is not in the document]
maqi: maybe another good reason to use a proper struct  :)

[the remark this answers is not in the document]
jeanphilippeD: Not really an option here as it is what we get from the parsing infrastructure, but I'll see if I can improve how we destructure it now that it has become a very big tuple.

jeanphilippeD:
Fixed review comments.

jeanphilippeD:
Fixed comments, It took me 2 tries to fix it and It only show the last force push now for reasons.
So I forced push the original again, and force push the new version again so it shows all that has changed.

src/parsec.rs
```diff
@@ -2360,6 +2362,22 @@ impl Parsec<Transaction, PeerId> {
                 .map(|info| info.consensused = true);
         }
 
+        if let Some(serialized_key_gens_and_next_id) =
+            &parsed_contents.serialized_key_gens_and_next_id
+        {
+            // `serialisation::deserialise` is over constrained and require `Serialize` which is
+            // only available with `dump-graphs` enable (as it is not safe to dump secrets).
```
fizyk20: Typo, "enable" → "enabled" ;)

jeanphilippeD:
Fixed typo.

PR #339, opened 2019-08-26
resolve the issues in tie-breaking and interesting content ordering

maqi:
This PR contains two commits that trying to resolve the issues in tie-breaking and interesting content ordering.
Also contains the tests to verify the resolvement.

The last commit of the PR contains the re-generated dot files due to the refactoring.
Better review the diff by selecting the first two commits to reduce the diffs.


src/dev_utils/network.rs
```diff
@@ -73,7 +73,15 @@ impl fmt::Debug for DifferingBlocksOrder {
             .zip(self.order_2.order.iter())
             .enumerate()
         {
-            writeln!(formatter, "  {}. {:?} / {:?}", i + 1, block1, block2)?;
+            writeln!(
```
jeanphilippeD: could you keep the same `writeln!(formatter, "  {}. {:?} / {:?}", i + 1, block1, block2)?;` won't it format the tuple reasonably?

src/dev_utils/network.rs
```diff
@@ -170,14 +178,19 @@ impl Network {
             .running_non_malicious_peers()
             .find(|peer| peer.blocks_payloads() != payloads)
         {
+            let order_1 = first_peer
```
jeanphilippeD: Can you add a free function in this file (to avoid poluting the peer interface)

`fn block_keys(&Peer) -> Vec <...> {}`

And use it in the 4 instance in this file

src/dev_utils/network.rs
```diff
@@ -320,21 +338,18 @@ impl Network {
         Ok(())
     }
 
-    fn block_key<'a>(
-        &self,
-        block: &'a Block<Transaction, PeerId>,
-    ) -> (&'a Observation, Option<&'a PeerId>) {
+    fn block_key(&self, block: &Block<Transaction, PeerId>) -> (Observation, Option<PeerId>) {
```
jeanphilippeD: Why did you need to change that interface? why are reference not working.
Could/should you clone from the caller of these function if it is needed?

src/dev_utils/schedule.rs
```diff
@@ -800,9 +826,21 @@ pub enum Sampling {
...
+    rng: &mut R,
+    items: &[T],
+    strategy: Sampling,
+    choose_all: bool,
+) -> Vec<T> {
     let amount = match strategy {
-        Sampling::Constant(amount) => amount,
+        Sampling::Constant(amount) => {
+            // For the special case that all peers vote for the same opaqueload
+            if amount == 1 && choose_all {
```
jeanphilippeD: We probably want to use Sampling to do that.
Could use Fraction(1.0,1.0), should be fine (though I don't really trust floating point). Prefered option
Alternatively we can add a 'All' variant.

src/parsec_helpers.rs
```diff
@@ -79,7 +81,13 @@ where
 
     // Sort the payloads in the order the creator voted for them, followed by the ones
     // not voted for by the creator (if any).
-    interesting_payload_keys.sort();
+    interesting_payload_keys.sort_by(|(l_index, l_key), (r_index, r_key)| {
+        if l_index == r_index {
```
jeanphilippeD: I think you can express that as `l_index.cmp(r_index).and_then(|| consistent_cmp(l_key, r_key))`

src/parsec_helpers.rs
```diff
@@ -93,6 +101,7 @@ mod tests {
     use super::*;
     use crate::{
         hash::Hash,
+        // mock::PeerId,
```
jeanphilippeD: uneeded?

src/parsec_helpers.rs
```diff
@@ -132,6 +146,20 @@ mod tests {
             }
         }
 
+        fn new_single(
```
jeanphilippeD: only difference seem to be the consensus mode => Probably best to pass it as parameter?

src/parsec_helpers.rs
```diff
@@ -231,6 +259,23 @@ mod tests {
                     ],
                 }
             }
+
+            /// A simple setup for Events that can be used in multiple tests.
+            fn new_basic_setup_single() -> Self {
```
jeanphilippeD: Same only difference seem to be consensus mode, probably best to pass it as parameter.

src/parsec_helpers.rs
```diff
@@ -246,6 +291,10 @@ mod tests {
             let payloads = find_interesting_content_for_event(
                 &events.builder_event,
                 events.unconsensused_events.iter(),
+                |lhs_key, rhs_key| {
+                    NAMES[unwrap!(lhs_key.peer_index()).value()]
```
jeanphilippeD: Would it be enough with the following, then we don't need names NAMES?
It probably also make clearer what we test.

```
|lhs_key, rhs_key| {
   // Reverse order
   lhs_key.peer_index().cmp(rhs_key.peer_index()).rev() 
}
```

src/peer_list/peer_index.rs
```diff
@@ -23,6 +23,11 @@ impl PeerIndex {
     pub fn new_test_peer_index(index: usize) -> Self {
         Self(index)
     }
+
+    #[cfg(test)]
+    pub fn value(self) -> usize {
```
jeanphilippeD: No need with other changes.

src/parsec_helpers.rs
```diff
@@ -8,13 +8,15 @@
...
 use itertools::Itertools;
+use std::cmp::Ordering;
 use std::usize;
 
 /// Find interesting payloads for the builder_event.
 /// For payload observed from builder_event, order them by creation index.
 pub(crate) fn find_interesting_content_for_event<'a, E>(
     builder_event: E,
     unconsensused_events: impl Iterator<Item = E>,
+    consistent_cmp: impl Fn(&ObservationKey, &ObservationKey) -> Ordering,
```
jeanphilippeD: I would want to see in commit comment a seed of a failing Parsec integration test that this fixes.
=> This could be the one added in this PR, or maybe it needs to be a bit more complex, but should probably fail within 10 iteration of running the test.
=> I'm thinking we probably need to have not all peers voting in same order, so at least 2 opaque payloads will be needed, possibly more.

src/dev_utils/schedule.rs
maqi: Will have a check, maybe an `All` variant is more reasonable

src/parsec_helpers.rs
maqi: I don't think so. 
This still demos the comparing of peer_index, although in the different order.
I think we'd like to demo the way it being used in the production code?

jeanphilippeD: I see that as a mock.
if you want to make it clearer you can create a test function name it properly and add a comment, but we don't want to add complexity for the sake of looking more like the real code. As long as the injected function object does what needed for the test we are good.


```
fn fake_consistent_cmp(lhs_key,rhs_key) -> Ordering {
   // Real consistent_cmp will compare the PublicId, as a result they can be in any order.
   // For test purpose we ensure they are reverse order of the peer_index so we do not sort by it.
   lhs_key.peer_index().cmp(rhs_key.peer_index()).rev() 
}
```


Or to fully test the code, ensure that we test both:

```
fn fake_consistent_cmp(lhs_key,rhs_key, is_reverse: bool) -> Ordering {
   // Real consistent_cmp will compare the PublicId, as a result they can be in any order.
   // For test purpose we ensure they are either the same or reverse order of the peer_index so we do not sort by it.
   if is_reverse {
   lhs_key.peer_index().cmp(rhs_key.peer_index()).rev()
   } 
  else {
   lhs_key.peer_index().cmp(rhs_key.peer_index())
 }
}
```

And then run the same test with either true or false (use sub function).

src/parsec_helpers.rs
maqi: It was the routing test failing :(

>we probably need to have not all peers voting in same order, so at least 2 opaque payloads will be needed, possibly more

maybe, need to have a check how to do it

jeanphilippeD: Yes, we found it with the routing test failing, what i meant is we probably also needed to make an integration test in parsec that was showing the failure and post the seed that was failing (it can all be in the same commit) :-)

src/parsec_helpers.rs
maqi: Ok, sounds a good iead

src/parsec_helpers.rs
maqi: hmm...  not sure whether there is a failing parsec test with the fix, will have a check.
if not, probably need a new test. However, maybe the new test will fail always without the resolvement, so including a seed may not necessary?

jeanphilippeD: > However, maybe the new test will fail always without the resolvement, so including a seed may not necessary?

Yes, that would be a great thing if it fail quickly can just say that the new test fail quickly without the sorting fix :-)

> probably need a new test.
Yes most likely, I'm thinking the new test added in this PR is not complex enough to catch that issue, this is what I was trying to say in this review comment.

maqi: got a failing seed of the new test `consensus_mode_single_all_vote_same` within about 15 iterations. Shall I use this?
Also, I think this seed then shall be included in the `next` resolvement commit, i.e. in the commit of consistent ordering, instead of `this` commit (i.e. refactoring tie-breaking rules) ?

jeanphilippeD:
Test, can you reply?

maqi:
> Test, can you reply?

can only reply here

src/dev_utils/network.rs
```diff
@@ -320,6 +320,19 @@ impl Network {
         Ok(())
     }
 
+    fn block_keys(&self, peer: &Peer) -> Vec<(Observation, Option<PeerId>)> {
+        peer.blocks()
+            .map(|block| {
+                let (obs, opt_peer_id) = self.block_key(block);
```
jeanphilippeD: Is that not enough?
```
.map(|block| {
  let (obs, opt_peer_id) = self.block_key(block); 
  (obs.clone(), opt_peer_id.cloned())
  }
```

src/dev_utils/schedule.rs
```diff
@@ -590,7 +616,13 @@ impl Schedule {
         if options.votes_before_gossip {
             let opaque_transactions = obs_schedule.extract_opaque();
             let sampling = match env.network.consensus_mode() {
-                ConsensusMode::Single => Sampling::Constant(1),
+                ConsensusMode::Single => {
+                    if options.choose_all {
+                        Sampling::Constant(peers.all_peers().count())
```
jeanphilippeD: this is not ideal because it only work if number of peer do not change.

src/dev_utils/schedule.rs
```diff
@@ -684,7 +716,13 @@ impl Schedule {
                     ObservationEvent::Opaque(payload) => {
                         let observation = ParsecObservation::OpaquePayload(payload);
                         let sampling = match env.network.consensus_mode() {
-                            ConsensusMode::Single => Sampling::Constant(1),
+                            ConsensusMode::Single => {
+                                if options.choose_all {
+                                    Sampling::Constant(peers.all_peers().count())
```
jeanphilippeD: Same

src/dev_utils/schedule.rs
```diff
@@ -398,17 +404,37 @@ impl ObservationSchedule {
...
+        let allowed_to_same = if options.vote_for_same {
+            options.opaque_to_add * (options.genesis_size - 1)
+        } else {
+            0
+        };
+
         let mut step: usize = 1;
         while num_observations
-            < options.opaque_to_add + options.peers_to_add + options.peers_to_remove
+            < options.opaque_to_add
```
jeanphilippeD: Do we really need that change, Supermajority conensus was able to vote the same payload from different peers without it.

src/dev_utils/schedule.rs
```diff
@@ -398,17 +404,37 @@ impl ObservationSchedule {
...
+                + allowed_to_same
         {
             if opaque_count < options.opaque_to_add && rng.gen::<f64>() < options.prob_opaque {
                 schedule.push((
                     step,
                     ObservationEvent::Opaque(Transaction::new(opaque_count.to_string())),
                 ));
                 num_observations += 1;
-                opaque_count += 1;
+                if options.vote_for_same {
```
jeanphilippeD: Same

src/dev_utils/schedule.rs
```diff
@@ -280,6 +280,10 @@ pub struct ScheduleOptions {
     pub intermediate_consistency_checks: bool,
     /// The only genesis members that will compute consensus if provided. All if none.
     pub genesis_restrict_consensus_to: Option<BTreeSet<PeerId>>,
+    /// Allows for voting for the same OpaquePayload
+    pub vote_for_same: bool,
+    /// Enforce all the peers are selected to vote for observation
+    pub choose_all: bool,
```
jeanphilippeD: Why is there 2 flags?

tests/integration_tests.rs
```diff
@@ -641,3 +641,19 @@ proptest! {
...
     }
 }
+
+#[test]
+fn consensus_mode_single_all_vote_same() {
+    let mut env = Environment::with_consensus_mode(SEED, ConsensusMode::Single);
+    let options = ScheduleOptions {
+        genesis_size: 8,
+        opaque_to_add: 2,
+        opaque_voters: Sampling::Constant(8),
```
jeanphilippeD: All of these seem tightly coupled and we probably want only 1 parameter to drive them (ideally)

```
         opaque_voters: Sampling::Constant(8),
         vote_for_same: true,
        choose_all: true, 
```

src/dev_utils/network.rs
maqi: No, cannot.
The `opt_peer_id` is actually `Option<&PeerId>`, cloning it cannot covert it into `Option<PeerId>`.
Doing so will have an error of 
```a collection of type `std::vec::Vec<(observation::Observation<mock::Transaction, mock::PeerId>, std::option::Option<mock::PeerId>)>` cannot be built from `std::iter::Iterator<Item=(observation::Observation<mock::Transaction, mock::PeerId>, std::option::Option<&mock::PeerId>)>`
```

src/dev_utils/schedule.rs
maqi: will remove the `choose_all` flag

src/dev_utils/network.rs
jeanphilippeD: But `cloned` function can.

https://doc.rust-lang.org/std/option/enum.Option.html#method.cloned

maqi: ah, sorry, my fault. messed `cloned` with `clone`.
yeah, you are right, it works.

src/dev_utils/schedule.rs
```diff
@@ -575,8 +578,12 @@ impl Schedule {
         let mut pending = PendingObservations::new(options);
 
         // the +1 below is to account for genesis
-        let max_observations =
-            obs_schedule.count_observations() + obs_schedule.count_expected_accusations() + 1;
+        let max_observations = if options.vote_for_same {
```
jeanphilippeD: Maybe 
```rust
let observation_multiplier = if options.vote_for_same {
   options.genesis_size
} else {
   1
};
let max_observations =
            obs_schedule.count_observations() * observation_multiplier + obs_schedule.count_expected_accusations() + 1; 
```

tests/integration_tests.rs
```diff
@@ -641,3 +641,18 @@ proptest! {
...
```
jeanphilippeD: Why do we need that `opaque_voters: Sampling::Constant(8),` do the test fail if it is not there or does no longer detects the bug we fixed?

maqi: forgot to remove. will remove it.

src/dev_utils/schedule.rs
madadam: +1 for `Sampling::All`. Would make the logic easier to follow. Also you wouldn't need the `choose_all` param in `ScheduleOptions` and elsewhere.

madadam: > Could use Fraction(1.0,1.0), should be fine (though I don't really trust floating point)

That should be fine too. Note that `1.0` can be represented exactly as `f64`, so there should be no weird rounding issues in this case. Agree that floats are tricky so we might want to replace it with some kind of rational type (maybe this: https://docs.rs/num-rational/0.2.2/num_rational) eventually.

madadam: Hm, I see the issue with that: https://github.com/maidsafe/parsec/blob/master/src/dev_utils/schedule.rs#L592-L595 - if `consensus_mode` is `Single`, it hardcodes the sampling to `Sampling::Constant(1)`, so the suggested changes wouldn't work. 

Here is an idea: use `options.opaque_voters` even if `consensus_mode == Single`. Then change the tests which use `ConsensusMode::Single` to explicitly use `opaque_voters: Sampling::Constant(1)` (except that one new test you wrote).

Alternative idea: rename `ScheduleOptions::opaque_voters` to `supermajority_opaque_voters` and add new `single_opaque_voters` with default value `Sampling::Constant(1)`. Then use one or the other depending on `consensus_mode` (this would have to be changed here: https://github.com/maidsafe/parsec/blob/master/src/dev_utils/schedule.rs#L644-L653, here: https://github.com/maidsafe/parsec/blob/master/src/dev_utils/schedule.rs#L686-L689 and here: https://github.com/maidsafe/parsec/blob/master/src/dev_utils/schedule.rs#L592-L595.)

src/dev_utils/schedule.rs
madadam: I have the same question. I don't think we need `vote_for_same`, because if `choose_all` is `true` we end up casting multiple votes (with the same payload) even if `consensus_mode == Single` anyway.

src/parsec_helpers.rs
```diff
@@ -216,27 +222,42 @@ mod tests {
...
+            } else {
+                lhs_key.peer_index().cmp(&rhs_key.peer_index())
+            }
+        }
+
         /// Test function for find_interesting_content_for_event
         /// Call with different TestSimpleData for data driven tests
         /// (Follows AAA(Arrange/Act/Assert) test Pattern)
-        fn test_find_interesting_content_for_event(data: TestSimpleData) {
+        fn test_find_interesting_content_for_event(data: TestSimpleData, is_reverse: bool) {
```
madadam: I would put the `is_reverse` argument as a field of `TestSimpleData`. I think it would be more readable.

src/dev_utils/schedule.rs
```diff
@@ -574,9 +577,15 @@ impl Schedule {
     ) -> Schedule {
         let mut pending = PendingObservations::new(options);
 
+        let observation_multiplier = if options.vote_for_same {
```
madadam: Should we make sure the `vote_for_same` option only applies when `consensus_mode` is `Single`? Or does it make sense for `Supermajority` too?

src/dev_utils/schedule.rs
```diff
@@ -280,6 +280,8 @@ pub struct ScheduleOptions {
     pub intermediate_consistency_checks: bool,
     /// The only genesis members that will compute consensus if provided. All if none.
     pub genesis_restrict_consensus_to: Option<BTreeSet<PeerId>>,
+    /// Allows for voting for the same OpaquePayload
```
madadam: I would expand this comment to mention that this applies only to `consensus_mode == Single`, because in `consensus_mode == Supermajority` peers vote for the same payload anyway.

src/dev_utils/schedule.rs
jeanphilippeD: @madadam is that still relevant, I think this code is no longer changed in the final version of the PR (with fixups?)

madadam: This is marked as outdated, so no, this is not relevant anymore.

src/dev_utils/schedule.rs
maqi: I will put a checker of `consensus_mode is Single` here.
Or do you think a comment below (as you suggested) will be enough?

PR #344, opened 2019-10-02
unresponsive node handling

maqi:
This PR contains the work of detecting un-responsive node, then voted for remove it following the normal node removal flow. Correspondent integration tests are also included.
The `un-responsive` is defined as a node does not appear as a voter during certain amount of consensused observations.

### DIFF
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -85,6 +85,10 @@ impl VoteStatus {
         }
     }
 
+    fn remove_voter(&mut self, voter: PeerIndex) {
+        let _ = self.unvoted.remove(voter);
+    }
+
     fn set_consensused(&mut self, payload_key: &ObservationKey) -> bool {
         if &self.payload_key == payload_key {
             self.consensused = true;
@@ -335,6 +339,15 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
 
         if self.have_voted_for(&observation) {
             return Err(Error::DuplicateVote);
+        }
+
+        // To avoid unnecessary Left vote for a removed peer.
+        if let Observation::Remove { ref peer_id, .. } = observation {
+            if let Some(peer_index) = self.peer_list.get_index(peer_id) {
+                self.vote_statuses
+                    .iter_mut()
+                    .for_each(|vote_status| vote_status.remove_voter(peer_index));
+            }
         }
 
         self.flush_pending_events()?;
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -54,6 +54,51 @@ use std::{
 
 pub(crate) type KeyGenId = usize;
 
+/// Affects the threshold (number of unvoted votes) a node be considered as unresponsive.
+const UNRESPONSIVE_THRESHOLD_FRACTION: usize = 3;
+/// Affects the period (X consensued observations) after which node be considered as unresponsive.
+const UNRESPONSIVE_FACTOR: usize = 10;
+
+struct VoteStatus {
+    payload_key: ObservationKey,
+    unvoted: PeerIndexSet,
+    consensused: bool,
+}
+
+impl VoteStatus {
+    fn new(payload_key: &ObservationKey, voter: PeerIndex, all_voters: PeerIndexSet) -> Self {
+        let mut vote_status = VoteStatus {
+            payload_key: *payload_key,
+            unvoted: all_voters,
+            consensused: false,
+        };
+        let _ = vote_status.add_vote(payload_key, voter);
+        vote_status
+    }
+
+    fn add_vote(&mut self, payload_key: &ObservationKey, voter: PeerIndex) -> bool {
+        if &self.payload_key == payload_key {
+            let _ = self.unvoted.remove(voter);
+            true
+        } else {
+            false
+        }
+    }
+
+    fn set_consensused(&mut self, payload_key: &ObservationKey) -> bool {
+        if &self.payload_key == payload_key {
+            self.consensused = true;
+            true
+        } else {
+            false
+        }
+    }
+
+    fn is_unresponsive(&self, peer_index: PeerIndex) -> bool {
+        self.consensused && self.unvoted.contains(peer_index)
+    }
+}
+
 /// The main object which manages creating and receiving gossip about network events from peers, and
 /// which provides a sequence of consensused [Block](struct.Block.html)s by applying the PARSEC
 /// algorithm. A `Block`'s payload, described by the [Observation](enum.Observation.html) type, is
@@ -121,6 +166,8 @@ pub struct Parsec<T: NetworkEvent, S: SecretId> {
     ignore_process_events: bool,
     // Provided RNG: Needs to be cryptographically secure RNG as it is used for DKG key generation.
     secure_rng: ParsecRng,
+    // Including all on-going, consensused and polled blocks
+    vote_statuses: VecDeque<VoteStatus>,
 }
 
 impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
@@ -265,6 +312,7 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
             ignore_process_events: false,
 
             secure_rng: ParsecRng::new(secure_rng),
+            vote_statuses: Default::default(),
         }
     }
 
@@ -696,6 +744,10 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
                 }
             });
 
+        if event.is_observation() {
+            self.add_observation_vote(&event);
+        }
+
         let event_index = self.insert_event(event);
 
         let _ = unconsensused_payload_key.map(|payload_key| {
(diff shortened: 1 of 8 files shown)


src/parsec.rs
```diff
@@ -715,6 +771,74 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+            self.vote_status.len() / 3,
+        );
+
+        // Only keep 10 * threshold consensused voting status and the unconsensused ones among them.
+        let mut concensused = self
+            .vote_status
+            .iter()
+            .filter(|vote_status| vote_status.consensused)
+            .count();
+        while concensused > 10 * threshold {
```
madadam: Don't use magic numbers in the code. Define a named constant with this value instead.

madadam: Can this condition ever be `true`? So if the number of entries in `vote_status` is `N`, then `consensused` is at most `N`, so this is like saying: `N > 10/3 * N` which is never `true`. Am I missing something?

src/parsec.rs
```diff
@@ -715,6 +771,74 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+            if let Some(vote_status) = self.vote_status.pop_front() {
+                if vote_status.consensused {
+                    concensused -= 1;
+                }
```
madadam: If `vote_status.consensused` is `false`, we are still removing it from `self.vote_status`. Isn't that a problem?

src/parsec.rs
```diff
@@ -715,6 +771,74 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+            }
+        }
+
+        let peers_info: BTreeSet<(PeerIndex, S::PublicId)> = self
```
madadam: Two minor points:
1. No need to specify the type of the item of the `BTreeSet`. Just do `BTreeSet<_>` and let the compiler infer the rest.
2. It doesn't seem the collection needs to be a `BTreeSet`. I'd use `Vec` for slightly better performance.

src/parsec.rs
```diff
@@ -54,6 +54,55 @@ use std::{
 
 pub(crate) type KeyGenId = usize;
 
+struct VoteStatus {
+    payload_key: ObservationKey,
+    expected_voters: BTreeSet<PeerIndex>,
+    voted: BTreeSet<PeerIndex>,
```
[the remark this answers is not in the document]
madadam: Alternatively, just a single `BTreeSet<PeerIndex>` initialized with the expected voters and each time a voter votes, remove them from it?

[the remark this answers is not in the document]
madadam: Also, we have `PeerIndexSet` which is more efficient than `BTreeSet<PeerIndex>`.

src/parsec.rs
madadam: Hm wait, this is probably fine. Ignore the previous comment.

src/parsec.rs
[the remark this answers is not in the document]
maqi: Ok, I will have a check with the PeerIndexSet

src/parsec.rs
maqi: shall be not, as it only means it's a staled vote? i.e. not got consensused for a long time

src/parsec.rs
```diff
@@ -54,6 +54,55 @@ use std::{
...
+    consensused: bool,
+}
+
+impl VoteStatus {
+    fn new<P: PublicId>(event: &Event<P>, voters: BTreeSet<PeerIndex>) -> Self {
+        let payload_key = if let Some(payload_key) = event.payload_key() {
+            *payload_key
+        } else {
+            panic!("input event shall be an observation!");
+        };
```
fizyk20: How about `let payload_key = *event.payload_key().expect("input event should be an observation!");`?

src/parsec.rs
```diff
@@ -265,6 +316,7 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
             ignore_process_events: false,
 
             secure_rng: ParsecRng::new(secure_rng),
+            vote_status: VecDeque::default(),
```
fizyk20: Might be better to do `Default::default()` and let type inference handle the rest - this way we won't have to change this line if we ever decide to use a different container (as long as the other one will also implement `Default`).

src/parsec.rs
```diff
@@ -715,6 +771,74 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+                    .voters()
+                    .map(|(peer_index, _)| peer_index)
+                    .collect();
+                self.vote_status.push_back(VoteStatus::new(event, voters))
+            }
+        }
+    }
+
+    fn set_vote_consensused(&mut self, payload_keys: &[ObservationKey]) {
+        for payload_key in payload_keys.iter() {
```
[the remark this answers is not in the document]
fizyk20: Clippy usually catches such issues, I wonder why it didn't this time?

src/parsec.rs
```diff
@@ -715,6 +771,74 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
         Ok(event_index)
     }
 
+    fn add_observation_vote(&mut self, event: &Event<S::PublicId>) {
+        if let Some(payload_key) = event.payload_key() {
+            if payload_key.consensus_mode() == ConsensusMode::Supermajority
+                && !self
+                    .vote_status
+                    .iter_mut()
+                    .any(|vote_status| vote_status.add_vote(event))
```
fizyk20: My code smell sense is tingling when I see a potentially mutating method called inside `any()` :wink: I think I'd prefer something like:
```rust
if let Some(vote_status) = self.vote_status.iter_mut().find(|vote_status| vote_status.has_payload(event.payload())) {
    vote_status.add_vote(...); 
    // etc.
}
```
But this would add another nested `if`, so it might not be that much better... If you have any idea how to make this prettier, go ahead, otherwise feel free to ignore this comment. :wink:

src/parsec.rs
```diff
@@ -715,6 +771,74 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+    }
+
+    fn check_vote_status(&mut self) {
+        let threshold = ::std::cmp::min(
+            self.peer_list.voters().count() / 3,
+            self.vote_status.len() / 3,
+        );
+
+        // Only keep 10 * threshold consensused voting status and the unconsensused ones among them.
+        let mut concensused = self
```
fizyk20: Typo: `consensused`

src/parsec.rs
maqi: I will pick your approach if there is no those two extra level of if clause already.  :(

fizyk20: No worries! It's not bad, I just think it's less intuitive if a condition can mutate the data behind the scenes. I'll give it some more thought, but it should be okay as is :+1:

src/parsec.rs
```diff
@@ -715,6 +772,70 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+                .iter_mut()
+                .any(|vote_status| vote_status.set_consensused(payload_key));
+        }
+    }
+
+    fn check_vote_status(&mut self) {
+        let threshold = std::cmp::min(
+            self.peer_list.voters().count() / 3,
+            self.vote_status.len() / 3,
+        );
```
madadam: Might be worth to also extract those 3s into constants. Say `UNRESPONSIVENESS_THRESHOLD_FRACTION` or something like that.

src/parsec.rs
```diff
@@ -715,6 +772,70 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+                .vote_status
+                .iter()
+                .filter(|vote_status| vote_status.is_unresponsive(*peer_index))
+                .count()
+                > threshold
+            {
+                let _ = self.vote_for(Observation::Remove {
+                    peer_id: peer_id.clone(),
+                    related_info: Vec::new(),
+                });
```
madadam: Isn't unresponsiveness actually a form of malice? Then we should maybe vote for `Accusation` here instead of `Remove`.

src/parsec.rs
```diff
@@ -715,6 +772,70 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+    fn set_vote_consensused(&mut self, payload_keys: &[ObservationKey]) {
+        for payload_key in payload_keys {
+            let _ = self
+                .vote_status
+                .iter_mut()
+                .any(|vote_status| vote_status.set_consensused(payload_key));
+        }
+    }
+
+    fn check_vote_status(&mut self) {
```
madadam: I'd add some comment explaining what's going on in this function. For example why is the threshold chosen the way it is, and so on.

src/parsec.rs
```diff
@@ -54,6 +54,56 @@ use std::{
...
+    expected_voters: PeerIndexSet,
+    unvoted: PeerIndexSet,
+    consensused: bool,
+}
+
+impl VoteStatus {
+    fn new<P: PublicId>(event: &Event<P>, voters: PeerIndexSet) -> Self {
+        let payload_key = *event
+            .payload_key()
+            .expect("input event should be an observation!");
```
madadam: It might be possible to avoid the `expect` by making this function take `payload_key` and `creator` instead of `event`:
```rust
fn new(payload_key: ObservationKey, voter: PeerIndex, all_voters: PeerIndexSet) -> Self
```
Same for `add_vote`.

src/parsec.rs
```diff
@@ -54,6 +54,56 @@ use std::{
 
 pub(crate) type KeyGenId = usize;
 
+/// Affects the period (X consensued observations) after which node be considered as unresponsive.
+const UNRESPONSIVE_FACTOR: usize = 10;
+
+struct VoteStatus {
+    payload_key: ObservationKey,
+    expected_voters: PeerIndexSet,
```
madadam: Do we need this? Isn't the `unvoted` container enough to do everything we need?

src/parsec.rs
```diff
@@ -121,6 +171,8 @@ pub struct Parsec<T: NetworkEvent, S: SecretId> {
     ignore_process_events: bool,
     // Provided RNG: Needs to be cryptographically secure RNG as it is used for DKG key generation.
     secure_rng: ParsecRng,
+    // Including all on-going, consensused and polled blocks
+    vote_status: VecDeque<VoteStatus>,
```
madadam: Nitpick: this should be plural: `vote_statuses`.

tests/integration_tests.rs
```diff
@@ -199,6 +199,70 @@ fn faulty_nodes_terminate_at_random_points() {
...
 }
 
+fn faulty_peers_considered_as_left(
+    num_peers: usize,
+    num_faulty: usize,
+    min_fail_step: usize,
+    max_fail_step: usize,
+) {
+    let num_observations = 10;
+    let mut env = Environment::new(SEED);
```
jeanphilippeD: This only cover consensus mode supermajority, which is probably a problem.

But also how does the design work for opaque payload with single consensus mode as used by routing (I could not find the answer in the hackmd I looked at).

PR #351, opened 2019-11-14
Remove safe_crypto dependency

### LINKED ISSUE #349: Remove rust-sodium dependency
Currently the crate depends on `safe_crypto` for crypto related primitives which in turn has `rust-sodium` working for it under the hood. We would want both of these gone in favour of pure rust dependencies in our eco-system.

Removing it as a dev-dependency is left to https://github.com/maidsafe/parsec/issues/350

### LINKED ISSUE #350: Remove rust-sodium dev dependency
Currently the crate depends on safe_crypto for crypto related primitives which in turn has rust-sodium working for it under the hood. We would want both of these gone in favour of pure rust dependencies in our eco-system.

This require updating all the functional test as seeds will be broken.
Can use the same mock crypto code implemented in routing.

madadam:
Instead use locally defined mock crypto primitives.
Also fix some clippy warnings.

Closes #349 
Closes #350 

~TODO: fix test failure of `functional_tests::unpolled_observations`.~


src/functional_tests.rs
```diff
@@ -403,14 +403,16 @@ fn unpolled_observations() {
...
-        parse_test_dot_file("alice.dot"),
-        new_rng(&mut common_rng),
-    );
+    // Reset, re-run, and add another vote by Alice. Both `Add(Eric)` and the new vote should be
+    // returned in `out_unpolled_observations`. `Add(Eric)` should appear first, because it is
+    // consensused while the other vote is not.
+    let mut alice_contents = parse_test_dot_file("alice.dot");
+    let d_21 = unwrap!(alice_contents.remove_last_event());
+    let mut alice = TestParsec::from_parsed_contents(alice_contents, new_rng(&mut common_rng));
+    unwrap!(alice.add_event(d_21));
```
maqi: just to confirm. so the previous test doesn't have the step of adding d_21 into new created parsec instance, why it's required now?

src/mock.rs
```diff
@@ -27,28 +27,18 @@ pub const NAMES: &[&str] = &[
...
-
 /// **NOT FOR PRODUCTION USE**: Mock type implementing `PublicId` and `SecretId` traits.  For
 /// non-mocks, these two traits must be implemented by two separate types; a public key and secret
 /// key respectively.
 #[derive(Clone, Serialize, Deserialize)]
 pub struct PeerId {
     id: String,
-    pub_sign: PublicSignKey,
-    sec_sign: SecretSignKey,
+    pub_sign: PublicKey,
```
maqi: not sure, however as the names of members still contains `sign`, then why the type name got it removed?

src/mock.rs
```diff
@@ -139,18 +120,33 @@ impl Ord for PeerId {
...
     type Signature = Signature;
+
     fn verify_signature(&self, signature: &Self::Signature, data: &[u8]) -> bool {
-        self.pub_sign.verify_detached(&signature.0, data)
+        let mut hasher = DefaultHasher::new();
+        hasher.write(data);
+        hasher.write(&self.pub_sign.0);
+        let hash = hasher.finish().to_le_bytes();
+
+        signature.0[..8] == hash
```
maqi: this bit of hard coded length is concerning. At least we shall use a const here?

src/functional_tests.rs
madadam: I think it was a mistake and the test was previosly passing only accidentally. The problem is that `TestParsec::from_parsec_contents` does not actually calculate the consensus. Re-adding the event is what does it. The result is that the `Add(Eric)` vote gets consensused, but the new  vote does not which guarantees they will be returned from `our_unpolled_observations` in the expected order.

src/mock.rs
madadam: It doesn't matter much as those types are used only internally by this module. But the reason is that the keys are actually used both for signing and encryption. Maybe the fields should be renamed too (say `pub_key`, `sec_key`)?

PR #353, opened 2020-01-07
Update rand to 0.7.2

madadam:


examples/basic.rs
```diff
@@ -95,7 +96,7 @@ const ADD_PEER_CHANCE: u32 = 2;
 // We generate Remove observation(s) in a given round with 1 in `REMOVE_PEERS_CHANCE` odds.
 const REMOVE_PEERS_CHANCE: u32 = 5;
 
-type Seed = [u32; 4];
+type Seed = u64;
```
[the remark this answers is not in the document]
madadam: No, this is only for this example. This doesn't affect routing.

src/key_gen/parsec_rng.rs
```diff
@@ -32,16 +36,43 @@ impl ParsecRng {
...
+#[cfg(feature = "dump-graphs")]
+impl RngCore for ParsecRng {
     fn next_u32(&mut self) -> u32 {
         let next = self.secure_rng.next_u32();
         self.generated_values.push(next);
         next
     }
+
+    fn next_u64(&mut self) -> u64 {
+        impls::next_u64_via_u32(self)
```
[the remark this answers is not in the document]
madadam: Yes, with a small performance hit potentially. For example, the underlying rng might implement `fill_bytes` more efficiently than via `next_u32`.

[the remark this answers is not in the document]
maqi: if we can use this same `impls::` for the above, then we don't need to put these trait implementations under different feature flags.

need to decide which one is priorized.

examples/basic.rs
```diff
@@ -393,16 +394,15 @@ impl Params {
...
-                        println!("'{}' {}", SEED_ARG_LONG_NAME, seed_info);
-                        process::exit(6);
-                    }
-                });
+        params.seed = matches
+            .value_of(SEED_ARG_LONG_NAME)
+            .map(|seed_str| match seed_str.parse() {
+                Ok(seed) => seed,
+                Err(_) => {
+                    println!("'{}' {}", SEED_ARG_LONG_NAME, seed_info);
```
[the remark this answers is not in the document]
madadam: I think the message is sufficient as is. This error happens when you pass something that is not a number which should be pretty obvious. Besides, there was no additional error message before this commit and I don't want to introduce changes that are not related to the purpose of this PR.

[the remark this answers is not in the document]
maqi: I'd suggest something like:
```rust
Err(err) => {
                    println!("Failed in parsing '{}' {}, with error {}", SEED_ARG_LONG_NAME, seed_info, err);
```
However, if you do think it's not necessary. I will take your point and won't make this as a blocker.

src/key_gen/parsec_rng.rs
```diff
@@ -36,29 +35,11 @@ impl ParsecRng {
     }
 }
 
-#[cfg(not(feature = "dump-graphs"))]
-impl RngCore for ParsecRng {
```
jeanphilippeD: I'm not sure this is a good change.
What we had conceptually before in ParseRng being just the secure_rng in production, but now it does something different. I'd rather leave that out of the PR.

src/key_gen/rng_adapter.rs
```diff
@@ -31,7 +30,7 @@ where
     }
 
     #[inline]
-    fn try_fill_bytes(&mut self, bytes: &mut [u8]) -> Result<(), rand_core::Error> {
+    fn try_fill_bytes(&mut self, bytes: &mut [u8]) -> Result<(), rand_threshold_crypto::Error> {
         self.0.fill_bytes(bytes);
```
jeanphilippeD: Why not use self.0.try_fill_bytes

src/seeded_rng.rs
```diff
@@ -0,0 +1,57 @@
...
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use rand::{Error, RngCore};
+use rand_maidsafe_utilities::Rng;
+use std::fmt::{self, Debug, Formatter};
+
+// Wrapper for maidsafe_utilities::SeededRng that is compatible with the latest rand crate.
+pub struct SeededRng(maidsafe_utilities::SeededRng);
```
jeanphilippeD: Would it make sense to just copy the SeededRng from maidsafe_utilities since we want to get rid of that dependency?

tests/integration_tests.rs
```diff
@@ -569,7 +569,7 @@ fn extensive_dynamic_membership() {
     }
 
     loop {
-        if env.rng.gen_weighted_bool(chance_to_add) {
+        if env.rng.gen_ratio(1, chance_to_add) {
```
jeanphilippeD: I think for these I used https://docs.rs/rand/0.7.2/rand/trait.Rng.html#method.gen_bool in routing, but I guess your make more sense.

src/seeded_rng.rs
madadam: I think it would make more sense to just not use `SeededRng` instead of copying it over. can be done in a followup PR.

src/key_gen/parsec_rng.rs
madadam: I made the change base on the [comments](https://github.com/maidsafe/parsec/pull/353#discussion_r364123272) by @maqi . I can revert it but maybe you guys should agree first on what we want.

src/seeded_rng.rs
jeanphilippeD: whatever we did for routing i guess.

tests/integration_tests.rs
madadam: Since we have both `gen_bool` and `gen_ratio` now, I used the one requiring fewer changes.

src/key_gen/parsec_rng.rs
maqi: Sorry, don't get it.
Previously the feature guard is on the `next_u32` function only. Any my comment is mainly just about remain with the previous ?

jeanphilippeD: But the type change, that is why we now need to provide all the functions.
Before only next_u32 existed, and in production code, it was using directly the next_u32.
But RngCore require others, and instead of forwarding it like it was done, we now have different implementation for all of these functions.

maqi: Ah, right, I got what you mean.
Yeah, you are right, the change shall be reverted.
sorry, @madadam , my bad suggestion

src/key_gen/rng_adapter.rs
madadam: Actually, that doesn't work, because the `Error` types are different between the two rand versions.

src/key_gen/parsec_rng.rs
madadam: No worries. Will revert it.

src/key_gen/rng_adapter.rs
jeanphilippeD: unfortunate.

PR #356, opened 2020-02-25
Optimize create gossip

madadam:


src/peer_list/peer.rs
```diff
@@ -86,12 +95,24 @@ impl<P: PublicId> Peer<P> {
...
+            self.has_fork = true;
+        }
     }
 
     #[cfg(all(test, feature = "mock"))]
     pub(super) fn remove_last_event(&mut self) -> Option<EventIndex> {
         self.events.remove_last()
     }
+
+    pub fn has_fork(&self) -> bool {
```
maqi: has_fork, has_fork_at only required for malice_detection?

src/parsec.rs
```diff
@@ -1623,19 +1623,50 @@ impl<T: NetworkEvent, S: SecretId> Parsec<T, S> {
...
+            .flat_map(|(peer_index, peer)| {
+                let first = last_ancestors
+                    .get(peer_index)
+                    .map(|index_by_creator| index_by_creator + 1)
+                    .unwrap_or(0);
+                peer.events_from(first)
+            })
+            .collect();
+
+        // Include older forked events.
```
maqi: fork only required for malice detection, and here can we assume no fork happens?

src/peer_list/peer.rs
dirvine: Ah nice catch @maqi

madadam: I think we detect forks even without `malice_detection`, we just don't punish them.

dirvine: Is it worth even detecting them? or do we need to detect them to ignore them or something? Although if this is caused by bad behavior perhaps we should at least fire back an error or something identifying the bad node and then vote for his demise.

madadam: We need to detect forks otherwise malicious node could easily break consensus. This is regardless of whether we actually want to punish the offending node.

maqi: A fork in a gossip graph could ruin the caculation of meta election totally. I remembered we assumed such bad behaviour could happen quite possiblelly.
I raised the question here just would like to confirm for the baby fleming, is the same assumpation still apply.

madadam: > we should at least fire back an error or something identifying the bad node and then vote for his demise

Yes, this already happens when we enable the `malice_detection` feature which is currently disabled by default. Problem is we would currently immediately remove the offending node from parsec which is probably not what we want. What we should do is to raise the accusation to the upper layers (routing) and let then decide what to do. This requires some modifications to parsec, but probably not too extensive. Maybe worth creating an issue for this (unless one already exists).

dirvine: baby fleming perhaps not, but it looks like we should have a big FIXME and an issue for Fleming for it. 
Cheers @madadam  I was meaning if we ignore the result of this then is it worth running that code, but I think not running it is bogus as we must actually punish nodes that fork like this. Not running it would be a false sensation of running faster.

madadam: > I raised the question here just would like to confirm for the baby fleming, is the same assumpation still apply.

I guess for baby fleming we don't need to worry about forks as we are fully in control of all the nodes. But since the detection is already implemented I see no reason disabling it. It isn't computationally expensive anyway.

PR #363, opened 2020-05-20
feat(audit): add scheduled security audit scan

S-Coyle:


S-Coyle:
> Just wondering is that `no new line at end of file` normal ?

@maqi ah yeah I had missed that, looks like I had left the indentation spaces in instead of it being an empty line at the end of the file - have pushed up a fix.

If either of you guys could approve again, I'll merge. Thanks

PR #375, opened 2020-11-23
chore(cd): replace deprecated set-env in CD pipeline

S-Coyle:
set-env deprecated as of Nov 16th


S-Coyle:
Merged despite clippy CI fails as they are not related to this PR

